PR #18, opened 2017-03-03
allow interruption of python code

jjallaire:
This code implements the ability to interrupt executing Python code using the standard R interrupt handling mechanism.

Since execution of Python code occurs within native C++ code the normal R interpreter checking for interrupts that occurs during do_eval does not have  a chance to execute. Typically within Rcpp packages long running user code will call `Rcpp::checkForInterrupts` periodically to check for an interrupt and exit via an exception if there is one. However there is no opportunity  to do this during Python execution, so we need to find a way to get periodic callbacks during the Python interpreter's processing to perform this check.

This is provided for via the `Py_AddPendingCall` function, which enables the scheduling of a C callback on the main interpreter thread *during* the execution of Python code. Unfortunately this call occurs very eagerly, so we can't just schedule a callback and have the callback reschedule itself (this completely swamps the interpreter). Rather, we need to have a background thread that will perform the `Py_AddPendingCall` on a throttled basis (both time wise and in terms of only scheduling an additional callback while the Python interpreter remains running).

Having arranged for callbacks during Python interpretation at the appropriate rate, we then need to interact with R to check for a user interrupt, then interact with Python to notify it of any interrupt. We poll for interrupts in R using the same technique as `Rcpp::checkForException`. If we determine that an interrupt has occurred we call `PyErr_SetInterrupt`, which signals Python that an interrupt has been requested, which will ultimately result in a `KeyboardInterrupt` error being raised.

@kevinushey Would appreciate a careful review!

### DIFF
--- a/src/interrupt.cpp
+++ b/src/interrupt.cpp
@@ -1,0 +1,152 @@
+// This code implements the ability to interrupt executing Python code using 
+// the standard R interrupt handling mechanism.
+// 
+// Since execution of Python code occurs within native C++ code the normal R 
+// interpeter checking for interrupts that occurs during do_eval does not have 
+// a chance to execute. Typically within Rcpp packages long running user code 
+// will call Rcpp::checkForInterrupts periodically to check for an interrupt 
+// and exit via an exception if there is one. However there is no opportunity 
+// to do this during Python execution, so we need to find a way to get periodic
+// callbacks during the Python interpeter's processing to perform this check.
+// 
+// This is provided for via the Py_AddPendingCall function, which enables the 
+// scheduling of a C callabck on the main interpeter thread *during* the 
+// execution of Python code. Unfortunately this call occurs very eagerly, so we
+// can't just schedule a callback and have the callback reschedule itself (this
+// completely swamps the interpeteter). Rather, we need to have a background 
+// thread that will perform the Py_AddPendingCall on a throttled basis (both 
+// time wise and in terms of only scheduling an additional callback while the 
+// Python interpeter remains running).
+// 
+// Having arranged for callbacks during Python interpretation at the 
+// appropriate rate, we then need to interact with R to check for a user 
+// interrupt, then interact with Python to notify it of any interrupt. We poll 
+// for interrupts in R using the same technique as Rcpp::checkForException. If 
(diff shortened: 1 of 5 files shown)


terrytangyuan:
Nice addition!

src/interrupt.cpp
```diff
@@ -0,0 +1,152 @@
...
+// main thread while the interpreter is executing. Note that we can't just have
+// this function keep reschedulding itself or the interpreter would be swamped
+// with just calling and re-calling this function. Rather, we need to throttle
+// the scheduling of the function by using a background thread + a sleep timer.
+int pollForInterrupts(void*) {
+  
+  // Check whether an interrupt has been requested by the user. If one
+  // has then set the Python interrupt flag (which will soon after result
+  // in a KeyboardInterrupt error being thrown).
+  if (R_ToplevelExec(checkUserInterrupt, NULL) == FALSE)
```
kevinushey: Perhaps add in the comments that `R_ToplevelExec` returns `FALSE` if the execution of the associated function resulted in a jump to the top level? (and, when checking for interrupts, that will trigger and catch such a longjmp if the user has indeed requested an interrupt)

src/interrupt.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+  
+} // extern "C"
+} // anonymous namespace
+
+
+// Initialize interrupt polling background thread
+void initialize_interrupt_polling() {
+  thread t(interruptPollingWorker, NULL);
+  t.detach();
```
kevinushey: Do we need to worry about what happens to this thread if the Python or R interpreters go down? (ie: if Python or R crash, what happens to this thread?)

src/interrupt.cpp
kevinushey: Ah, I see that's documented below on the `checkUserInterrupt()` wrapper.

src/interrupt.cpp
jjallaire: If Python or R crashes then the whole process is going to crash (either is basically an end of the world scenario). I could put `catch(...)` around the handler code but it's all C code it's calling so there shouldn't be any C++ exceptions (and if there are I'd like to preserve whatever stack trace we can get).

PR #21, opened 2017-03-15
Added test for function signature

terrytangyuan:
@jjallaire This does not seem to pass. Am I missing anything? 

```
Failed -----------------------------------------------------------------------------------------------
1. Failure: Python function signatures are converted correctly (@test-python-closures.R#24) ----------
help$generate_signature_for_function(func) not equal to "(features, labels)".
1/1 mismatches
x[1]: "(..., ...)"
y[1]: "(features, labels)"
```

Some context of this: I am trying to get the new Estimator API in core work. However, it seems like it failed when validating my custom model function's signature [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/estimator/estimator.py#L687) even though my function contains those required arguments.

### DIFF
--- a/inst/python/rpytools/help.py
+++ b/inst/python/rpytools/help.py
@@ -1,4 +1,3 @@
-
 import sys
 import types
 import inspect
@@ -105,4 +104,3 @@ def generate_signature_for_function(func):
     if argspec.keywords:
       args_list.append("...")
     return "(" + ", ".join(args_list) + ")"
-
--- a/tests/testthat/test-python-closures.R
+++ b/tests/testthat/test-python-closures.R
@@ -16,3 +16,11 @@ test_that("R functions can accept named arguments from Python", {
   expect_equal(func(x = 15, y = 10), test$callFunc(pyfunc, y = 10, x = 15))
 })
 
+test_that("Python function signatures are converted correctly", {
+  skip_if_no_python()
+  help <- import("rpytools.help")
+  func <- function(features, labels) features - labels
+  pyfunc <- test$reflect(func)
+  expect_equal(help$generate_signature_for_function(func), "(features, labels)")
+})
+


jjallaire:
@terrytangyuan This is a good one! The problem is that we do indeed generate a Python signature that is basically `(..., ...)` (for handling a dynamic list of args and kwargs respectively). I think there will be a way to generate a higher fidelity signature but I need to dig into it a bit further.

jjallaire:
How many functions of this nature will we be required to marshall back into TF? I ask because I know of a one-off way for you to get this to work with the existing reticulate. Doing this would be much cheaper than solving the general problem, however if we expect to have to do this over and over again throughout the TF API it would be worth solving generally right now.

terrytangyuan:
Yep, one function should suffice for now.

jjallaire:
Okay, let me know the exact python signature of the function required and
I'll give you a PR to tflearn that implements it.

On Wed, Mar 15, 2017 at 9:24 AM, Yuan (Terry) Tang <notifications@github.com

terrytangyuan:
Thanks! Here it is:

```
model_fn <- function(features, labels, mode, params, config){
 ...
 return(someObject)
}
```

jjallaire:
Okay, none of the parameters have defaults?

On Wed, Mar 15, 2017 at 9:53 AM, Yuan (Terry) Tang <notifications@github.com

terrytangyuan:
mode can have default "train".

jjallaire:
If I try to define a Python function with just mode having a default value:

def model_function(features, labels, mode = "train", params, config)

Then I get:

SyntaxError: non-default argument follows default argument

Is it instead this:

def model_function(features, labels, mode = "train", params = NULL, config
= NULL)

Or is it this (i.e.. the function must accept all parameters):

def model_function(features, labels, mode, params, config)

On Wed, Mar 15, 2017 at 9:53 AM, Yuan (Terry) Tang <notifications@github.com

terrytangyuan:
Yes, the last two should both be fine.

jjallaire:
Okay, I implemented with all arguments required in this PR:

https://github.com/rstudio/tflearn/pull/5

Feel free to change as necessary.

On Wed, Mar 15, 2017 at 2:38 PM, Yuan (Terry) Tang <notifications@github.com

terrytangyuan:
Thanks! I'll give it a try soon!

PR #28, opened 2017-03-21
Converting Numpy Scalars to R causes an error

pourzanj:
The following code will yield an error because the array has shape ().

```r
np <- import("numpy")
a <- np$array(c(1.1))
```

I fixed this by just treating it as an array of shape (1). It's a quick and dirty fix, but I couldn't figure how the code that interacts with the numpy API was working.


jjallaire:
Thanks for this! I'll take a closer look in a couple of days and either merge this as is or do something else if looks cleaner.

jjallaire:
Addressed this issue here: https://github.com/rstudio/reticulate/commit/fbd81004ce5d53473bcded32b54fdb619a34d355

PR #30, opened 2017-03-22
Correct typo

drknexus:
speicfy -> specify


eddelbuettel:
The same typo is in the vignette on line 130.

PR #32, opened 2017-03-29
Support args of type list and tuple for doc generation

terrytangyuan:
Left a TODO for arg of type named list for future PR when someone gets a chance.


jjallaire:
Note that I'm also hoping to enhance things to pickup more of the docs (not just the title and the args but some of the other commentary typically there)

terrytangyuan:
Sounds good. Look forward to it!

PR #36, opened 2017-03-30
Fixes #35: incorrect signature when function parameters include list literals

### LINKED ISSUE #35: incorrect signature when function parameters include list literals
For the following function wrapper invocation:

```r
reticulate::py_function_wrapper("tf$contrib$keras$layers$Conv2D")
```

I get this signature:

```r
Conv2D <- function(1L, 1L), activation = NULL, use_bias = TRUE, kernel_initializer = "glorot_uniform", bias_initializer = "zeros", kernel_regularizer = NULL, bias_regularizer = NULL, activity_regularizer = NULL, kernel_constraint = NULL, bias_constraint = NULL, ...) {
}
```

It looks like the first incidence of a list literal results in the function signature up to that point getting consumed.

@terrytangyuan

terrytangyuan:


terrytangyuan:
Also added test to make sure things can run but not 100% what's the best way to match the output. 

Test is necessary because the issue happens in the file I didn't touch.

terrytangyuan:
Actually could you check my latest commit to see if this approach is ok?

PR #40, opened 2017-04-04
don't print error tracebacks by default

jjallaire:
- overridable via 'reticulate.traceback' option
- added `py_last_error` function to retreive the last error encountered


jjallaire:
@terrytangyuan Let me know if you think this is reasonable behavior. I think stack traces are great for development-time but mostly a confusing distraction for end users. If there is an easy way to instruct users to provide the stack trace to developers I'm hoping that's good enough.

terrytangyuan:
Yeah that sounds good to me

PR #53, opened 2017-05-06
Fixed issue where section is always empty in section_from_doc()

terrytangyuan:


R/help.R
```diff
@@ -277,7 +277,6 @@ sections_from_doc <- function(doc) {
 
 # Extract section from doc
 section_from_doc <- function(section, doc) {
-  section <- ""
```
jjallaire: I think the right fix *might* be to leave this here, but to rename the `section` parameter to `name` (otherwise the capture section text will always start with the section name).

terrytangyuan: That's right! Changed.

PR #58, opened 2017-05-14
Added names() for builtin objects/modules

terrytangyuan:


jjallaire:
Here's a modified implementation that accounts for a couple more things: https://github.com/rstudio/reticulate/commit/3d9e3698b2138b8601ef13d54e84326c0cfe5805

Note that the `names()` generic doesn't have a `...` argument (which is why roxygen didn't identify this as an S3 method in your confirm and travis complained). We also need to strip a couple of attributes provided by `.DollarNames` for the completion system (`as.character` strips the attributes).

terrytangyuan:
Thanks!

PR #88, opened 2017-08-21
force loading of modules in py_attr fns

kevinushey:
This PR ensures that the various APIs for accessing object / module attributes force the loading of lazy-loaded modules.


R/python.R
```diff
@@ -760,6 +783,8 @@ py_set_attr <- function(x, name, value) {
 #' @export
 py_list_attributes <- function(x) {
   ensure_python_initialized()
+  if (py_is_module_proxy(x))
+    py_resolve_module_proxy(x)
```
jjallaire: py_resolve_module_proxy may already do the equivalent of ensure_python_initialized (not sure, worth a check)

kevinushey: Looks like you're right. Should I take out the `ensure_python_initialized()` check? Would it ever be possible that we could get a non-module Python object before Python itself is initialized? (seems not possible but want to be sure since I haven't spent as much time in this codebase)

jjallaire:
Yes we definitely could, but in that case it would be essentially a null
XPtr which the native functions I believe check for

jjallaire:
Superceded by https://github.com/rstudio/reticulate/pull/89

PR #104, opened 2017-09-27
make reticulate work in an embedded python environment

### LINKED ISSUE #98: when R process is embedded in a Python process
`reticulate` is not working if the R process is embedded in a Python process. Particularly, `reticulate` doesn't run on [rice](https://github.com/randy3k/Rice) which is a python program embedding R. I would also imagine that it will not work with [rpy2](http://rpy2.readthedocs.io/).

I think the issue is that `reticulate` failis to initialize a Python instance when another python instance is running. However, I am not sure how easy/difficult the fix would be.

cf: https://github.com/randy3k/Rice/issues/7

randy3k:
close #98

When a python environment is detected., we load `rpycall` module rather than to insert it into the initialization table.

### DIFF
--- a/src/libpython.cpp
+++ b/src/libpython.cpp
@@ -154,6 +154,7 @@ bool LibPython::loadSymbols(bool python3, std::string* pError)
   bool is64bit = sizeof(size_t) >= 8;
 
   LOAD_PYTHON_SYMBOL(Py_Initialize)
+  LOAD_PYTHON_SYMBOL(Py_IsInitialized)
   LOAD_PYTHON_SYMBOL(Py_AddPendingCall)
   LOAD_PYTHON_SYMBOL(PyErr_SetInterrupt)
   LOAD_PYTHON_SYMBOL(PyExc_KeyboardInterrupt)
--- a/src/libpython.h
+++ b/src/libpython.h
@@ -139,6 +139,7 @@ void initialize_type_objects(bool python3);
 #define PyComplex_Check(o) (Py_TYPE(o) == Py_TYPE(Py_Complex))
 
 LIBPYTHON_EXTERN void (*Py_Initialize)();
+LIBPYTHON_EXTERN int (*Py_IsInitialized)();
 
 LIBPYTHON_EXTERN int (*Py_AddPendingCall)(int (*func)(void *), void *arg);
 LIBPYTHON_EXTERN void (*PyErr_SetInterrupt)();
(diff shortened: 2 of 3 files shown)


jjallaire:
I see a couple of issues here (one of which you already alluded to in another comment).

With this PR, reticulate will go through it's regular process to determine which version of Python to load (including caching this information for calls to `py_config()` and then at the last minute will substitute the version of Python used in the previous initialization. The least of the problem created here is that `py_config` and 'py_discover_config` will be producing wrong/misleading output (that certainly needs to be addressed as part of this PR).

The bigger issue is that any attempt by the user to influence which version of Python to use will be bypassed. This can be the explicit gestures like `use_python` or `use_virtualenv` but more importantly the implicit gesture of loading a package that requires a certain version of Python (e.g. the keras package requires the keras python package). As a result, you will need to document for users in this configuration that they are using an R interpreter hosted in Python and they should therefore *ignore all installation/configuration* instructions for packages that call Python, and instead do the appropriate pip/conda installs within the environment that bootstraps rpycall. I doubt that many users will even read this far into the documentation or if they do fully understand the implications.

jjallaire:
I am going to go ahead and merge this PR as it will unblock your scenario, but note that you have inherited a documentation burden associated with the fact that user attempts to influence the Python environment from R will be thwarted and therefore the documentation for packages like tensorflow doesn't apply (the user needs to pip/conda install all of their dependencies within the environment used by e.g. rice).

jjallaire:
I think we can solve the `py_config` and `py_discover_config` issues by forcing the use of `RETICULATE_PYTHON` when it's specified (that way we will always "discover" the correct version of Python and the py_config will match whatever the embedding environment is using). I'll make that change.

randy3k:
Thanks for considering the PR. It makes a lot of sense to use `RETICULATE_PYTHON` to force the version of python. Also, do we want to disable commands like `install_tensorflow()` and `install_keras()` when the version of python is forced by `RETICULATE_PYTHON`? Since it doesn't make a lot of sense to install the packages into the other python environments when the version of python was prespecified.

jjallaire:
I don't know if I'd disable a command entirely as the user might still want to take advantage of the side effect of `install_tensorflow()` in another session. Perhaps a warning and/or confirmation prompt though.

As I was saying before in this thread, this particular configuration Python -> R -> Python does create a documentation burden for e.g. rice as the user ultimately needs to understand that they are executing within a Python environment which will have implications for attempts to call R packages that in turn call Python. I think this needs to be in big bold letters for rice users or there will be many GitHub issues filed by users who don't understand why this or that R package that uses reticulate doesn't work.

randy3k:
Thanks, I will say something to rice users that when `reticulate` is loaded via `setHook(packageEvent("reticulate", "onLoad"),...)`.

jjallaire:
That's a great idea! Something along the lines of the fact that any Python packages used from R need to be available in the host Python environment.

randy3k:
No problem. I will make sure they will spam me rather than you. 😄

PR #107, opened 2017-10-07
initial work on knitr python engine

kevinushey:
This PR adds a `reticulate`-powered Python engine for use by `knitr`. Using this implies that it's now possible for Python chunks within a multimode document to share state during render, and building this on `reticulate` potentially opens the doors to other kinds of data sharing between Python and R chunks.

## TODO

- [ ] Sharing of objects between the R and Python sessions. This could likely be implemented with chunk options; we'll probably want to chat realtime about what this looks like.

- [x] Provide a module (or something similar) that makes it possible to call into R from Python.


R/knitr-engine.R
```diff
@@ -0,0 +1,110 @@
...
+    # attempt evaluate of current chunk of code
+    snippet <- extract(code, start, end)
+    captured <- py_capture_output(
+      result <- tryCatch(
+        py_run_string(snippet, convert = FALSE),
+        error = identity
+      )
+    )
+    
+    # detect syntax errors: assume that this means we need to pass an extra
```
kevinushey: Note to self: this probably won't work when submitting e.g. function definitions; e.g.

```
def foo:
    print 1
    print 2
```

If we go this route of interleaving input and output, we'll need some way of computing the span of a Python statement.

kevinushey:
I looked a bit at `compile()`, but unfortunately that's a bit too low-level for what we need, and the generated Python code object is not really separated into meaningful statements that can be mapped back to the source.

Fortunately, the AST generated by [ast.parse](https://docs.python.org/2/library/ast.html#ast.parse) is. So, given a chunk of Python code, I think we can do the following to perform statement-by-statement execution:

1. Call `ast.parse()` to parse the whole chunk,
2. Iterate through the body of that AST (the body is an array of nodes),
3. Extract the line numbers of those statements from the AST,
4. Use those to construct the boundaries for statements to be executed.

I'll take a stab at this approach in subsequent commits to this PR.

R/knitr-engine.R
```diff
@@ -20,12 +20,12 @@
 #' @export
 eng_python <- function(options) {
   
-  ast <- import("ast")
+  ast <- import("ast", convert = TRUE)
```
jjallaire: convert = TRUE is the default, why include it explicitly? (actually don't have any problem with you doing so as it provides documentation of intent, just curious)

R/knitr-engine.R
```diff
@@ -53,7 +53,31 @@ eng_python <- function(options) {
...
   # line index from which source should be emitted
-  pending <- 1
+  pending_source_index <- 1
+  
+  # plots captured from Python which need to be emitted
+  pending_plots <- list()
+  
+  # monkey-patch matplotlib's show method: rather than showing a plot
+  # on the screen, we write the plot to file
+  matplotlib <- import("matplotlib", convert = FALSE)
```
jjallaire: Can we affect plot sizes somehow? (we get the knitr figure sizes passed to us in options)

R/knitr-engine.R
kevinushey: Right; I just did that to be explicit here.

R/knitr-engine.R
kevinushey: I think we should be able to; let me investigate a bit more.

kevinushey:
Off to a good start :)

screen shot 2017-10-09 at 3 46 02 pm

R/knitr-engine.R
```diff
@@ -82,6 +79,9 @@ eng_python <- function(options) {
       knitr::include_graphics(path)
   }
   
+  # set up figure dimensions
+  plt$rc("figure", figsize = tuple(options$fig.width, options$fig.height))
```
jjallaire: If it's useful or better to have figure height in pixels you might want `out.width.px` (that's what we use in htmlwidgets: https://github.com/ramnathv/htmlwidgets/blob/346f87c390d118a8a2c2e13c2faa0b3a5cf9e5d6/R/sizing.R#L210-L211)

kevinushey: From what I can see thus far, most of the `matplotlib` APIs accept inches + DPI rather than raw pixels, hence why I'm using these directly.

jjallaire:
Marvelous!!!!!

jjallaire:
We will also want to inspect the `engine.path` knitr option (see http://rmarkdown.rstudio.com/authoring_knitr_engines.html#python) and emit a call to `use_python`  for that path (this will basically forward the engine.path request on to reticulate). Of course if reticulate is already initialized to another path this will have no effect but `use_python` won't complain.

yihui:
This looks terrific to me and is exactly what I have been looking forward to!! It is going to complete the (probably earliest) TODO in knitr: https://github.com/yihui/knitr/commit/5e51f61d0abf09275f143cb2dc6575dbd29682d8 😀 

Kevin asked me another question on Slack and I still need time to think about it.

jjallaire:
I am hopeful that we can make the reticulate engine either on by default in knitr or on by default in rmarkdown. Of course this requires that one of these packages take a dependency on reticulate.

If that's too much to swallow then we could make it a Suggests dependency and then have reticulate be included in the list of packages auto-installed alongside rmarkdown in RStudio.

R/knitr-engine.R
terrytangyuan: So this functionality depends on matplotlib?

jjallaire: We probably want to do a check for matplotlib before executing this. You can use `py_module_available("matplotlib")` for this.

kevinushey: `matplotlib` seems like the primary way through which Python users generate plots. Is there anything else we should consider supporting?

terrytangyuan: Others are a bit too fancy and may introduce heavier dependencies. I think matplotlib should suffice for now.

kevinushey:
Any thoughts about how we should allow users to communicate between R / Python chunks at this point? I can think of a couple possibilities:

We could provide some custom chunk options, giving a list of variables to be synchronized to / from the Python session. Something like:

    ```{python, import = "mtcars", export = "model"}
    model = fit(mtcars, ...)
    ```

In other words, we'd take care of marshaling the objects (using `reticulate`) based on the names provided. However, this won't scale very well if the user wants to share a lot of state.

Alternatively (or in addition?), `reticulate` could provide a Python module (which we would put on the Python path) which could provide some set of operations for interacting with the R session, e.g.

    ```{python}
    import reticulate as r
    mtcars = r.import("mtcars")
    model = fit(mtcars, ...)
    r.export(model, as = "model")
    ```

In each case, the functionality would be powered by `reticulate::py_to_r()` and `reticulate::r_to_py()`. Any other possibilities on your mind that might be worth exploring?

I think this work also means that we'll want to get `pandas` <-> `data.frame` marshaling ready soon, as well.

jjallaire:
Let's try to discuss this realtime tomorrow (I'll follow up with a schedule item)

R/knitr-engine.R
```diff
@@ -0,0 +1,156 @@
...
+  matplotlib <- import("matplotlib", convert = FALSE)
+  plt <- matplotlib$pyplot
+  
+  # save + restore old show hook
+  show <- plt$show
+  defer(plt$show <- show, envir = envir)
+  plt$show <- function(...) {
+    
+    # write plot to file
+    plot_counter <- yoink("knitr", "plot_counter")
```
yihui: `yoink` is okay; just remember to reset the counter in `eng_python()`:

```r
on.exit(plot_counter(reset = TRUE), add = TRUE)
```

kevinushey:
Just to recap on some earlier discussion, we've decided that we should try pursuing implicit sharing of objects first, e.g.

    ```{r}
    x <- 1
    ```

    ```{python}
    x += 1  # automagically reads 'x' from R sessions
    ```

    ```{r}
    x <- x + 1  # automagically sees updated 'x' from Python
    ```

This is definitely ambitious and there may well be dragons down this path, but it seems worth pursuing just out of how nice it would be able to work in a mode where R and Python just share the same namespace.

Barring this (or, perhaps, in addition to this), we'll consider an explicit scheme where R users use e.g. `py$object` to access a Python object, and Python users will use `r.object` to access an R object.

R/knitr-engine.R
```diff
@@ -0,0 +1,160 @@
...
+{
+  if (!py_module_available("matplotlib"))
+    return()
+  
+  matplotlib <- import("matplotlib", convert = FALSE)
+  plt <- matplotlib$pyplot
+  
+  # rudely steal 'plot_counter' (used below), and reset
+  # it when we're done
+  plot_counter <- yoink("knitr", "plot_counter")
```
yihui: I don't quite understand the `defer()` magic below, but the counter should be reset in `eng_python()` after `wrap()` is done (e.g. in `on.exit()`), otherwise I guess you will always get the same figure file path `chunk-label-1.png`.

kevinushey: `defer()` is basically a way of calling `on.exit()` in an arbitrary frame; ie, I use the fact that this function will only be called within `eng_python()`, and attach the `on.exit()` handler there.

yihui: Okay, that sounds clever!

kevinushey:
I spent some time today exploring the 'automagic' synchronization between R + Python chunks, but I think there are indeed too many dragons to slay here. Doing this right basically implies having a rock-solid AST / dependency analyzer for both Python and R code (so we know what variables are required by a particular chunk), and also some stateful tracking of what variables have been changed by each chunk (so we know whether a particular value for a variable should be retrieved from the R session, or the Python session).

The implicit version will also complicate other static analysis tools (e.g. discovering whether a variable in an R chunk is defined or not would now require introspection of Python code as well, and that's an expensive investment to make).

To wit, I think we should stick with explicit `r.var` and `py$var`.

kevinushey:
We now have the ability to communicate with the R session (specifically, read and write to objects in the global environment) from Python using code of the form:

````
```{python}
r['hello'] = "World"
data = r['x']
```
````

We support both of `r.x` and `r['x']` for accessing these items. Right now, there is automatic (eager) conversion done so that e.g. Python objects are eagerly converted to R objects (and vice-versa); we can likely make this lazy with e.g. active bindings or a bit of extra indirection on the Python side.

**Comment by jjallaire** [comment]
I think we will definitely want to make this lazy b/c of the prospect of copying large lists / matrixes between environments. Hopefully Python properties and R promises and/or active bindings will do the job.

**Comment by kevinushey** [comment]
How should we allow users to define how R / Python objects should be converted? Is there a way for us to have the user request that an R integer vector is converted to a `numpy` array vs. a Python list?

**Comment by jjallaire** [comment]
They just use as.array on the R side

There is currently no mechanism to customize this behavior. I think that
"casting" on the R side should be enough here, but we could consider
another mechanism if that doesn't seem adequate

**Comment by kevinushey** [comment]
I think this PR should be ready for review + merge. The conversion of dates/times + data frames will come in a separate PR.

The main change here is the introduction of a `py` active binding to `reticulate` -- users can use this to access the Python main module. It will be `NULL` until the Python session has actually been initialized.

File: R/zzz.R [hunk 7]
```diff
@@ -0,0 +1,33 @@
+#' Interact with the Python Main Module
+#' 
+#' The `py` object provides a means for interacting
+#' with the Python main session directly from \R.
+#'
+#' @format An \R object acting as an interface to the
+#'   Python main module.
+#' 
+#' @export
+"py"
```
jjallaire: I think it would be preferable if we made this available only during execution of R chunks in knitr. I have often used `py <- import_builtins()` in example code so exporting `py` might conflict with this.  I also don't think that `library(reticulate)` should be required in order to access Python chunks.  Do you think this is reasonable or do you feel strongly about making it globally available?

jjallaire:
The bit about scoping the availability of `py` is my only comment. Once we resolve that I'll merge.

Could you also bump the version and add an item to NEWS.md?

kevinushey:
I was thinking of having this be an object exported directly by `reticulate` for a couple of reasons:

1. It's easier for users to iterate (today) directly in e.g. an R Markdown document or an R Notebook, as the `py` object will just automagically be available in those contexts. Without this, we have to do some extra magic on the IDE side to make the `py` object available (+ make sure it works with autocomplete, etc)

2. The engine isn't the right place to create the `py` object. The engine only knows about Python chunks, not R chunks; in addition, one could imagine that in an R Markdown document one might want to interact with the Python session from R before actually running any Python chunks.

If name collision is an issue, we could rename this variable to e.g. `python` or `pymain` or something similar.

jjallaire:
I guess any existing use of `py` would still work (as the local assignment would have priority over the exported object). I also think `py` is a much more elegant handle than `pymain`. So given what you said I'm inclined to keep it as-is.

kevinushey:
Thanks for merging (+ thanks for bumping NEWS + version!)

jjallaire:
In terms of the data frame work, not sure when you will tackle this but
let's discuss before you get too far along. If it's possible to construct a
Pandas Data Frame from a named list of NumPy arrays then I believe we can
have Pandas data frames created from R data frames point to R memory and
therefore require no copying.

In terms of Pandas to R data frames, it's possible we could do something
with ALTREF vectors to have R data frames point to NumPy columns:
https://www.r-project.org/dsc/2017/slides/dsc2017.pdf

kevinushey:
> If it's possible to construct a
Pandas Data Frame from a named list of NumPy arrays then I believe we can
have Pandas data frames created from R data frames point to R memory and
therefore require no copying.

Some reading of the documentation suggests that `pandas` is pretty eager about copying any input data when a DataFrame is constructed. While it's definitely possible to construct a DataFrame from a Python list / dict of numpy arrays, those arrays will ultimately be copied when the object is constructed.

Fortunately the copies should be cheap when the datatypes match (ie, they'll just be a simple `memcpy` under the hood) but we won't be able to allow R and Python directly reference the same set of data.

FWIW it looks like we can already construct Pandas DataFrames from R `data.frames` (which is awesome!!), see e.g.

```
> pd <- import("pandas")
> py$mtcars <- pd$DataFrame(mtcars)
> py$mtcars$head()
    am  carb  cyl   disp  drat  gear     hp   mpg   qsec   vs     wt
0  1.0   4.0  6.0  160.0  3.90   4.0  110.0  21.0  16.46  0.0  2.620
1  1.0   4.0  6.0  160.0  3.90   4.0  110.0  21.0  17.02  0.0  2.875
2  1.0   1.0  4.0  108.0  3.85   4.0   93.0  22.8  18.61  1.0  2.320
3  0.0   1.0  6.0  258.0  3.08   3.0  110.0  21.4  19.44  1.0  3.215
4  0.0   2.0  8.0  360.0  3.15   3.0  175.0  18.7  17.02  0.0  3.440
```

This works because `mtcars` gets translated to a dictionary of numpy arrays, which pandas knows how to read directly.

The only issue is that `py_to_r` doesn't yet know how to convert Pandas DataFrames back to R `data.frame`s, but presuming there's a method to go from Pandas DataFrame to list of numpy arrays we might actually be a lot closer than we thought on this!

terrytangyuan:
You are looking for `df.as_matrix()`!

R/knitr-engine.R
jimhester: FWIW defer has been in the CRAN version of withr for a while now as `withr::defer()` if you don't want to duplicate the code here (but I guess presumably you do?).

shrektan: It's so great to have `withr::defer()`. Solved my headache elegantly. Thanks.

stuzog:
I don't know if this will help, but in following the RStudio Dashboards tutorial on [Dashboards](http://rmarkdown.rstudio.com/lesson-12.html) in RStudio 1.1.414 on OS X 10.11.6, I get the following error when trying to knit to flex_dashboard:

```
Quitting from lines 105-107 (11-dashboard.Rmd) 
Error in py_get_attr_impl(x, name, silent) : 
  AttributeError: 'module' object has no attribute 'pyplot'
Calls: <Anonymous> ... $.python.builtin.object -> py_get_attr -> py_get_attr_impl -> .Call
Execution halted
```
The offending line 105 is a simple python chunk:

``````{python}
x = "hello, python world!"
print(x.split(' '))
```
which parses fine in RStudio by itself. Attached is the example file. Hope that helps your efforts.

[11-dashboard.Rmd.txt](https://github.com/rstudio/reticulate/files/1657742/11-dashboard.Rmd.txt)

kevinushey:
That likely implies that either `matplotlib` is not available in your Python installation, or (for some reason) your version of `matplotlib` that doesn't provide `pyplot`.

stuzog:
@kevinushey That seems to be correct:

```
bash$ pip show matplotlib
Name: matplotlib
Version: 2.1.2
Summary: Python plotting package
Home-page: http://matplotlib.org
Author: John D. Hunter, Michael Droettboom
Author-email: matplotlib-users@python.org
License: BSD
Location: /Library/Python/2.7/site-packages
Requires: backports.functools-lru-cache, six, pytz, subprocess32, cycler, python-dateutil, pyparsing, numpy
bash$ pip show pyplot
bash$ pip install pyplot
Collecting pyplot
  Could not find a version that satisfies the requirement pyplot (from versions: )
No matching distribution found for pyplot
```
Is it possible to install pyplot on OS X 10.11.6?

kevinushey:
`pyplot` is actually a sub-module of `matplotlib`. Does this succeed or give you a module load error?

```
python -c "import matplotlib.pyplot"
```

(if that works the invocation should just quietly exit successfully)

kevinushey:
Also, if run from an R session:

```
library(reticulate)
matplotlib <- import("matplotlib")
matplotlib$pyplot
```

Does that work successfully for you?

stuzog:
No:

```
bash$ python -c "import matplotlib.pyplot"
  File "<string>", line 1, in <module>
  File "/Library/Python/2.7/site-packages/matplotlib-override/matplotlib/pyplot.py", line 34, in <module>
    from matplotlib.figure import Figure, figaspect
  File "/Library/Python/2.7/site-packages/matplotlib-override/matplotlib/figure.py", line 40, in <module>
    from matplotlib.axes import Axes, SubplotBase, subplot_class_factory
  File "/Library/Python/2.7/site-packages/matplotlib-override/matplotlib/axes/__init__.py", line 4, in <module>
    from ._subplots import *
  File "/Library/Python/2.7/site-packages/matplotlib-override/matplotlib/axes/_subplots.py", line 10, in <module>
    from matplotlib.axes._axes import Axes
  File "/Library/Python/2.7/site-packages/matplotlib-override/matplotlib/axes/_axes.py", line 21, in <module>
    import matplotlib.dates as _  # <-registers a date unit converter
  File "/Library/Python/2.7/site-packages/matplotlib-override/matplotlib/dates.py", line 126, in <module>
    from dateutil.rrule import (rrule, MO, TU, WE, TH, FR, SA, SU, YEARLY,
  File "/Library/Python/2.7/site-packages/dateutil/rrule.py", line 19, in <module>
    from six.moves import _thread, range
ImportError: cannot import name _thread
bash$ 

```

stuzog:
No:

```
> library(reticulate)
> matplotlib <- import("matplotlib")
> matplotlib$pyplot
Error in py_get_attr_impl(x, name, silent) : 
  AttributeError: 'module' object has no attribute 'pyplot'

```

However, `bash$ python3 -c "import matplotlib.pyplot"` seems to work, just silently returning the bash prompt. So maybe a knit of flex-dashboard would work in Python3?

fdrennan:
For me, the issue for solving the error:

```
Error in py_get_attr_impl(x, name, silent) : 
  AttributeError: 'module' object has no attribute 'pyplot'
```

was to run the following as root in order to import matplotlib.pyplot
```
apt-get -y install python-tk
```

PR #113, opened 2017-10-20
initial bits of pandas -> R conversion

kevinushey:
NOTE: not ready for merge; this is just a place for various discussion around the conversion to land (and for me to record notes as I go along).

Associated `feather` code that will likely be informative:

https://github.com/wesm/feather/blob/master/R/src/feather-types.cpp

## TODO:
- [x] Conversion of factors
- [x] Conversion of ordered factors
- [x] Conversion of dates
- [x] Conversion of strings (do we need to ensure UTF-8 encoding?)
- [x] Think about handling of integers (it looks like Pandas converts `int32` arrays to `int64` arrays by default; should we enforce `int32` if possible or just allow conversion?)
- [x] Remove dimensions from P -> R vectors (ie: prefer vectors to 1D arrays)
- [x] Double-check whether we need anything special to handle Pandas indexes


DavisVaughan:
Additional request for Pandas Series support!

jjallaire:
@kevinushey It looks like you are adding completion-related code to the Pandas PR. Is that on purpose?

kevinushey:
Sorry; no, I should've put that on a separate branch. I'll fix that up today.

kevinushey:
How should R factors be converted? Currently, they're converted as a plain numpy integer array, which seems wrong. Should they be character instead?

For Pandas data frames, we can convert R factors to categorical variables. Should we only do this within the context of a Pandas data frame? (I ask because it could potentially be surprising that the way we convert an R factor depends on where that conversion takes place, but perhaps that's fine)

terrytangyuan:
Converting to characters seems more intuitive to me. Also, it might be good to see some benchmarks for this pandas df -> R df conversion so users are aware of any potential overhead.

jjallaire:
For Pandas we should definitely convert to categorical variables. I don't think it's at all problematic to have a special behavior for the case of data frames.

I agree with @terrytangyuan that converting to character is more intuitive for the non pandas case.

kevinushey:
A question for the handling of datetimes: should we prefer converting R POSIXct objects to a list of vanilla Python `datetime` objects, or to a numpy array of `datetime64` objects? Should we toggle the behavior based on whether `numpy` is available?

kevinushey:
Also, we should think about how to handle pandas DataFrame row indexes (which I think can be an arbitrary numpy arrays) vs. R data.frame row names (which must be character). Some potential options:

- Attempt conversion of the Pandas index to character (this may be lossy for some data types)
- Generate a column in the R data.frame (`.index`) which more closely maps to the R analog of the index used for the Pandas dataframe

Any other thoughts?

DavisVaughan:
Re: _Or should this be a behavior that's toggled based on whether the array lives within a pandas DataFrame?_ - I think at the very least, this should be the case. 

I like the `.index` idea when going from Pandas -> R. Especially if the alternative is creating character data.frame row names, because if we use a tidyverse function after that, it often gets converted to `tibble` and those row names become more of a pain to extract.

Also on going from R -> Pandas, should there be an `index` argument that allows you to set the column to use as the index while converting over? I can imagine this being especially useful with converting to Series objects if those get supported. Or maybe that should all be done on the Pandas side after you convert over using `pandas$set_index()`.

kevinushey:
For the conversion operators (`r_to_py()` and `py_to_r()`) I think we want to keep the interface as simple as possible, so we just need some kind of sensible default behavior.

We might want implement separate standalone functions like `to_pandas_dataframe()` or `from_pandas_dataframe()` to extend that interface a bit, to allow for explicit control over the index.  We'll have to ponder this a bit because we definitely want to get the interface right!

Given that Pandas series are really just numpy arrays under the hood I think handling them shouldn't be too tough, although some work might be necessary to handle some of the numpy data types that don't map as cleanly to R.

DavisVaughan:
Makes sense on conversion operators. 

I could imagine separate standalone functions as: 

`as.data.frame.pandas.core.frame.DataFrame(x, index, ...)` to come from Pandas

`as_pandas_dataframe(x, ...)` to go to Pandas. This could be generic and have `as_pandas_dataframe.data.frame(x, index = NULL, ...)` as an implementation. Not sure what other things could be converted directly to a pandas data frame, but it gives the flexibility and keeps the naming convention for coercion functions.

I'm sure there are also reasons why this wouldn't be a good idea over more explicit functions like `from_pandas_dataframe()`, but it seems to make sense to me.

kevinushey:
Interestingly, with `feather`, indexes are dropped altogether, so there's some 'prior art' in this space that users should manage the index themselves:

https://github.com/wesm/feather/tree/master/python#limitations

It might not be _too_ bad to put the onus of index handling on the user. Alternatively, we could adopt some convention that the Python index becomes an R column called `.index` or something similar by default.

We could also just preserve R row.names in the case that the index is indeed a numpy array of strings.

PR #115, opened 2017-10-24
add 'rearray' helper for C-style reshaping

kevinushey:
This PR should help in cases where users want to reshape multidimensional arrays both at the R and Python level. By using the `rearray()` helper, users can ensure that arrays are reshaped using C-ordering for construction of the new indices, which should help ensure consistency between R and Python's notion of reshaping.

I went with `rearray()` for a couple reasons:

- `reshape()` is already taken by the `stats` namespace,
- `rearrange()` is a little bit too close to `dplyr::arrange()` and so the semantics might be confused by some users


tests/testthat/test-python-arrays.R
```diff
@@ -0,0 +1,32 @@
...
+  expect_reshape(a, c(2, 12))
+  expect_reshape(a, c(12, 2))
+  expect_reshape(a, c(2, 6, 2))
+  expect_reshape(a, c(6, 2, 2))
+  expect_reshape(a, c(2, 2, 3, 2))
+  expect_reshape(a, c(3, 2, 2, 2))
+  
+})
+
+test_that("rearray and dim<- don't do the same thing", {
```
terrytangyuan: mistyped a "<-" here?

terrytangyuan: Oh nvm I see this is intentional now.

kevinushey: That's intentional -- the method is literally called `dim<-`. In other words,

    dim(x) <- c(2, 2)

is actually equivalent to a call of the form

    x <- `dim<-`(x, c(2, 2))

jjallaire: Here: "`rearray()` will fill the new dimensions in row-major (`F`ortran-style) ordering"

Do you actually mean column-major?

jjallaire: I think `rearray` might be a bit too cryptic. How about `reshape_array` or `array_reshape`?  (I sort of prefer the second because it aligns in autocomplete with `array` and whatever other array helper functions we might add).

kevinushey: You're right that I have a mistake here -- it should be "row-major (C-style) ordering". I'll fix that up!

kevinushey: That sounds fine to me as well (@hadley felt somewhat strongly that `array` should be in the name too)

R/array.R
```diff
@@ -44,3 +44,47 @@ np_array <- function(data, dtype = NULL, order = "C") {
...
+#' dim(x) <- c(2, 2)
+#' x
+#' #      [,1] [,2]
+#' # [1,]    1    3
+#' # [2,]    2    4
+#' @export
+array_reshape <- function(x, dim, order = c("C", "F")) {
+  np <- import("numpy", convert = FALSE)
+  order <- match.arg(order)
+  reshaped <- np$reshape(x, as.integer(dim), order)
```
jjallaire: At one point had you speculated that the `order` argument affected how the inbound memory was interpreted? I wasn't sure whether that was in fact the case, but if it is then it seems like `order = "C"` would be wrong for R arrays. Let me know what you think.

kevinushey: This bit is definitely a bit confusing, but I think this is still correct.

The 'order' argument controls how, exactly, the data is reshaped / re-indexed; in particular, when increasing dimensionality, it controls whether the last dimension grows fastest (C-style) or the first (Fortran, or F-style). For example, if you had the vector 1:4, reshaping it to a 2x2 matrix in a C-style way would imply filling that new matrix by rows, then by columns, e.g.

```R
> array_reshape(1:4, dim = c(2, 2))
     [,1] [,2]
[1,]    1    2
[2,]    3    4
```

It's worth noting that doing C-style reshaping on an F-array implies that the underlying data buffer will be mutated; that is, we're doing more than just adjusting the dimensions, we're also reshaping the data so the new indexing scheme can be accommodated. It's also worth noting that this is not wrong; it's just a consequence of doing this C-style reshaping of an F-array.

```R
> c(array_reshape(1:4, dim = c(2, 2)))
[1] 1 3 2 4
```

It does imply that this form of reshaping will be inefficient for R arrays, but that's the cost we pay for correctness here.

jjallaire: Okay, sounds good! Just wanted to make sure.

PR #116, opened 2017-10-25
add section to array vignette on reshaping

jjallaire:
@kevinushey Could you take a look at this and let me know if it sounds correct?

@bwlewis We found that if you mixed NumPy reshaping and R reshaping in the same script (e.g. a tensorflow script that calls tf$reshape or np$reshape for training data, then R dim for prediction data) then the alternate ways of filling the reshaped array would cause problems. The `array_reshape()` function is intended to reshape R arrays in a manner consistent with NumPy.

A less hypothetical example would be a tensorflow model that was trained in Python (with reshaping of the raw input data) and then used in R (with dim based reshaping).

### DIFF
--- a/vignettes/arrays.Rmd
+++ b/vignettes/arrays.Rmd
@@ -508,6 +508,50 @@ o$sub(np$arange(1, 25)$reshape(4L, 3L, 2L), np$array(y))
 The above NumPy arrays are the same, their element-wise difference is zero.
 
 
+### Reshaping arrays
+
+In R you would typically reshape an array using the `dim<-()` function. For example:
+
+```{r, eval=FALSE}
+dim(x) <- c(1000, 28, 28)
+```
+
+In R, this operation simply changes the "dim" attribute of the array, effectively 
+re-interpreting the array indices as specified using column-major semantics.
+
+However, the NumPy `reshape` method uses row-major semantics, so if you are mixing
+R and Python code that reshapes arrays you will find that the reshaping will be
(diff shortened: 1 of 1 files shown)


vignettes/arrays.Rmd
```diff
@@ -508,6 +508,50 @@ o$sub(np$arange(1, 25)$reshape(4L, 3L, 2L), np$array(y))
...
+In R you would typically reshape an array using the `dim<-()` function. For example:
+
+```{r, eval=FALSE}
+dim(x) <- c(1000, 28, 28)
+```
+
+In R, this operation simply changes the "dim" attribute of the array, effectively 
+re-interpreting the array indices as specified using column-major semantics.
+
+However, the NumPy `reshape` method uses row-major semantics, so if you are mixing
```
kevinushey: We might want to say "by default", since technically one can switch between C-style reshaping and F-style reshaping by using `order = "{C/F}"` with `np$reshape()` / `array_reshape()`. We might also want to emphasize that some Python modules (e.g. Keras) will call `np$reshape()` without specifying an ordering (or allowing the user to specify an ordering) so our hands are effectively tied and to ensure consistency we need to do this row-major-style reshaping.

kevinushey:
Looks good to me! I just have one nit that we might want to emphasize that NumPy does C-style reshaping by default, and in general NumPy arrays are C-contiguous by default -- but both of these are configurable.

jjallaire:
Okay, just clarified that.

PR #125, opened 2017-11-16
Add support to trace call stacks for python threads

javierluraschi:
When an R program using Python hangs, it is helpful to print the callstacks from the running threads every X milliseconds, this PR adds this functionality. Usage:

```r
Sys.setenv(RETICULATE_DUMP_STACK_TRACE = 1000)
library(reticulate)
reticulate::py_eval("1+1")
```

Output:

```
THREAD: []
THREAD: [<module> <module> compile _compile compile parse _parse_sub _parse _parse_sub _parse _parse_sub _parse _parse_sub _parse _parse_sub _parse _parse_sub _parse ]
[1] 2
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,10 @@ Install the development version with: `install_github("rstudio/reticulate")`
 
 ## reticulate 1.3
 
+- Support for `RETICULATE_TRACE` environment variable which can be set to
+  the number of milliseconds in which to output into stderr the call stacks
+  from all running threads.
+
 - Use existing instance of Python when reticulate is loaded within an 
   embedded Python environment (e.g. rpy2, rice, etc.)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@
 
 Install the development version with: `install_github("rstudio/reticulate")`
 
-- Support for `RETICULATE_TRACE` environment variable which can be set to
+- Support for `RETICULATE_DUMP_STACK_TRACE` environment variable which can be set to
   the number of milliseconds in which to output into stderr the call stacks
   from all running threads.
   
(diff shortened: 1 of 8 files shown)


R/package.R
```diff
@@ -95,7 +96,8 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
                 config$virtualenv_activate,
                 config$version >= "3.0",
                 interactive(),
-                numpy_load_error)
+                numpy_load_error,
+                as.numeric(trace_ms))
```
jjallaire: This should probably be `as.integer` since that the type you end up with in C++.

jjallaire:
Looks great! I am going to wait to merge this until the CRAN release of Keras lands (b/c install of keras are pulling from reticulate master right now).

R/package.R
```diff
@@ -87,6 +87,7 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
                           Sys.getenv("PATH"),
                           sep = .Platform$path.sep))  
   
+  trace_ms <- Sys.getenv("RETICULATE_TRACE", 0)
```
jjallaire: Let's name this a bit more specifically (since `TRACE` could refer to general diagnostic logging). How about `RETICULATE_DUMP_STACK_TRACE`  or something like that?

R/package.R
javierluraschi: changed.

R/package.R
javierluraschi: changed

NEWS.md
```diff
@@ -1,8 +1,14 @@
+## reticulate 1.4 (development)
 
+<<<<<<< HEAD
```
jjallaire: Let's remove this merge artifact.

NEWS.md
```diff
@@ -13,6 +19,10 @@ Install the development version with: `install_github("rstudio/reticulate")`
 
 
 ## reticulate 1.3
+=======
```
jjallaire: Merge artifact here as well

PR #135, opened 2017-12-16
Fix issue detecting miniconda

terrytangyuan:
Also removed `anaconda4` since it doesn't seem to exist yet.

### DIFF
--- a/R/use_python.R
+++ b/R/use_python.R
@@ -260,7 +260,9 @@ find_conda <- function() {
       path.expand("~/anaconda/bin/conda"),
       path.expand("~/anaconda2/bin/conda"),
       path.expand("~/anaconda3/bin/conda"),
-      path.expand("~/anaconda4/bin/conda")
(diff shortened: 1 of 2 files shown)


R/use_python.R
```diff
@@ -260,7 +260,9 @@ find_conda <- function() {
       path.expand("~/anaconda/bin/conda"),
       path.expand("~/anaconda2/bin/conda"),
       path.expand("~/anaconda3/bin/conda"),
-      path.expand("~/anaconda4/bin/conda")
```
jjallaire: That's there for if/when there is an anaconda 4.

It seems like we'd also want the system level conda paths here? (i.e. just like https://github.com/rstudio/reticulate/blob/9e02a168788962539e83d1ea0fb85f084e85b274/R/config.R#L238)

terrytangyuan: Oh yeah. I just added a commit that includes some missing ones.

Issue #130, opened 2017-11-26
eng_python does not respect echo


otommod:
R does it like [this](https://github.com/yihui/knitr/blob/87789284d153f8ef36fe4f098a037d012af009db/R/block.R#L195-L204).  The rest of the engines [rely](https://github.com/yihui/knitr/blob/33a3ba65771e2ebc37a403cb37f941f2c6660bfb/R/engine.R#L81) on `knitr::engine_output()`.

`eng_python` is not using `engine_output()` yet, however even when it will, that won't help since it will [bypass the "echo" logic](https://github.com/yihui/knitr/blob/33a3ba65771e2ebc37a403cb37f941f2c6660bfb/R/engine.R#L63)

Currently, as a workaround I've defined
```r
eng_python <- function(opts) {
  out <- reticulate::eng_python(opts)
  if (opts$echo) {
    out
  } else if (length(out) > 1) {
    out[seq(2, length(out), 2)]
  }
}
```
in my Rmds, though it's not particularly pretty.

jjallaire:
@kevinushey Could you take a look at this?

kevinushey:
Will take a look next week (will be back in office then).

jjallaire:
Okay, thanks!

jjallaire:
@kevinushey I know you are neck deep in QWebEngine right now, but I'm hoping to do a CRAN release of reticulate very soon and it would be good to have this issue resolved. It looks on the surface to be something pretty easily addressed, can you take a look early next week?

eddelbuettel:
Reminder of clock ticking with CRAN closing for Xmas break on Dec 22.

jjallaire:
Thanks! I was wondering if/when there would be a holiday break. Do you know when they will resume accepting submissions in January?

eddelbuettel:
Incoming will be down Dec 22 to Jan 3.  Uwe posted it to r-devel and r-package-devel, and it is on the CRAN submission page but I wasn't sure you'd seen it.   

Not a terribly long time.   So I am probably just releasing some lighter-weight fare -- a littler update is kinda overdue -- and resume with "normal" uploads in January.  Not sure where you are with with reticulate.  It has been montly to bi-monthly so either way fits the bill...

jjallaire:
Okay, I am going to try get a few submissions in next week then!

On Sat, Dec 16, 2017 at 3:11 PM, Dirk Eddelbuettel <notifications@github.com

kevinushey:
[...]

PR #136, opened 2017-12-17
implement support for variety of chunk options

kevinushey:
This PR implements support for a couple knitr chunk options (mainly prompted by #130). With this PR, we now have support for the boolean forms of `echo`, `eval`, and `include`.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -96,28 +96,39 @@ eng_python <- function(options) {
   
   for (range in ranges) {
     
-    # evaluate current chunk of code
+    # extract code to be run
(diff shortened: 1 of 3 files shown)


jjallaire:
Excellent! The code looks relatively straightforward so net pretty safe. Are you comfortable with me merging and then submitting to CRAN ~ 24 hours later or do you want to hold this fix for the next CRAN release?

kevinushey:
I think we can take this now. The changes are indeed straightforward and we have a test .Rmd document that verifies everything behaves as expected so we should be good to go.

jjallaire:
Great, thanks!

PR #140, opened 2017-12-27
Remove implicit doc extraction for Python classes

terrytangyuan:
This PR removes the implicit call to `func.__init__` and `func.__call__` when extracting docs for classes. Developers should instead just call `ClassName.__init__` explicitly instead. Otherwise, it's unclear which part of the generated docs belongs to which methods.

### DIFF
--- a/inst/python/rpytools/help.py
+++ b/inst/python/rpytools/help.py
@@ -12,21 +12,9 @@ def isstring(s):
 
 
 def normalize_func(func):
-    # convert class to __init__ method if we can
-    if inspect.isclass(func):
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks for this (and the other PR). I am preparing to submit reticulate to CRAN in the next few days so don't want to merge any changes that aren't critical bug fixes. I'll merge both of your PRs after that.

terrytangyuan:
Sure, no problem!

terrytangyuan:
@jjallaire Can this be safely merged now? I don't think there are too many other people using this besides us.

PR #143, opened 2018-01-03
allow configuration of reticulate engine

kevinushey:
This PR opens up the reticulate engine a bit, providing the required endpoints for the RStudio IDE (or other front-ends) to hook in and handle various kinds of outputs produced by the engine.

This PR adds support for the following R options:

- `reticulate.engine.matplotlib.show`: Controls what happens when a `matplotlib` plot is shown;
- `reticulate.engine.wrap`: Controls how the generated list of outputs should be post-processed.
- `reticulate.engine.environment`: Controls what R environment is used for R <-> Python synchronization.

The intention is for the RStudio IDE to set these hooks to route console + plot output back to the Notebook cache, and display them in the IDE.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -112,9 +112,9 @@ eng_python <- function(options) {
       
       # append pending source to outputs (respecting 'echo' option)
       if (!identical(options$echo, FALSE)) {
-        extracted <- extract(code, c(pending_source_index, range[2]))
         if (is.numeric(options$echo))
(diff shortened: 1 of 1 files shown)


jjallaire:
I'm thinking of doing a release of reticulate very soon (in preparation for final releases of keras, tensorflow, etc. prior to rstudio::conf). Are these changes safe to take at the 11th hour or should we wait and keep them on a branch? (realize that there is significant utility in being able to rely on reticulate from CRAN for IDE features so am definitely open to taking the PR, just want to double-check on safety)

PR #150, opened 2018-02-01
candidate fix for crash seen w/new Anaconda

kevinushey:
This PR fixes a crash that seems to occur when attempting to interleave the loading of R packages with Python / Keras code. The code I was running to trigger the crash:

```R
library(keras)

imdb <- dataset_imdb()

c(x_train, y_train) %<-% list(imdb$train$x, imdb$train$y)

library(digest)

model <- keras_model_sequential()
model %>% 
  layer_embedding(input_dim = 80, output_dim = 128) %>% 
  layer_lstm(units = 64) %>% 
  layer_dense(1, activation = "sigmoid")
```

When run from the R terminal, I saw:

```R
> model <- keras_model_sequential()
> model %>%
+   layer_embedding(input_dim = 80, output_dim = 128) %>%
+   layer_lstm(units = 64) %>%
+   layer_dense(1, activation = "sigmoid")
Intel MKL FATAL ERROR: Cannot load mkl_intel_thread.dll.
```

The `mkl_intel_thread.dll` library is located within the `Library/bin` subdirectory of the Python 'home', so adding that to the PATH appears to alleviate the issue.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -78,11 +78,29 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   
   # add the python bin dir to the PATH (so that any execution of python from 
   # within the interpreter, from a system call, or from within a terminal 
-  # hosted within the front end will use the same version of python. On 
-  # Windows also add the Scripts path
-  python_dirs <- normalizePath(dirname(config$python))
-  if (is_windows())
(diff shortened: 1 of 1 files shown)


jjallaire:
This actually does fix things for me as well, yay!!!

PR #155, opened 2018-02-13
source_python() function for sourcing Python scripts into R environments

jjallaire:
@kevinushey Could you review?


R/source.R
```diff
@@ -0,0 +1,31 @@
...
+  Encoding(names) <- "UTF-8"
+  
+  # assign the objects into the specified environment
+  if (!is.null(envir)) {
+    for (name in names)
+      assign(name, dict[[name]], envir = envir)
+  }
+  
+  # return nothing
+  invisible(NULL)
```
kevinushey: Any reason why we return `NULL` rather than the dictionary produced by `py_run_file()`? It might be useful to enable:

```R
output <- source_python("file.py", NULL)
```

jjallaire: The dictionary by itself might not be that useful (because they need to use Python dictionary methods to interact with it). We could copy it to an R list but that would create an extra copy operation that would often be unnecessary. My thinking was that we should emulate the R `source()` function and that as you said `py_run_file()` could be used for explicitly fetching the results of execution.

kevinushey: Got it; sounds good!

PR #156, opened 2018-02-13
implement a simple Python REPL on top of reticulate

kevinushey:
This PR implements a simple Python interpreter built on top of `reticulate`, effectively transforming the R REPL into a Python REPL.

screen shot 2018-02-13 at 11 40 39 am

Because the Python REPL is attached to the 'main' Python session, changes made in the REPL persist for any future usages of Python (either within the REPL or through other `reticulate` APIs).

R frontends (e.g. RStudio) could take advantage of this by detecting if the user is attempting to execute code within an R file vs. a Python file, and activate the R or Python REPL as appropriate before sending code to the console.


R/repl.R
```diff
@@ -0,0 +1,117 @@
...
+  quit_requested <- FALSE
+  
+  repl <- function() {
+    
+    # read user input
+    prompt <- if (length(buffer)) "... " else ">>> "
+    contents <- readline(prompt = prompt)
+    
+    # special handling for e.g. 'quit', 'exit'
+    if (contents %in% c("quit", "exit")) {
```
jjallaire: do we need to trim whitespace before the comparison?

kevinushey: Yes we should!

R/repl.R
```diff
@@ -0,0 +1,117 @@
...
+      error <- py_last_error()
+      message(paste(error$type, error$value, sep = ": "))
+    }
+    
+  }
+  
+  # notify the user we're entering a reticulate-powered Python REPL
+  config <- py_config()
+  
+  fmt <- c(
```
jjallaire: during line by line execution from RStudio it seems like we'd want to suppress this banner .

jjallaire:
For more end-user oriented functions I tend to not use a `py_` prefix (e.g. `import()`, `source_python()`, `dict()`, etc. Based on this I wonder if this function should be called `python_repl()` or `repl()`. I could still go for `py_repl()` just want us to consider this.

jjallaire:
It would be nice if from within RStudio we could exit the REPL implicitly (i.e. don't always enter and exit for every line, but rather exit automatically when a non-Python command goes to the REPL

kevinushey:
That's exactly what I was thinking of wiring up on the RStudio side. We originally discussed prefixing lines with e.g. `%py` or some marker, but I thought the experience might be nicer if we could automatically toggle the Python REPL on and off depending on whether Python code or R code is being executed.

kevinushey:
I'm also a fan of just naming the function `python_repl()`.

R/repl.R
kevinushey: I added the `banner` argument to control this.

kevinushey:
Found one ugly hiccup. R's `readline()` actually trims leading and trailing whitespace from the user input, which is quite problematic when trying to submit Python code (since indentation is required there).

Will have to think whether there's a reasonable workaround :-/

jjallaire:
Here is the code that implements readline:

https://github.com/wch/r-source/blob/bf0a0a9d12f2ce5d66673dc32cd253524f3270bf/src/main/scan.c#L994-L1039

It ultimately ends up calling R_ReadConsole. Can that be called from within a package? The R_ReadConsole function I believe ultimately resolves to ptr_R_ReadConsole which is implemented by the front-end.

If that isn't a possibility then we could also make this an RStudio feature (or make it a reticulate feature that requires some explicit coordination with a front-end, which we could implement in RStudio).

R/repl.R
```diff
@@ -7,8 +7,10 @@
 #' 
 #' @import utils
 #' 
+#' @param banner Boolean; print a startup banner when launching the REPL?
+#' 
 #' @export
-py_repl <- function() {
+python_repl <- function(banner = TRUE) {
```
jjallaire: maybe `quiet = FALSE` (to match the R `--quiet` startup option and the python `-q` startup option)

jjallaire: I am thinking that the IDE may indeed want to show this banner by default as it documents which version of Python is being used.

jjallaire: We may want a `module` argument that allows the user to specify that they want to interact with a specific module (e.g. "keras" or "tensorflow"). We can then call `import(module)` to force reticulate to bind to a version of Python that has this module.

kevinushey:
Fortunately `R_ReadConsole()` is not screened by `R CMD check` so I think we can safely use it!

R/repl.R
kevinushey: Should we allow for a vector of modules?

R/repl.R
```diff
@@ -5,11 +5,22 @@
...
+#' @param modules A set of Python modules (packages) to be imported before
+#'   the REPL is launched. Defaults to the value of the `reticulate.repl.modules`
+#'   option.
 #' 
 #' @export
-python_repl <- function(banner = TRUE) {
- 
+python_repl <- function(
+  quiet = FALSE,
+  modules = getOption("reticulate.repl.modules", default = character()))
```
jjallaire: This should just be a single `module` as the purpose of this is not actually to pre-import the module but to tell us which version of Python to use (this is a much nicer way of doing this than requiring the full path to a Python binary or the explicit call to a `use_` function). Also, this should be the first argument so people can write e.g.`python_repl("keras")`

kevinushey:
One thing I need to figure out is auto-printing for statements, e.g.

```
>>> x = {'a': 1}
>>> x['a']
```

Currently, that doesn't auto-print when passed through `py_run_string()`.

R/repl.R
```diff
@@ -5,11 +5,22 @@
...
+#' @param quiet Boolean; print a startup banner when launching the REPL? If
+#'   `FALSE`, the banner will be suppressed.
+#' @param modules A set of Python modules (packages) to be imported before
+#'   the REPL is launched. Defaults to the value of the `reticulate.repl.modules`
+#'   option.
 #' 
 #' @export
-python_repl <- function(banner = TRUE) {
- 
+python_repl <- function(
```
jjallaire: Now that I see `python_repl()`, I think we are better of going back to `py_repl()` for consistency with the other `py_` functions. I think if we could get away with something simple like `repl()` that would be better, but I think that would be over-reaching. Let me know if you agree.

src/readline.cpp
```diff
@@ -0,0 +1,28 @@
+#include <Rcpp.h>
+using namespace Rcpp;
+
+#include <Rinterface.h>
+
+extern "C" int R_ReadConsole(const char*, unsigned char*, int, int);
```
jjallaire: If you do have trouble with this (e.g. on Windows) then you might also be able to import `ptr_R_ReadConsole` (although we would want to see if R_ReadConsole does any wrapping of that call we'd want to emulate)

R/repl.R
kevinushey: I just worry about in-theory masking of other functions called `repl()` in other packages, but maybe that's fine and we could / should just encourage people to call `reticulate::repl()` directly if that ever did come up.

jjallaire:
Since we already have `py` available via reticulate, do we also want to make `r.` available to the execution of the Python code? That would make the REPL and the notebook have the same behavior. If that's too tricky / expensive though we could wait for someone to request it.

jjallaire:
I agree about `repl()` being too generic. I also think we are going to want people to do `library(reticulate)` followed by `py_repl()` so that they can get access to `py$` from R.

R/repl.R
jjallaire: I agree about repl() being too generic. I also think we are going to want people to do library(reticulate) followed by py_repl() so that they can get access to py$ from R.

jjallaire: I wonder if there is a Python function somewhere in their standard libraries that would give us the printing behavior we are looking for.

jjallaire: I wonder if you should be using `py_eval()` (which returns the result of evaluating the expression) rather than `py_run_string()`

kevinushey: `py_eval()` unfortunately doesn't work for certain kinds of Python statements; e.g.

```
> py_eval("x = 1")
Error in py_eval_impl(code, convert) : 
  SyntaxError: invalid syntax (reticulate_eval, line 1)
```

R/repl.R
```diff
@@ -14,7 +14,7 @@
 #' @export
 py_repl <- function(
   module = getOption("reticulate.repl.module", default = character()),
```
jjallaire: I think this should default to `NULL` (with no option) just to make the interface more straightforward. We can add an R option if we discover that it's useful, but I don't want to start with this.

R/repl.R
```diff
@@ -0,0 +1,171 @@
...
+  locals <- py_run_string("locals()")
+  globals <- py_run_string("globals()")
+  
+  # check to see if the current environment supports history
+  has_history <- tryCatch(
+    { utils::savehistory(tempfile()); TRUE },
+    error = function(e) FALSE
+  )
+  
+  if (has_history) {
```
jjallaire: Does this mechanism also account for the RStudio global history? (IIRC we don't save entires from the non-default prompt but I'm not sure if that's the case)

R/repl.R
```diff
@@ -0,0 +1,204 @@
...
+  
+  # command compiler (used to check if we've received a complete piece
+  # of Python input)
+  compiler <- codeop$CommandCompiler()
+  
+  # record whether the used has requested a quit
+  quit_requested <- FALSE
+  
+  # inform others that the reticulate REPL is active
+  options(reticulate.repl.active = TRUE)
```
jjallaire: This could also set a mutable global in the reticulate package and then we could expose a `py_repl_active()` function. Would that work?

jjallaire:
On OSX I get a very long Python version string:

```
Python 2.7.10 (default, Feb  7 2017, 00:08:15)  [GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)]
Reticulate 1.4.0.9001 REPL -- A Python interpreter in R.
```

I think we should truncate this.

I also think it would be valuable to print the full path to the Python binary on the same line, so maybe:

```
Python 2.7.10 (/opt/local/bin/python)
Reticulate 1.4.0.9001 REPL -- A Python interpreter in R.
```

R/repl.R
kevinushey: We can do that. Any reason to prefer that over the R option?

kevinushey:
@jjallaire one issue for integration with RStudio is that attempts to open an interactive console from within Python can lock up the R session. This can happen, for example, when executing `help()`:

```
>>> help()
```

What do you think is the best way to handle this? My initial thought was to allow for the IDE to define a REPL hook that re-routes what happens when the user executes `help()`, but curious if you have other thoughts.

R/repl.R
kevinushey: It does! Effectively, there are two separate histories -- one for R, and one for Python, and the active history gets toggled on / off as the REPL is activated / deactivated.

R/repl.R
```diff
@@ -202,3 +220,14 @@ py_repl <- function(
...
   
 }
+
+#' Check Whether the Python REPL is Active
+#' 
+#' Check to see whether the Python REPL is active. This is primarily
+#' for use by R front-ends, which might want to toggle or affect
+#' the state of the Python REPL while it is running.
+#' 
+#' @export
```
jjallaire: I think that we can make this non-exported for now (since R front ends can call non-exported methods without problem) just to not clutter up the API.

jjallaire:
I think we should indeed have a hook that we use to block these actions (I don't think we need to emulate them, the main priority is just preventing the lockup). BTW if `help()` is invoked with an argument we can delegate to `py_help()`

jjallaire:
Seems like we are good to go here. Do you want to make any more tweaks before I merge?

kevinushey:
Yes, I think this is good to go! I might need to supply a couple fix-up PRs while I work out integration with RStudio but this should be in a good state.

PR #158, opened 2018-02-15
Export py_dict

BaruchYoussin:
Export `py_dict` function.
I use it in tensorflow to create python dictionary indexed by python objects when its number of elements is not known in advance and is defined programmatically.


jjallaire:
What about the `dict()` function? Would that already exported function do what you need?

BaruchYoussin:
No. `dict` requires key-value pairs to be known in advance hard coded while I need to create them programmatically, and I do not know their number in advance.

BaruchYoussin:
That works!  Thanks!  This is good enough for my purposes; I did not realize I could do this.
If you do not wish to export `py_dict` then you can close this pull request but better keep this discussion public and searchable, in case other people need it.

jjallaire:
This would also work:

```r
dict(list(a = 1, b = 2, c = 3))
```

BaruchYoussin:
Thanks!

BaruchYoussin:
On the second thought, I do not understand how it can work.
I need to create a Python dictionary from a list of keys which are Python objects, and a list of values for them.
As far as I understand the code of `dict()`, I need to put each key individually onto my environment rather than being an element of my list, and give it a unique name.
After that I need to create a list of my values naming them with the names of my keys, and give this list to `dict()`.
This is not what I want to do.
Am I missing something?

jjallaire:
Good point!

I have a branch which fully implements this feature here: https://github.com/rstudio/reticulate/tree/feature/py_dict

I haven't merged it yet because I'm wondering if I can somehow get the `dict()` function to encompass this functionality so we don't have 2 functions that serve such similar purposes.

jjallaire:
I merged the branch (there isn't a great way to add arguments to `dict()` since these would mask arguments of the same name intended to be dict elements)

BaruchYoussin:
Thanks!
Just wanted to point out:
You still have the lines 488-489 in python.R:
```
#' @param keys Keys to dictionary (can be Python objects)
#' @param values Values for dictionary
```
before the definition of `dict(..)`
while these lines are missing before the definition of `py_dict(..)`

jjallaire:
That's correct (the docs for those 2 functions are consolidated)

BaruchYoussin:
I see, thanks!

PR #168, opened 2018-02-27
Support Sphinx style documentation help handlers and wrapper generation

terrytangyuan:
* Allows to register help handlers from Python functions documented in [Sphinx style](http://www.sphinx-doc.org/en/master/), which is popular in many Python packages. 
* Allows to generate wrappers for functions documented in Sphinx style. 
* Fixed issue with `py_function_wrapper` not handling enum items in param, e.g. `embedding_type` parameter in the example below.

Below is an example:

```
LFDA <- py_run_string("
class LFDA(object):
  '''
  Local Fisher Discriminant Analysis for Supervised Dimensionality Reduction
  Sugiyama, ICML 2006
  '''
  def f1(self, num_dims=None, k=None, embedding_type='weighted'):
    '''
    Initialize LFDA.

    Parameters
    ----------
    num_dims : int, optional
      Dimensionality of reduced space (defaults to dimension of X)
    k : int, optional
      Number of nearest neighbors used in local scaling method.
      Defaults to min(7, num_dims - 1).
    embedding_type : str, optional
      Type of metric in the embedding space (default: 'weighted')
        'weighted'        - weighted eigenvectors
        'orthonormalized' - orthonormalized
        'plain'           - raw eigenvectors

    Returns
    -------
     values : array-like
    '''
    if embedding_type not in ('weighted', 'orthonormalized', 'plain'):
      raise ValueError('Invalid embedding_type: %r' % embedding_type)
    self.num_dims = num_dims
    self.embedding_type = embedding_type
    self.k = k
")

py_function_wrapper("LFDA$LFDA$f1")
```

would generate:

```
#' Initialize LFDA.
#'
#'
#'
#' @param num_dims Dimensionality of reduced space (defaults to dimension of X)
#' @param k Number of nearest neighbors used in local scaling method.
#' Defaults to min(7, num_dims - 1).
#' @param embedding_type Type of metric in the embedding space (default: 'weighted')
#'
#' 'weighted'        - weighted eigenvectors
#' 'orthonormalized' - orthonormalized
#' 'plain'           - raw eigenvectors
#'
#' @return values : array-like
#'
#' @export
f1 <- function(num_dims = NULL, k = NULL, embedding_type = "weighted") {
  LFDA$LFDA$f1(
    num_dims = num_dims,
    k = k,
    embedding_type = embedding_type
  )
}
```

Output is not perfect yet but should be a good start.

### DIFF
--- a/R/help.R
+++ b/R/help.R
@@ -89,20 +89,14 @@ help_completion_handler.python.builtin.object <- function(topic, source) {
   if (is.null(doc))
     doc <- ""
   
-  arguments_matches <- regexpr(pattern = "\nParameters\n+[-]+", doc)
+  arguments_matches <- regexpr(pattern = sphinx_doc_params_pattern(), doc)
   if (arguments_matches[[1]] != -1) {
     # Docs in Sphinx style
     doctree <- sphinx_doctree_from_doc(doc)
-    returns <- doctree$ids$returns$astext()
+    returns <- gsub("Returns\n", "", doctree$ids$returns$astext(), fixed = TRUE)
     description <- substring(doc, 1, arguments_matches[[1]])
-    
-    params <- doctree$ids$parameters$children[[2]]$children
-    params_kv_pairs <- lapply(params, function(param) {
-      list(
-        term = param$children[[1]]$astext(),
-        definition = param$children[[3]]$astext())
-    })
-    sections <- lapply(names(doctree$ids), function(name) if (!name %in% c("parameters", "returns")) doctree$ids[[name]])
+    sections <- lapply(names(doctree$ids), function(name) 
+      if (!name %in% c("parameters", "returns")) doctree$ids[[name]])
     sections[sapply(sections, is.null)] <- NULL
   } else {
     # Docs in other styles, e.g. TensorFlow
@@ -253,18 +247,20 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
   NULL
 }
 
+sphinx_doc_params_pattern <- function() { "\nParameters\n+[-]+" }
+
 is_sphinx_doc <- function(doc) {
-  arguments_matches <- regexpr(pattern = "\nParameters\n+[-]+", doc)
+  arguments_matches <- regexpr(pattern = sphinx_doc_params_pattern(), doc)
   arguments_matches[[1]] != -1
 }
 
 sphinx_doctree_from_doc <- function(doc) {
   module_name <- "docutils"
   if (py_module_available(module_name)) {
-    docutils <- import("docutils")
+    docutils <- import(module_name)
     docutils$core$publish_doctree(doc)
   } else {
-    stop("docutils module needs to be installed to extract doc from Sphinx style documentation.")
+    stop(paste0(module_name, " module needs to be installed to extract doc from Sphinx style documentation."))
   }
 }
 
--- a/R/help.R
+++ b/R/help.R
@@ -89,12 +89,11 @@ help_completion_handler.python.builtin.object <- function(topic, source) {
   if (is.null(doc))
     doc <- ""
   
-  arguments_matches <- regexpr(pattern = sphinx_doc_params_pattern(), doc)
-  if (arguments_matches[[1]] != -1) {
+  if (is_sphinx_doc(doc)) {
     # Docs in Sphinx style
     doctree <- sphinx_doctree_from_doc(doc)
     returns <- gsub("Returns\n", "", doctree$ids$returns$astext(), fixed = TRUE)
-    description <- substring(doc, 1, arguments_matches[[1]])
+    description <- substring(doc, 1, sphinx_doc_params_matches(doc)[[1]])
     sections <- lapply(names(doctree$ids), function(name) 
       if (!name %in% c("parameters", "returns")) doctree$ids[[name]])
     sections[sapply(sections, is.null)] <- NULL
@@ -249,9 +248,12 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
 
 sphinx_doc_params_pattern <- function() { "\nParameters\n+[-]+" }
 
+sphinx_doc_params_matches <- function(doc) {
+  regexpr(pattern = sphinx_doc_params_pattern(), doc)
+}
+
 is_sphinx_doc <- function(doc) {
-  arguments_matches <- regexpr(pattern = sphinx_doc_params_pattern(), doc)
-  arguments_matches[[1]] != -1
+  sphinx_doc_params_matches(doc)[[1]] != -1
 }
 
 sphinx_doctree_from_doc <- function(doc) {
--- a/R/help.R
+++ b/R/help.R
@@ -91,21 +91,31 @@ help_completion_handler.python.builtin.object <- function(topic, source) {
   
   arguments_matches <- regexpr(pattern = "\nParameters\n+[-]+", doc)
   if (arguments_matches[[1]] != -1) {
-    # Sphinx style docs
-    docutils <- import("docutils.core")
-    doctree <- docutils$publish_doctree(doc)
+    # Docs in Sphinx style
+    docutils <- import("docutils")
+    doctree <- docutils$core$publish_doctree(doc)
     returns <- doctree$ids$returns$astext()
     description <- substring(doc, 1, arguments_matches[[1]])
+    
+    # Get XML representation so it's easier to work with
+    # etree <- import("xml.etree.ElementTree")
+    # doctree <- etree$fromstring(doctree$ids$parameters$asdom()$toxml())
+    params <- doctree$ids$parameters$children[[2]]$children
+    params_kv_pairs <- lapply(params, function(param) {
+      list(
+        term = param$children[[1]]$astext(),
+        definition = param$children[[3]]$astext())
+    })
+    lapply(names(doctree$ids), function(name) if (!name %in% c("parameters", "returns")) doctree$ids[[name]])
   } else {
+    # Docs in other styles, e.g. TensorFlow
     # extract preamble
     arguments_matches <- regexpr(pattern = '\n(Arg(s|uments):)', doc)
     if (arguments_matches[[1]] != -1)
       description <- substring(doc, 1, arguments_matches[[1]])
     else
       description <- doc
-    
-    
-    
+
     # collect other sections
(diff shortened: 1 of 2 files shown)


R/help.R
```diff
@@ -238,33 +246,63 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
...
+  sphinx_doc_params_matches(doc)[[1]] != -1
+}
+
+sphinx_doctree_from_doc <- function(doc) {
+  module_name <- "docutils"
+  if (py_module_available(module_name)) {
+    docutils <- import(module_name)
+    docutils$core$publish_doctree(doc)
+  } else {
+    stop(paste0(module_name, " module needs to be installed to extract doc from Sphinx style documentation."))
```
terrytangyuan: This is probably `NULL` instead of `stop()`. But is there a way handle this in help handlers so no help is shown in this case instead of failing for users?  @jjallaire

jjallaire:
Thanks for this! I might not be able to merge it for a while because this codepath is hit by the IDE during code completion and I don't want to risk any unexpected error conditions there in the v1.6 release (which we'll be blogging about and encouraging people to kick the tires on more than we have in the past).

terrytangyuan:
No problem! Note that many packages are documented in Sphinx style so users
should expect that part of the help handlers won’t work for those packages.
It might be good to point this out in the blog.

jjallaire:
Well that does give us a bit more motivation to take the PR sooner. I don't have time to review it in the near term. @kevinushey I know you are busy with lots of other reticulate stuff. If you have time to review and merge this for the v1.6 release that's great, otherwise we will wait for v1.7.

R/help.R
```diff
@@ -238,33 +248,63 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
   NULL
 }
 
+sphinx_doc_params_pattern <- function() { "\nParameters\n+[-]+" }
```
kevinushey: We can probably just use the regular expression directly (no need for a separate function since it's not called anywhere else)

kevinushey:
I'll try to take a closer look tomorrow / Friday.

jjallaire:
Okay, that's great. I think that if we can take this PR it's definitely a big plus. My main concern is eliminating the possibility of runtime errors occurring during code completion.

terrytangyuan:
Yes that is my main concern as well. I need some guidance on what the best way is to avoid that since I am not familiar with the IDE APIs. Ideally, I'd like to display a friendly message in help saying "docutils module not available (let users choose whether to install or not) or some failure happened". Or maybe we shouldn't display anything at all in that case and give instructions on this in the documentation.

R/help.R
kevinushey: Shouldn't a return of `NULL` effectively do that? The help handler does that early on when checking for e.g. Python availability.

Is there some sane fallback we can use for Sphinx documentation when `docutils` is not available? (Does the default help handler do something sensible in that case?)

kevinushey:
Could we wrap the help handler implementations in something like:

```R
help <- tryCatch(help_handler(), error = identity)
if (inherits(help, "error"))
  return(NULL)
```

just to ensure that any errors that do occur are caught and don't propagate up to the user?

jjallaire:
I thought of that, my only concern is that this guarantees that these
errors will never be reported :-\

R/help.R
```diff
@@ -89,12 +89,36 @@ help_completion_handler.python.builtin.object <- function(topic, source) {
   if (is.null(doc))
     doc <- ""
   
-  # extract preamble
-  arguments_matches <- regexpr(pattern ='\nArg(s|uments):', doc)
-  if (arguments_matches[[1]] != -1)
-    description <- substring(doc, 1, arguments_matches[[1]])
-  else
-    description <- doc
+  if (is_sphinx_doc(doc)) {
```
kevinushey: I think we should check if both this is a Sphinx document and whether `docutils` is available before entering this code path. (In other words, we only attempt Sphinx-style parsing if we know the tools for it are available).

In addition, I wonder if we can use our default help codepaths for Sphinx documentation -- does it at least do something sensible?

R/help.R
```diff
@@ -238,33 +248,63 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
...
-      arg_description <- substring(line, regexpr(':', line)[[1]] + 1)
-      next_line <- arg_line + 1
-      while((arg_line + 1) <= length(doc)) {
-        line <- doc[[arg_line + 1]]
-        if (!grepl("^\\s*$", line) && !grepl("^\\s+\\w+: ", line)) {
-          arg_description <- paste(arg_description, line)
-          arg_line <- arg_line + 1
+  if (is.null(doc)) {
+    arg_descriptions <- args
+  } else if (is_sphinx_doc(doc)) {
```
kevinushey: Same comment here re: checking for `docutils` before entering this code path

R/help.R
```diff
@@ -238,33 +248,63 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
...
+
+is_sphinx_doc <- function(doc) {
+  sphinx_doc_params_matches(doc)[[1]] != -1
+}
+
+sphinx_doctree_from_doc <- function(doc) {
+  module_name <- "docutils"
+  if (py_module_available(module_name)) {
+    docutils <- import(module_name)
+    docutils$core$publish_doctree(doc)
```
kevinushey: Could this ever fail? (documentation cannot be parsed; differences in Sphinx documentation structure between different versions of `docutils`)

R/help.R
```diff
@@ -238,33 +248,63 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
...
-      while((arg_line + 1) <= length(doc)) {
-        line <- doc[[arg_line + 1]]
-        if (!grepl("^\\s*$", line) && !grepl("^\\s+\\w+: ", line)) {
-          arg_description <- paste(arg_description, line)
-          arg_line <- arg_line + 1
+  if (is.null(doc)) {
+    arg_descriptions <- args
+  } else if (is_sphinx_doc(doc)) {
+    doctree <- sphinx_doctree_from_doc(doc)
+    params <- doctree$ids$parameters$children[[2]]$children
```
kevinushey: Could this ever fail? (I think this is overall just an argument for us to wrap everything in `tryCatch()` to catch any unanticipated errors while calling Python code)

terrytangyuan: I believe these are relatively stable since Sphinx cannot break backward compatibility like that. But we definitely need try catch to wrap everything.

kevinushey:
I think we would still get error reports; it would just be of the form "Python help doesn't show for my object / class" rather than "R error emitted when I try to request help". Although it's probably easier to ignore the first rather than the second.

R/help.R
terrytangyuan: yes, at least the arguments/auto-completion could get populated with no problem since those are obtained from our Python code instead of from the documentation.

R/help.R
terrytangyuan: I will have to test that out. If the default help handler does something sensible, should we use the default help handler as the fallback when 1) docutils is not available or 2) any exceptions happen during the runtime of sphinx help handler?

kevinushey: I would vote for both cases just so the user always gets some help (but hopefully perceptive users will notice if / when Sphinx documentation isn't show as they except and let us know)

terrytangyuan: So we don’t display any message at all? I am still not sure how to display it if needed anyways.

terrytangyuan: Just verified the default help handlers (before this PR) gives the following for functions documented in Sphinx style:

screen shot 2018-03-01 at 5 25 17 pm

So I believe it is good to just fallback to the default ones if any of 1) or 2) happens. 

@jjallaire @kevinushey Could you confirm or let me know if there's better way to handle this? Also, what's the best way to let users know that they should install `docutils` module?

terrytangyuan: Two ideas:

a) Display the note "...need to install docutils..." under each parameter description.
b) A warning message for "...need to install docutils..." when the user is trying to look it up.

jjallaire: I don't think a user will know what to do with "need to install docutils" so this seems close to a message that will mostly be frustrating rather than helpful. Do we not expect that this will be commonly available?

terrytangyuan: It maybe available through some conda setup or through other dependencies.We could give them better instructions, e.g. telling them to `pip install docutils` or give them a method similar to `install_tensorflow()` or `install_keras()`. 

Or we could just simply fallback and (as Kevin said) hopefully perceptive users will notice if / when Sphinx documentation isn't shown as they except and let us know.

terrytangyuan:
@kevinushey @jjallaire I updated the PR to address your comments/concerns:
* Wraps Sphinx handlers with `tryCatch`, which means if any failures happen during Sphinx documentation extract, it will fallback to default handlers (which is what reticulate used to behave, e.g. working for TensorFlow docs)
* If `docutils` is not available, we use the default handlers (same as above)

I've done some testing using TensorFlow docs, pandas docs (Sphinx style), sklearn (Sphinx style), and some other manually drafted docs for edge cases. More testing from other people would be helpful. 

Only thing we need to address/discuss is:
* If `docutils` is not available, should we inform users? If so, how exactly do we want to do that?

jjallaire:
I think that telling users to install docutils is too much inside baseball and will cause confusion / increased support tickets. If we don't have docutils let's just silently fall back.

terrytangyuan:
Sounds good. @kevinushey  I think this PR is ready for another review whenever you get a chance. Thank you for your time!

R/help.R
```diff
@@ -66,6 +66,51 @@ register_module_help_handler <- function(module, handler) {
   .module_help_handlers[[module]] <- handler
 }
 
+# Extract docs in TensorFlow-like styles
+help_completion_handler_default <- function(doc) {
+  arguments_matches <- regexpr(pattern = '\n(Arg(s|uments):)', doc)
+  if (arguments_matches[[1]] != -1)
```
kevinushey: Just curious why you use `[[1]]` indexing here? From what I understand, `regexpr()` only returns the first match, and so the return value should always be a length-one vector (although with some extra attributes). However, those attributes won't be used during comparison with `==` or `!=`.

I think the existing code is still correct but curious if there is some other motivation.

jjallaire:
@kevinushey I'll let you go ahead and do the merge (I don't want to merge if there's one last thing you want tweaked)

PR #169, opened 2018-03-01
use tokenizer for more deterministic autocompletion

kevinushey:
This PR improves the autocompletion support by adding a Python code tokenizer, which is used to tokenize the code before attempting a completion.

Using this, we can more effectively autocomplete things like function argument names.

This is kind of a big scary PR but it borrows most of the ideas that have worked in RStudio's tokenizer + completion system -- they're just translated to R.

(This work sort of makes me think that somewhere down the road we should provide this kind of tokenizer support in a separate R package that other completers could make use of)

### DIFF
--- a/R/completer.R
+++ b/R/completer.R
@@ -1,293 +1,247 @@
-# basic autocompletion support (used for the Python REPL)
-py_completer <- function(line) {
- 
-  # helper function for constructing a regular expression pattern from token
-  pattern <- function(token) { paste("^\\Q", token, "\\E", sep = "") }
+py_completions <- function(token, candidates) {
+  pattern <- paste("^\\Q", token, "\\E", sep = "")
+  completions <- sort(grep(pattern, candidates, perl = TRUE, value = TRUE))
+  attr(completions, "token") <- token
+  completions
+}
+
+py_complete_none <- function() {
+  character()
+}
+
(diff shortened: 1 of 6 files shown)


jjallaire:
I am fine with this PR but of course I don't have the context to review it properly.

One question that comes to mind is whether it's possible to move any of these into RStudio at least while we are in development mode (because updating reticulate to stay in sync with the IDE is going to pretty much be impossible).

kevinushey:
I originally wanted this to be in `reticulate` so that autocompletion could work within a plain R session (not just RStudio). If we think that's a bit too ambitious I could move this back over to the RStudio side.

jjallaire:
I think we should move it into RStudio while we are developing and debugging it, and then move it back into reticulate when we feel like it's stabilized.

kevinushey:
Okay, sounds good. I'll start moving things over tomorrow.

Should we just take out the custom completer altogether in `reticulate` for now then?

jjallaire:
Yes, let's take it out entirely (other than whatever hinting we need to do in order to trigger the IDE code) then put it back in after it stabilizes within the IDE. I know that the completer shares some code with the REPL, hopefully that's easy enough to provide helpers to the IDE for.

PR #175, opened 2018-03-06
document extensibility of conversion operators

kevinushey:
This PR adds documentation related to the extensibility of conversion operators in the packages vignette, mainly letting users know that we handle conversion of builtin Python objects, NumPy arrays, and Pandas objects.

### DIFF
--- a/vignettes/package.Rmd
+++ b/vignettes/package.Rmd
@@ -98,6 +98,19 @@ py_str.MyModule.MyPythonClass <- function(object, ...) {
 
 The `print` and `summary` methods for Python objects both call the `str` method by default, so if you implement `py_str()` you will automatically inherit implementations for those methods.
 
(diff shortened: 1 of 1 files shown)


jjallaire:
A related question: if a user attempts to pass an R data frame to Python and pandas is not installed, what sort of error message do they get?

kevinushey:
Right now they don't get an error -- instead, they just get a dictionary (ie we fallback to the default converter when Pandas is not installed).

I wonder if we should print a warning or similar?

kevinushey:
We could also make that an error and suggest that if they really want to convert to a Python dictionary the user should call `as.list()` on the `data.frame` first.

jjallaire:
No if it has default semantics to become a dictionary that's fine.

PR #180, opened 2018-03-10
Fixes for `py_iterator` doc

terrytangyuan:
This fixes the formatting of docs for `py_iterator` and moved `py_iterator` to "Python Types" section of the reference page. 

Originally (note the additional indentation before `g` and `generator` should be `py_iterator`):
```
sequence_generator <-function(start) {
  value <- start
  function() {
    value <<- value + 1
    value
  }
}
    g <- generator(sequence_generator(10))
```

The unrelated diff is probably because I am using the newer version of pkgdown (just did a fresh install from Github).

### DIFF
--- a/R/generator.R
+++ b/R/generator.R
@@ -17,7 +17,7 @@
 #' mutates it's enclosing environment via the `<<-` operator. For example:
 #'
 #' ```r
-#' sequence_generator <-function(start) {
+#' sequence_generator <- function(start) {
 #'   value <- start
 #'   function() {
 #'     value <<- value + 1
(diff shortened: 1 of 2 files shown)


jjallaire:
Could you submit another PR with just this commit: https://github.com/rstudio/reticulate/pull/180/commits/ff284fe48283f0e5bd96446f302358f5a82907cf

(I'll do the regeneration as it looks like you might be using a more recent version of pkgdown that has caused a bunch of other files to change)

terrytangyuan:
@jjallaire Got it. Done!

Issue #185, opened 2018-03-15
Memory leak in assigning R variables to Python space


david-liu-brattle:
When assigning values from R to Python's &#95;&#95;main&#95;&#95; with

`py$x <- rVar`

each iterative assignment consumes more memory until the system runs out of memory and it crashes.  To reproduce:

```
library(reticulate)
main <- import_main(convert = FALSE)

for(i in 1:10){
  main$x <- rep(0,5000000)
  py_run_string("
del x
import gc
gc.collect()")
}
```
Each iteration of this loop increases memory use of rsession by ~200MB which is never released.  In comparison, if we generate the variable inside Python it does not cause the same issue:

```
library(reticulate)
py_run_string("")

for(i in 1:10){
  py_run_string(
  "
x=[0]*5000000
del x
import gc
gc.collect()")
}
```

Similarly, calling a Python function on a large R variable does not cause the issue either.  As a result, the following workaround prevents memory leaks:

```
py_run_string(
'
def load_global(x,n):
    globals()[n]=x
'
)
py$load_global(rep(0,5000000),'x')
```

jjallaire:
@kevinushey let’s take a look at this right away (definitely let me know if there are things about reticulate memory management I can further explain, I probably can’t talk real-time until Monday so if it as to wait until then that’s fine)

kevinushey:
Is it possible we're missing a `Py_DECREF()` call after our call(s) to `PyObject_SetAttrString()`? For example:

https://github.com/python/cpython/blob/d6e140466142018ddbb7541185348be2b833a2ce/Objects/moduleobject.c#L152-L156

Or do objects created from `r_to_py()` not need a `Py_DECREF` by default?

We use it here:

https://github.com/rstudio/reticulate/blob/0abad27db42241efd12f2ad0bcb4b8ea723c74ab/src/python.cpp#L1687-L1692

kevinushey:
If I call `Py_DecRef()` after assigning the object into the module the leak seems to go away.

jjallaire:
That sounds promising! I’ll take a closer look tomorrow or Monday

jingshuw:
Hi, is this issue solved? I download the package from CRAN recently and I still face the same problem. Should I download from Github?

jjallaire:
[...]

PR #187, opened 2018-03-16
manage refcount in py_set_attr

kevinushey:
Closes https://github.com/rstudio/reticulate/issues/185.

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -1686,9 +1686,8 @@ PyObjectRef py_get_attr_impl(PyObjectRef x, const std::string& name, bool silent
 
 // [[Rcpp::export]]
 void py_set_attr_impl(PyObjectRef x, const std::string& name, RObject value) {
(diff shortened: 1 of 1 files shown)


jjallaire:
This is definitely the right fix! Note that I changed the code to use the `PyObjectPtr` wrapper (which does decref on destruct).

kevinushey:
Sounds good!

Is it possible this issue might be biting us elsewhere? Do we typically need to explicitly manage the lifetime of a Python object returned by `r_to_py()` or do most Python APIs 'steal' the reference or similar?

jjallaire:
Some steal it and some don't. If you do a Find in Files you'll see that we always use PyObjectPtr and in the cases we don't there is an explicit comment indicating that the reference is stolen, e.g. https://github.com/rstudio/reticulate/blob/bugfix/py-set-attr-refcounting/src/python.cpp#L1759-L1767

PR #188, opened 2018-03-16
Added generic methods for pandas objects

terrytangyuan:
Currently, these generic methods don't work correctly:

```
> df <- data.frame(x = c(1, 3), y = c(4, 4), z = c(5, 5))
> pdf <- r_to_py(df)
> dim(pdf)
NULL
> length(pdf)
[1] 2
> summary(pdf)
     x    y    z
0  1.0  4.0  5.0
1  3.0  4.0  5.0
```

I added the corresponding generic methods to make sure they work correctly.


jjallaire:
For any S3 method we implement for a Python object we need to be aware that the method may be called on a `NULL` externalptr (i.e. if the object has been serialized and then brought back into a fresh R session). To work around this we need this code which will yield some sensible default:

https://github.com/rstudio/reticulate/blob/master/R/python.R#L315-L320

You need to check both for null xptr and for whether py is available because we don't want RStudio calling summary, dim, length, etc. from the workspace pane to cause Python to be initialized (thus bypassing the `import` mechanism and binding to the wrong version of Python.

terrytangyuan:
@jjallaire Good call. I just pushed the change. Is it what you had in mind?

R/conversion.R
```diff
@@ -160,7 +160,29 @@ py_to_r.pandas.core.series.Series <- function(x) {
   py_to_r(x$as_matrix())
 }
 
+#' @export
+summary.pandas.core.series.Series <- function(object, ...) {
+  if (py_is_null_xptr(object) || !py_available())
+    str(object)
```
terrytangyuan: I wasn't sure what the best default summary this should be but I believe this was the default prior to this PR (see the example in the PR description).

jjallaire: This should now print this when the xptr is NUlL:

```r
<pointer: 0x0>
```

Can you confirm that is what you see as well as test the other methods with null xptrs for both series and data frame? (extra important to test these as if we don't get them right they can cause crashing and/or binding to the wrong version of Python).

jjallaire:
@kevinushey Could you also review this just for a second set of eyes?

R/conversion.R
terrytangyuan: Yes, I confirmed by using `saveRDS` and `readRDS` on both pandas data frame and series. They yield the expected results:
```
> pdf
<pointer: 0x0>
> dim(pdf)
NULL
> length(pdf)
[1] 0
> summary(pdf)
<pointer: 0x0>
```

jjallaire: okay, we are good to merge then once @kevinushey has a look

R/conversion.R
```diff
@@ -160,7 +160,29 @@ py_to_r.pandas.core.series.Series <- function(x) {
...
+  else
+    object$describe()
+}
+
+#' @export
+length.pandas.core.series.Series <- function(x) {
+  if (py_is_null_xptr(x) || !py_available())
+    0L
+  else
+    py_to_r(x$size)
```
kevinushey: You might want to use `as_r_value(x$size)` since this might auto-convert depending on how the Series object was constructed.

R/conversion.R
```diff
@@ -160,7 +160,29 @@ py_to_r.pandas.core.series.Series <- function(x) {
...
+}
 
+#' @export
+dim.pandas.core.series.Series <- function(x) {
+  if (py_is_null_xptr(x) || !py_available())
+    NULL
+  else
+    unlist(py_to_r(x$shape))
```
kevinushey: Same `as_r_value()` comment here.

R/conversion.R
```diff
@@ -294,3 +316,18 @@ py_to_r.pandas.core.frame.DataFrame <- function(x) {
   df
   
 }
+
+#' @export
+summary.pandas.core.frame.DataFrame <- function(object, ...) {
+  summary.pandas.core.series.Series(object, ...)
```
kevinushey: FWIW this doesn't seem like a very natural delegation (although the code is correct). Why not just have this and the other `summary` method call a helper `py_describe()` function?

R/conversion.R
```diff
@@ -294,3 +316,18 @@ py_to_r.pandas.core.frame.DataFrame <- function(x) {
...
+}
+
+#' @export
+length.pandas.core.frame.DataFrame <- function(x) {
+  length.pandas.core.series.Series(x)
```
kevinushey: With an R `data.frame`, `length()` gives the number of columns. I think this would give the total number of elements?

R/conversion.R
```diff
@@ -294,3 +316,18 @@ py_to_r.pandas.core.frame.DataFrame <- function(x) {
...
+}
+
+#' @export
+dim.pandas.core.frame.DataFrame <- function(x) {
+  dim.pandas.core.series.Series(x)
```
kevinushey: Similar comment here (have a helper `py_shape()` function that these `dim` methods both call)

R/conversion.R
terrytangyuan: Oh wow I didn't know that before. Should I change this to match the data.frame behavior? Interestingly `length()` for an R matrix is the number of total elements.

R/conversion.R
jjallaire: The other thing we do elsewhere in the package is just assign the method we will delegate to directly, e.g. https://github.com/rstudio/reticulate/blob/master/R/python.R#L285-L286

R/conversion.R
kevinushey: I think this should match the data.frame behavior.

For some context, this is because a data.frame is really just an R list under the hood, and so `length()` counts the number of elements in that list (non-recursively).

Matrices on the other hand are just R vectors with an extra dimension attribute, so length does give the total number of elements (pretending the matrix were just a long vector)

terrytangyuan: Ok I'll change this. I doubt users will use length anyways. This is just to make sure the bahavior matches with data.frame. Users usually just use nrow, ncol, or dim as far as I know.

R/conversion.R
terrytangyuan: I like JJ's idea. @kevinushey Do you mind if I adopt this and delegate directly?

kevinushey: Sounds good!

R/conversion.R
```diff
@@ -160,7 +160,29 @@ py_to_r.pandas.core.series.Series <- function(x) {
...
+length.pandas.core.series.Series <- function(x) {
+  if (py_is_null_xptr(x) || !py_available())
+    0L
+  else {
+    pandas_shape(x)[[1]]
+  }
+}
+
+#' @export
+dim.pandas.core.series.Series <- function(x) {
```
jjallaire: Just want to confirm that the `NULL` return here is intentional

terrytangyuan:
@kevinushey Could you take another look as I made some additional change in order to match the behavior (e.g. `dim.pandas.core.series.Series` always returns `NULL` similar to an R vector)? The unit tests results now match the result of data.frame.

R/conversion.R
terrytangyuan: Yes, see my comment below.

terrytangyuan: @jjallaire Travis is having issues. Could you re-start it (I don't have write access to this repo)?

jjallaire:
I think we should also have an `as.data.frame` method for Pandas objects (nicer than calling `py_to_r` explicitly. These would also need the py_is_null_xptr check.

terrytangyuan:
That's a good idea. Do you think it's necessary to add a  `as.matrix()` too (might just be a simple conversion to data.frame and then call as.matrix)? These are obviously simple functions to developers but to users these might be more friendly.

PR #190, opened 2018-03-19
allow for parameter help from object directly

kevinushey:
This PR makes it possible to get help from a Python object directly, rather than just from the code that can be evaluated to produce the associated object.

This PR also augments the lookup logic such that the `__init__` method is used to provide parameter help for class objects.

### DIFF
--- a/R/help.R
+++ b/R/help.R
@@ -248,7 +248,6 @@ help_url_handler.python.builtin.object <- function(topic, source) {
   }
 
   # get help page
-  page <- NULL
   inspect <- import("inspect")
   if (inspect$ismodule(source)) {
     module <- py_get_name(source)
@@ -348,7 +347,7 @@ arg_descriptions_from_doc_sphinx <- function(doc) {
   
   nms <- vapply(params, function(param) {
     name <- param$children[[1]]$astext()
-    sub(":.*", "", name)
+    sub("\\s*:.*", "", name)
   }, character(1))
   
   names(output) <- nms
@@ -390,7 +389,7 @@ sections_from_doc <- function(doc) {
     
     # collect the sections text
     section_text <- c()
-    while((section_line + 1) <= length(doc)) {
+    while ((section_line + 1) <= length(doc)) {
       line <- doc[[section_line + 1]]
       if (grepl("\\w+", line)) {
         section_text <- paste(section_text, line)
--- a/R/help.R
+++ b/R/help.R
@@ -291,7 +291,7 @@ help_formals_handler.python.builtin.object <- function(topic, source) {
 }
 
 sphinx_doc_params_matches <- function(doc) {
-  regexpr(pattern = "\nParameters\n+[-]+", doc)
+  regexpr(pattern = "(?:^|\n)Parameters\n+[-]+", doc)
 }
 
 is_sphinx_doc <- function(doc) {
@@ -309,17 +309,17 @@ sphinx_doctree_from_doc <- function(doc) {
 
 # Extract arguments descriptions for docs in TensorFlow-like styles
 arg_descriptions_from_doc_default <- function(args, doc) {
+  
   # extract arguments section of the doc and break into lines
-  arguments <- section_from_doc('Arg(s|uments)', doc)
   doc <- strsplit(doc, "\n", fixed = TRUE)[[1]]
   
   sapply(args, function(arg) {
-    arg_line <- which(grepl(paste0("^\\s+", arg, ":"), doc))
+    arg_line <- which(grepl(paste0("^\\s*", arg, ":"), doc))
     if (length(arg_line) > 0) {
       line <- doc[[arg_line]]
       arg_description <- substring(line, regexpr(':', line)[[1]] + 1)
       next_line <- arg_line + 1
-      while((arg_line + 1) <= length(doc)) {
+      while ((arg_line + 1) <= length(doc)) {
         line <- doc[[arg_line + 1]]
         if (!grepl("^\\s*$", line) && !grepl("^\\s+\\w+: ", line)) {
           arg_description <- paste(arg_description, line)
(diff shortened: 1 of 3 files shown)


jjallaire:
@terrytangyuan Could you review this once since you've recently made changes there?

Very close to branching for CRAN release of v1.6 so an extra level of scrutiny is required here!

terrytangyuan:
@jjallaire Sure! I will do a full review soon. In the meantime, @kevinushey do you mind adding some tests?

kevinushey:
When running tests locally I was getting a bunch of failures in the parsing of Sphinx documentation, so I applied a few fixes on this PR as well. Can you review?

R/help.R
```diff
@@ -323,11 +336,22 @@ arg_descriptions_from_doc_default <- function(args, doc) {
...
 arg_descriptions_from_doc_sphinx <- function(doc) {
+  
   doctree <- sphinx_doctree_from_doc(doc)
   params <- doctree$ids$parameters$children[[2]]$children
-  sapply(params, function(param) {
-    param$children[[3]]$astext()
+  
+  output <- lapply(params, function(param) {
+    children <- param$children
+    children[[length(children)]]$astext()
```
kevinushey: I was seeing that in some cases we had two children in this node rather than three. I'm guessing the last entry is always the description, so we just grab that rather than hardcoding the number.

R/help.R
```diff
@@ -323,11 +336,22 @@ arg_descriptions_from_doc_default <- function(args, doc) {
...
   })
+  
+  nms <- vapply(params, function(param) {
+    name <- param$children[[1]]$astext()
```
kevinushey: It looks like the first entry can capture additional 'stuff' (e.g. the type declaration) so I think we need to strip that off.

tests/testthat/test-help-handlers.R
```diff
@@ -26,7 +26,7 @@ Just a placeholder here.
 "
 
   doctree <- sphinx_doctree_from_doc(docs)
-  expect_equal(names(doctree$ids), c("parameters", "returns", "section1"))
+  expect_setequal(names(doctree$ids), c("parameters", "returns", "section1"))
```
kevinushey: It seems like the ordering of names is not consistent, so using `setequal()` here seemed more appropriate.

R/help.R
```diff
@@ -185,17 +185,31 @@ help_completion_parameter_handler.python.builtin.object <- function(source) {
...
+    
+    # get the function
+    target <- help_get_attribute(source, topic)
+  }
+  
+  # if this is a class object, use the __init__ method
+  if (inherits(target, "python.builtin.type")) {
+    init <- py_get_attr(target, "__init__", silent = TRUE)
+    if (!is.null(init))
+      target <- init
```
terrytangyuan: We had something similar to this for the wrapper generation functions. But one potential issue could be, e.g. this [Estimator class](https://github.com/tensorflow/tensorflow/blob/e3d63c81f15f962c0841b3a6f97b7df93b0343cf/tensorflow/python/estimator/estimator.py#L70) that we are familiar with have docstrings in both `__init__` and the class. This is very common in other libraries too. I think we can:
1. concatenate both docs for the `description` field that we are extracting in some way
2. extract argument descriptions from `__init__` for classes.

terrytangyuan:
@kevinushey Thanks for the comments on your new code! The change looks good to me except for one comment for potential issue (actually quite common scenario).

R/help.R
terrytangyuan: The reason for (1) is that if we only extract doc string from `__init__`, this will be the only thing showing up "Constructs an `Estimator` instance.". However, users won't know what an `Estimator` is, which is where they need to look up the doc-string for `Estimator` class in addition to `__init__`.
 
I saw in RStudio there's a "press F1 for additional help" when the help pops up. Instead of concatenate as I suggested in (1), Is there a way for us to only show the class descriptions when pressing F1? Sometimes the descriptions are quite long.

tests/testthat/test-help-handlers.R
```diff
@@ -69,3 +69,27 @@ Section1:
...
+Parameters
+----------
+
+alpha: int, optional
+  The first argument.
+
+omega: int, optional
+  The last argument.
+"""
+  pass
```
terrytangyuan: This might be a bit nitpick but could you indent the docs to match with the indentation level of `pass`?

jjallaire:
One word of caution here: we are going to end up publishing this commit to CRAN very soon and I don't want to risk code that could throw runtime errors during RStudio completion. So let's try to get things right but let's not let our codepaths get so complex that we can't reason about whether errors might occur. I'd rather have completion be incomplete than have runtime errors showing up during editing.

kevinushey:
@jjallaire perhaps out of an abundance of caution I should just implement our own help handlers in RStudio rather than trying to get reticulate patches in so close to CRAN release?

jjallaire:
I think that's a good idea as I'd really like to the CRAN release today or tomorrow.

That said, I'd like to merge the code back into reticulate after we've had enough miles of testing underneath it within the IDE (so let's keep the code within the IDE as close as we can to what's in reticulate).

jjallaire:
Based on this discussion I am going to branch the current master of reticulate for the CRAN release and start final testing.

jjallaire:
@kevinushey Let me know once you've confirmed that you can do this all within RStudio (that will be my signal it's okay to submit reticulate to CRAN)

kevinushey:
@jjallaire we will indeed be able to implement everything we need on the IDE side.

jjallaire:
Okay, let's do that. I'll keep this PR open as a reminder that we want to merge that code back into reticulate.

terrytangyuan:
Now that [RStudio 1.2](https://blog.rstudio.com/2018/10/09/rstudio-1-2-preview-reticulated-python/) is there. Just curious - what's the status of this? Has the code been implemented on IDE side?

kevinushey:
I'll have to come back to this later -- I've lost the context around why we needed this and I'll have to see if all the IDE code that was added can be merged cleanly back to reticulate.

Issue #141, opened 2017-12-30
Incorrect Conversion of R Function Signature


terrytangyuan:
Many Python packages check the signature of the function that's being passed in as a parameter. However, reticulate doesn't convert the signature correctly. 

A couple of issues:

- Currently even though an R function does not have any keywords arguments, after it's converted to Python function, it will have keyword args. Some relevant code [here](https://github.com/rstudio/reticulate/blob/ff9373383a07c313fc7dae98d19ff17e8b6cdae4/src/python.cpp#L1292) and [here](https://github.com/rstudio/reticulate/blob/ff9373383a07c313fc7dae98d19ff17e8b6cdae4/inst/python/rpytools/call.py#L13). 
- The names of the function arguments are not reserved. Recall that we had to do [this](https://github.com/rstudio/tfestimators/blob/master/inst/python/estimatortools/functions.py) in tfestimators to make sure the signature is correct.

terrytangyuan:
An reproducible example that demonstrates this issue:

```
library(reticulate)

f1 <- function(a, b = 3) {
  a + b
}

util <- py_run_string("
def f1(a, b=3):
  return a + b
")

inspect <- import("inspect")
inspect$getargspec(f1)
inspect$getargspec(util$f1)
```

```
> inspect$getargspec(f1)
ArgSpec(args=[], varargs='args', keywords='kwargs', defaults=None)
> inspect$getargspec(util$f1)
ArgSpec(args=['a', 'b'], varargs=None, keywords=None, defaults=(3,))
```
You can see that the R function `f1()` is not converted properly with correct signature. Ideally, the signature should match with the native Python function `util$f1()`.

This issue will prevent users from using  certain Python libraries where the argument checks are strict. Sometimes applying [the trick we did in tfestimators](https://github.com/rstudio/tfestimators/blob/master/inst/python/estimatortools/functions.py) to other Python libraries would not work either.

PR #195, opened 2018-03-26
Add `py_func()` to wrap R function in a Python function with correct signature

terrytangyuan:
This is to address https://github.com/rstudio/reticulate/issues/141, basically a generalized version of [`as_model_fn` in tfestimators](https://github.com/rstudio/tfestimators/blob/master/inst/python/estimatortools/functions.py).

Note that this function is experimental and not exported since currently only simple function is supported (e.g. no `...` or other complicated cases).


jjallaire:
Awesome! Let me take a closer look at this and see if there is a lower level way to do this that doesn't rely on the string interpolation (e.g. some of the approaches described here: https://stackoverflow.com/questions/11291242/python-dynamically-create-function-at-runtime)

terrytangyuan:
Thanks for taking the time to do that! I've done some experiments using other approaches but no luck (either not fully working or hard to customize). I agree that string interpolation is not ideal but I think it should support most of the use cases already and easy to customize. There's definitely a lot of space for improvements and additional test cases.

kevinushey:
Is there some well-defined set of functions that we can / should be able to correctly wrap? I think the following examples might fail in the current implementation:

```R
function(a = 1, b) {}
function(x = NA) {}
function(a.b) {}
function(a = "abc") {}
function(a = list()) {}
function(a = NULL) {}
```

Do any of these need to be handled?

terrytangyuan:
@kevinushey Or wrap the whole thing in a `tryCatch`

kevinushey:
Oh, I see -- you meant a regex for capturing the bad case? You could just use e.g. `regepxr(".", fixed = TRUE)` to catch that case. But I agree that we can just point to the docs and use `tryCatch()`.

terrytangyuan:
@kevinushey I was trying to apply your suggested R syntactical enhancements but for some reason I was hitting issues on those so I'll probably keep this as it is for now. I think the rest of your comments should be addressed already except that we need a better name for the function + export it.

jjallaire:
I think it should be name `py_func()`

terrytangyuan:
@jjallaire Done renaming. 
@kevinushey I just applied your suggested syntactical improvements.

Let me know if there's anything else I missed.

R/py_func.R
```diff
@@ -0,0 +1,47 @@
...
+        lapply(sigs, function(sig) quote(expr =)))
+    }
+    wrap_fn_util <- py_run_string(sprintf("
+def wrap_fn(f):
+  def fn(%s):
+    return f(%s)
+  return fn
+", func_signature, func_pass_args))
+    wrap_fn_util$wrap_fn(f)
+  }, error = function(e) r_to_py(f))
```
jjallaire: I don't think the `tryCatch` with fallback to the default behavior is really helping us here. The user *thinks* they got a complete Python signature but instead they end up getting the default `...` style signature. Why is this better than an error? (I'm presuming that the only reason the user is calling `py_func` is because they *require* the correct signature for use with a library that checks callbacks -- silently giving them an unexpected return value seems bad).

jjallaire:
Let's add a `NEWS.md` entry for this. Let's also add a brief section to this doc: https://github.com/rstudio/reticulate/blob/master/vignettes/calling_python.Rmd

R/py_func.R
terrytangyuan: Makes sense. I’ll change this

PR #206, opened 2018-03-29
Support conversion between `Matrix::dgCMatrix` and `scipy.sparse.csc.csc_matrix`

terrytangyuan:
Scipy CSC matrix is a common format for sparse matrix in Python libraries. Same for `Matrix::dgCMatrix` in R libraries. For example, XGBoost supports loading data from CSC matrix in its Python API and supports `Matrix::dgCMatrix` in its R API. 

I couldn't find a good way to do this in a low-level way but I think this could at least give a good start and can get people going. There's definitely some level of data copying going on through this high-level approach but I think this trade-off is okay since this could greatly speeds up model training.


R/conversion.R
```diff
@@ -336,3 +336,35 @@ dim.pandas.core.frame.DataFrame <- function(x) {
...
+        "Failed to convert: dimensions of the original dgCMatrix ",
+        "object and the converted CSC matrix do not match"))
+  csc_x
+}
+
+#' @export
+py_to_r.scipy.sparse.csc.csc_matrix <- function(x) {
+  disable_conversion_scope(x)
+  converted <- as_r_value(x$toarray())
+  if (require("Matrix"))
```
kevinushey: This should be `requireNamespace()` rather than `require()`, with calls from the `Matrix` package explicitly qualified (e.g. `Matrix::Matrix`).

This is because calling `require()` explicitly loads and attaches the R package to the user's search path, and you typically don't want to do that.

R/conversion.R
```diff
@@ -336,3 +336,35 @@ dim.pandas.core.frame.DataFrame <- function(x) {
   else
     pandas_shape(x)
 }
+
+# Conversion between `Matrix::dgCMatrix` and `scipy.sparse.csc.csc_matrix`.
+# Scipy CSC Matrix: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csc_matrix.html
+
+#' @export
+r_to_py.dgCMatrix <- function(x, convert = FALSE) {
```
kevinushey: I think we might need to import `Matrix` if we're going to define S3 methods for classes exported by that package

terrytangyuan: Should I move Matrix to `Imports` in DESCRIPTION and then remove the `require` statement?

jjallaire:
It seems like it would be safe to import `Matrix` since it's a recommend packages and as such will pretty much always be installed.

terrytangyuan:
Sounds good. Just fixed.

terrytangyuan:
This PR still has some TODOs that I will work on soon:
- Add an entry in NEWS
- Add small section in array/matrix vignette
- Add S3 methods to csc matrix class

Let me know in the meantime if there’s anything else that I should also take care of.

terrytangyuan:
@jjallaire @kevinushey Ready for another review whenever you get a chance this week.

jjallaire:
+ @bwlewis

jjallaire:
This LGTM but I'll let @kevinushey have the final review before we merge.

bwlewis:
Just a minor question:  Why not use sparseMatrix directly in https://github.com/rstudio/reticulate/pull/206/files#diff-8f5039eb2839438bfbe209fa4eb688ddR367  instead of `Matrix(as_r_value(x$toarray()))` ?

See, for instance the ancient prototype (pre-reticulate) example here:

https://github.com/bwlewis/python/blob/3200340f6df2401b91c7ab468b3bdc78d42a951b/R/internal.R

terrytangyuan:
@bwlewis Oh nice! You were thinking way ahead of me there! I guess I could do that too which may give a much better performance for large matrices. I just changed it to use `sparseMatrix` directly.

kevinushey:
LGTM!

PR #213, opened 2018-04-03
Add source_python() support to source script from URL

terrytangyuan:
`source()` can accepts a wider range of R `connection` objects but I think users of reticulate may only want to source file from local or from URL (e.g. Github snippet).

### DIFF
--- a/R/source.R
+++ b/R/source.R
@@ -15,10 +15,21 @@
 #' @export
 source_python <- function(file, envir = parent.frame(), convert = TRUE) {
   
+  # Copy lines from URL to a local tempory file
+  tmp_file <- NULL
+  if (!file.exists(file) && isTRUE(grepl("http", file))) {
+    lines <- readLines(file, warn = FALSE)
+    file <- tmp_file <- tempfile(fileext = ".py")
+    cat(lines, file = tmp_file, sep = "\n")
+  }
+
   # source the python script (locally so we can track what mutations are
   # made in the file scope)
   dict <- py_run_file(file, local = TRUE, convert = convert)
   
+  # Remove the temporarily created file (if any)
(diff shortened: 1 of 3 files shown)


R/source.R
```diff
@@ -15,10 +15,21 @@
 #' @export
 source_python <- function(file, envir = parent.frame(), convert = TRUE) {
   
+  # Copy lines from URL to a local tempory file
+  tmp_file <- NULL
+  if (!file.exists(file) && isTRUE(grepl("http", file))) {
+    lines <- readLines(file, warn = FALSE)
+    file <- tmp_file <- tempfile(fileext = ".py")
```
jjallaire: Here we can just do `on.exit(unlink(file), add = TRUE)` (then we don't need `tmp_file` nor do we need the conditional removal below)

terrytangyuan: Thanks! Looks much better now!

R/source.R
```diff
@@ -15,10 +15,18 @@
 #' @export
 source_python <- function(file, envir = parent.frame(), convert = TRUE) {
   
+  # Copy lines from URL to a local tempory file
+  if (!file.exists(file) && isTRUE(grepl("http", file))) {
+    lines <- readLines(file, warn = FALSE)
```
jjallaire: Why not use `download.file` (http://stat.ethz.ch/R-manual/R-devel/library/utils/html/download.file.html) here to just directly download the full file contents (as opposed to the lines treatment) e.g.

```r
tmpfile <- tempfile()
download.file('https://www.google.com', tmpfile)
```

terrytangyuan: Thanks! Just modified the code to use this instead!

PR #240, opened 2018-04-26
Use values rather than as_matrix() for column extraction prior to conversion attempt.

DavisVaughan:
This is a simple PR request to use `values` rather than `as_matrix()` in the conversion of Series and DataFrames column from python to R.

There are two reasons:

1) Pandas recommends it. See the last line [here](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.as_matrix.html).

2) Some columns are not fully reduced to their corresponding Numpy array format with `as_matrix()`, but are reduced with `values.` For example, time zone aware date cols:

(With Pandas >=0.20.3 where I'm pretty sure some of this changed)

``` r
library(reticulate)
pd <- import("pandas", convert = FALSE)

# time zone aware column! (time zone non-aware date cols work fine already)
rng = pd$date_range("2011-01-01", periods=2L, freq='H', tz = "UTC")
df <- pd$DataFrame(dict("datecol" = rng, "val" = list(1,2)))

# The as_matrix() method doesn't flatten to a ndarray
df$datecol$as_matrix()
#> DatetimeIndex(['2011-01-01 00:00:00+00:00', '2011-01-01 01:00:00+00:00'], dtype='datetime64[ns, UTC]', freq='H')

class(df$datecol$as_matrix())
#>  [1] "pandas.tseries.index.DatetimeIndex"       
#>  [2] "pandas.tseries.base.DatelikeOps"          
#>  [3] "pandas.tseries.base.TimelikeOps"          
#>  [4] "pandas.tseries.base.DatetimeIndexOpsMixin"
#>  [5] "pandas.indexes.numeric.Int64Index"        
#>  [6] "pandas.indexes.numeric.NumericIndex"      
#>  [7] "pandas.indexes.base.Index"                
#>  [8] "pandas.core.base.IndexOpsMixin"           
#>  [9] "pandas.core.strings.StringAccessorMixin"  
#> [10] "pandas.core.base.PandasObject"            
#> [11] "pandas.core.base.StringMixin"             
#> [12] "python.builtin.object"

# Values does
df$datecol$values
#> ['2011-01-01T00:00:00.000000000' '2011-01-01T01:00:00.000000000']

class(df$datecol$values)
#> [1] "numpy.ndarray"         "python.builtin.object"
```

Upon changing to `values`, things convert smoothly. I also ran your tests and the same number of tests pass as did before.

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -157,7 +157,7 @@ py_to_r.datetime.date <- function(x) {
 #' @export
 py_to_r.pandas.core.series.Series <- function(x) {
   disable_conversion_scope(x)
-  py_to_r(x$as_matrix())
+  py_to_r(x$values)
 }
 
 py_object_shape <- function(object) unlist(as_r_value(object$shape))
@@ -233,7 +233,7 @@ py_to_r.pandas.core.frame.DataFrame <- function(x) {
   # extract numpy arrays associated with each column
   columns <- py_to_r(x$columns$values)
   converted <- lapply(columns, function(column) {
-    py_to_r(x[[column]]$as_matrix())
+    py_to_r(x[[column]]$values)
   })
   names(converted) <- columns
 


jjallaire:
@kevinushey will also take a look at this when he gets back.

Issue #242, opened 2018-04-27
Improved str() for python objects


mikaelpoul:
Currently, using `str()` on a arbitrary non-dict/list/tuple `python.builtin.object ` only prints the default rep, which is not very useful. Personally I use `str()` all the time to see the structure of R objects, both to see what the structure is like but also just to see whats inside. Would be nice if we could do the same with python objects.

PR #243, opened 2018-04-27
More useful str() for py objects

mikaelpoul:
This is just a suggestion for  #242

Calling str() on `python.builtin.object` now also lists the attributes which describe the python object under question, more similar to expected str() behavior.

For instance, running

```r
library(reticulate)

gensim <- import("gensim")

sentences <- gensim$models$word2vec$LineSentence("example.txt")
str(sentences)
```

now returns

image


R/python.R
```diff
@@ -868,10 +868,16 @@ py_str.default <- function(object, ...) {
...
   # get default rep
-  str <- py_str_impl(object)
+  id <- py_str_impl(object)
 
   # remove e.g. 'object at 0x10d084710'
-  str <- gsub(" object at 0x\\w{4,}", "", str)
+  id <- gsub(" object at 0x\\w{4,}", "", id)
+
+  # get list of accesible attr of the python object
+  attr <- capture.output(str(object$`__dict__`))
```
jjallaire: What happens if the object doesn't have a `__dict__`? (or are Python objects universally guaranteed to have a `__dict__` member?)

mikaelpoul: Yeah, that would probably throw an error. 

They are not universally guaranteed to have `__dict__`. User defined classes will have, though, I think. But there are probably exceptions. 

Ill try to come up with a safer approach.

jjallaire:
Just use py_has_attrib to check before accessing

R/python.R
```diff
@@ -865,19 +865,20 @@ py_str.default <- function(object, ...) {
...
+  str <- gsub(" object at 0x\\w{4,}", "", str)
 
-  # get list of accesible attr of the python object
-  attr <- capture.output(str(object$`__dict__`))
-
-  # combine
-  str <- paste0(c(id, attr), "\n")
+  # if available, use __dict__ to list accessible attr
+  if (py_has_attr(object, "__dict__")) {
+    attr <- capture.output(str(object$`__dict__`))
```
jjallaire: What happens if you call `py_str_impl` on the `__dict__` here instead (I think that's ultimately what will happen under the hood)

mikaelpoul: It throws an error, `Unable to access object (object is from previous session and is now invalid)`.

What happens under the hood right now is equivalent to,

```r
sentences %>%
  py_get_attr("__dict__") %>%
  py_to_r() %>%                         # returns a list
  str() %>%
  capture.output()
```

Could use `py_to_r(py_get_attr(object, "__dict__"))` instead of `object$'__dict__'`, to be explicit.

jjallaire:
@mikaelpoul we unfortunately had to roll back this PR b/c it caused at least a couple of downstream problems, see https://github.com/rstudio/reticulate/issues/250

Issue #244, opened 2018-04-30
Cannot use name 'pop' in Pandas DataFrame


ijlyttle:
This may be related to #157 

```r
library("reticulate")

py_run_string("import pandas as pd")
py_run_string("df_bad = pd.DataFrame({'pop': ['foo']})")
py_run_string("df_good = pd.DataFrame({'apop': ['foo']})")
```

So far so good, but when I try to bring this into R:

```r
py$df_bad
```

```
Error in py_get_attr_impl(x, name, silent) : 
  AttributeError: 'function' object has no attribute 'as_matrix'
```

However:

```r
py$df_good
```

```
  apop
1  foo
```

It seems like the name `'pop'` is somehow messing things up, and I don't think it should.

AliciaSchep:
In python, doing 

```py 
df_bad.pop.as_matrix()
``` 

gives an error.  This is because `pop` is a method of a DataFrame.  

The following in python works:

```py
df_bad[['pop']].as_matrix()
```
It seems like somewhere in reticulate's conversion df_bad.pop.as_matrix is getting called and if that could be avoided by calling df_bad[['pop']].as_matrix that would solve the issue... there are likely other DataFrame methods other than pop that will lead to this kind of unexpected error.

ijlyttle:
Nice catch, thanks!

jjallaire:
Assuming we can now close this?

ijlyttle:
Yes, thanks!

PR #245, opened 2018-05-01
address issue with pandas DataFrame names that clash with DataFrame methods

AliciaSchep:
This PR seeks to fix the issue raised in https://github.com/rstudio/reticulate/issues/244, that reticulate fails to convert pandas DataFrames where a column name matches a DataFrame method.  

The change is to use ``x$`__getitem__`(column) `` instead of `x[[column]]` within py_to_r.pandas.core.frame.DataFrame method.  This avoids `x[[column]]` being converted to `x.column`.  Accessing a column as an attribute fails when the column name clashes with a pandas method, as described in [pandas documentation](https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access)

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -233,7 +233,7 @@ py_to_r.pandas.core.frame.DataFrame <- function(x) {
   # extract numpy arrays associated with each column
   columns <- py_to_r(x$columns$values)
   converted <- lapply(columns, function(column) {
-    py_to_r(x[[column]]$as_matrix())
+    py_to_r(x$`__getitem__`(column)$as_matrix())
   })
   names(converted) <- columns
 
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks for the PR -- LGTM!

I wonder if it opens a deeper question about whether we should consider having different semantics for `$` vs. `[[` or `[`, since the `.` and `[` operators in Python mean different things. At least, we might consider implementing this for Pandas DataFrames.

AliciaSchep:
I had similar thoughts re `$` and `[` for pandas object, e.g. that this same end could maybe be achieved by defining `[[` for pandas DataFrame object as something different than `$` but wasn't sure whether that would have any other implications...

PR #254, opened 2018-05-09
add 

kevinushey:
This PR adds `py_get_item()`, `py_set_item()` and `py_del_item()` as explicit entrypoints for getting / setting Python items (e.g. in dictionaries).


R/python.R
```diff
@@ -817,6 +817,94 @@ py_list_attributes <- function(x) {
...
+#'
+#' @param x A Python object.
+#' @param name The item name.
+#' @param silent Boolean; when \code{TRUE}, attempts to access
+#'   missing items will return \code{NULL} rather than
+#'   throw an error.
+#'
+#' @family item-related APIs
+#' @export
+py_get_item <- function(x, name, silent = FALSE) {
```
jjallaire: Do these methods need to respect the `convert` attribute of `x`? (i.e. return Python objects rather than R objects when `convert = FALSE`)

kevinushey: Good question. Right now, as implemented, we return an R object when the `convert` flag is set, and a Python object otherwise. Do you think we should instead always return a Python object (as `py_get_attr()` does), and instead propagate the `convert` flag?

jjallaire:
No, we should do exactly what we are doing now (respect the `convert` flag)

kevinushey:
Okay, great -- then this PR should be ready to go!

jjallaire:
Let's just add a NEWS item

Issue #251, opened 2018-05-08
should '$' and '[' behave differently for pandas DataFrames?


kevinushey:
E.g. in Python:

`````
```{python}
import pandas as pd
pdf = pd.DataFrame({"pop": [1]})
print(pdf["pop"]) # accesses pop column
print(pdf.pop)    # accesses pop DataFrame method
```
`````

However, for Python objects exposed to R, `[` attempts to first access attributes rather than columns:

```
> library(reticulate)
> df <- data.frame(pop = 1)
> pdf <- r_to_py(df)
> pdf$pop
<bound method DataFrame.pop of    pop
0  1.0>
> pdf[["pop"]]
<bound method DataFrame.pop of    pop
0  1.0>
```

Should `[[` prefer accessing items rather than attributes for DataFrames?

I believe a similar question exists for Python dictionaries, and other objects implementing `__getitem__` in general.

**Comment by kevinushey** [comment]
After a chat with @jjallaire, we agree that we should try to migrate the semantics such that:

- `$` is analogous to Python's `.` operator; that is, it attempts to retrieve attributes (typically methods) on the object;
- `[[` and `[` are analogous to Python's `[` operator; that is, it is used for accessing items (`__getitem__`).

We'll plan to issue a warning if the use of the `$` operator ended up resolving an item rather than an attribute, just so existing user code has a path for migration.

**Comment by flying-sheep** [comment]
I think another possibility would be to make `$` and `[[` equivalent to `getattr` and `[` equivalent to `__getitem__`. That would have the advantage that it’s easy to get attributes with calculated names, e.g.

```r
for (attr_name in attrs) print(py_obj[[attr_name]])
```

however I think it would be quite confusing to let `[[` and `[` do completely different things, and it’s still possible to get aforementioned functionality via the more obscure

```r
for (attr_name in attrs) print(`$`(py_obj, attr_name))
```

dfalbel:
Related to https://github.com/rstudio/reticulate/pull/1431

t-kalinowski:
After taking a closer look, I see that `$` already prefers accessing attributes. 

The benefit of adding a `@` method would be that it would remove the potential for silent errors, where a call like `x@foo` would raise an attribute error, while `x$foo` would fall back silently to `getitem()`.  

``` r
  library(reticulate)
  py_run_string("import pandas as pd")
  pdf <- py_eval('pd.DataFrame({"pop": [1], "abc": [2]})', convert = F)
  pdf$pop
[...]

PR #255, opened 2018-05-10
define separate semantics for `$`, and `[[`

kevinushey:
Related: https://github.com/rstudio/reticulate/issues/251

This PR seeks to define different semantics for `$` and `[[`. With this PR:

- `$` is used for attribute access, as Python's `.` is used;
- `[[` is used for item access, as Python's `[[` is used.

For backwards compatibility, with Python (builtin) dictionaries, `$` still returns items rather than attributes. We may want to consider changing this in the future.


kevinushey:
For what it's worth, some of the surprise here might've arisen from how Pandas defines the `.` operator. For Pandas DataFrames, `.` will first look at attributes; if no such attribute exists, then it will attempt to look for a column. This implies that code of this form works:

```
>>> import pandas as pd
>>> pdf = pd.DataFrame({'apple': [1]})
>>> pdf.apple
0    1
Name: apple, dtype: int64
```

However, if an attribute (method) of the same name exists, then that is chosen instead:

```
>>> import pandas as pd
>>> pdf = pd.DataFrame({'pop': [1]})
>>> pdf.pop
<bound method DataFrame.pop of    pop
0    1>
```

kevinushey:
To help preserve backwards compatibility, I've decided on the following:

- `$` accesses attributes first, items second;
- `[` accesses items first, attributes second.

This should ensure the least amount of breakage in existing code, but it does imply a change in this case for example:

```
> d <- dict(items = 42)
> d$items
<built-in method items of dict>
> d['items']
42.0
```

Previously, `d$items` would have accessed the `items` item; now it accesses the attribute (bound method) called `items`.

R/python.R
```diff
@@ -231,41 +250,55 @@ py_has_convert <- function(x) {
...
-  else if (inherits(x, "python.builtin.dict"))
-    attrib <- py_dict_get_item(x, name)
-  else
-    attrib <- py_get_attr(x, name)
+  # get the attrib and convert as needed
+  if (prefer_attr) {
 
-  # convert
-  if (convert || py_is_callable(attrib)) {
+    if (py_has_attr(x, name)) {
```
kevinushey: These checks should only occur if the user is passing in a scalar string (length one character vector).

jjallaire:
@kevinushey Is this good to merge?

kevinushey:
I want to double-check this doesn't regress anything in keras or tensorflow so I'll report back after testing a bit there.

jjallaire:
Okay, you can go ahead and merge whenever you think it's ready.

kevinushey:
Tests pass with both `keras` and `tensorflow` with this branch, so I'm going to merge.

Issue #248, opened 2018-05-07
Issue with translating DataFrame: py_ref_to_r(x) : Unsupported array conversion from 19


adrivinca:
Hi reticulate team,

First off, we love your product, and we know this issue will not be terribly easy to diagnose - apologies. We are trying to use reticulate with some pretty complex python bindings of some Java code we have. We've run into an issue getting things back into R that googling hasn't been able to solve. We won't be able to provide a minimal example that is easy to reproduce (unless you feel like installing our package, which has some nontrivial dependencies), but we will do our best to explain it here. 

We have a `Scenario` object with a method `par()` which returns in this case a `pandas.DataFrame` with three columns: 

```
In[11]:  scenario.par('interestrate')

Out[11]: 
  unit  value  year
0    %   0.05  2010
1    %   0.05  2020
2    %   0.05  2030
3    %   0.05  2040

In[12]: type(scenario.par('interestrate'))
Out[12]: pandas.core.frame.DataFrame

In[13]: scenario.par('interestrate').dtypes
Out[13]: 
unit      object
value    float32
year       int32
dtype: object
```

We get the following error using reticulate:

```
> scenario$par("interestrate")
Error in py_ref_to_r(x) : Unsupported array conversion from 19
```

It seems like we're hitting a [case/switch in `python.cpp`](https://github.com/rstudio/reticulate/blob/master/src/python.cpp#L844) where it's trying to translate unicode, but finding no specific case. 

Oddly enough, if we dump this object to csv and read it back in, all works:

foo.py:
```
import pandas as pd

def test_read():
    df =  pd.read_csv('~/test.csv')
    print(df)
    print(df.dtypes)
    return df
```

And this works:
```
import('foo')
foo$test_read()
```

Could you provide any guidance of how we might address this issue? We're happy to do type conversions etc on the Python side.

Cheers

adrivinca:
cc: @gidden

gidden:
And here is the R stacktrace:

```
Error in py_ref_to_r(x): Unsupported array conversion from 19

Traceback:

 

1. scenario$par("interestrate")

2. py_to_r(result)

3. py_to_r.pandas.core.frame.DataFrame(result)

4. py_to_r(x$columns$values)

5. py_to_r.numpy.ndarray(x$columns$values)

6. NextMethod()

7. py_to_r.default(x$columns$values)

8. py_ref_to_r(x)
```

jjallaire:
[...]

PR #257, opened 2018-05-11
minimal changes to fix issue #248

gidden:
Here are the minimal changes I needed to fix my use case. The enums had to be moved up because of ordering issues (could also add functions to header file). 

Here string checking also checks numpy scalar strings. Then it is also checked when converting to bytes.

As I said, I have not tested this against other issues - perhaps you have CI? I'm happy to respond to any comments you may have.

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -136,7 +136,7 @@ std::string as_std_string(PyObject* str) {
 
   // conver to bytes if its unicode
   PyObjectPtr pStr;
-  if (PyUnicode_Check(str)) {
+  if (PyUnicode_Check(str) | isPyArrayScalar(str)) {
     str = PyUnicode_AsBytes(str);
     pStr.assign(str);
   }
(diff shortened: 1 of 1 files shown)


src/python.cpp
```diff
@@ -136,7 +136,7 @@ std::string as_std_string(PyObject* str) {
 
   // conver to bytes if its unicode
   PyObjectPtr pStr;
-  if (PyUnicode_Check(str)) {
+  if (PyUnicode_Check(str) | isPyArrayScalar(str)) {
```
kevinushey: A nitpick, but should we use `||` rather than `|` here?

kevinushey:
This PR looks good to me -- thanks for digging in! I'll let @jjallaire say the final word as he's more familiar with the guts here just in case this change might prompt any other changes we might need around this part of the code.

gidden:
I addressed your comment in the most recent commit @kevinushey. If there are some instructions on how to add tests, I can look into that as well.

src/python.cpp
```diff
@@ -136,7 +136,7 @@ std::string as_std_string(PyObject* str) {
 
   // conver to bytes if its unicode
   PyObjectPtr pStr;
-  if (PyUnicode_Check(str)) {
+  if (PyUnicode_Check(str) || isPyArrayScalar(str)) {
```
jjallaire: Presumably at this point we don't need an additional check that the PyArrayScalar is in fact a string scaler? (i.e. the callers of as_std_string have already established that the Python object is a string?)

jjallaire: I reviewed the various calls to `as_std_string()` and it does in fact look like we always know that we have a string.

Issue #198, opened 2018-03-27
passing dates or datetime.date objects as parameters


rafazaya:
Not sure if this is considered a good reprex example for reticulate, but let me try this.  

I'm having issues using datetime.date objects as parameters into my python scripts, where they become floats in the process  Per the below, a datetime.date object as a default parameter works, passing it in does not, and returning the value returns the integer equivalent.  Thoughts on how to pass a date to a function?

test.py:
```
from datetime import datetime, timedelta, date

def prior_day(edate=date.today()):
    return edate - timedelta(days=1)

def return_param(edate):
	return edate
```

Then in R...
``` r
library(reticulate)
#> Warning: package 'reticulate' was built under R version 3.4.4
conda_path <- "my/path"
env_name <- "envir"
use_condaenv(condaenv = env_name, conda = conda_path)

dt <- import("datetime")
edate <- dt$date(2018L,3L,26L)
edate
#> [1] "2018-03-26"
source_python("test.py")
prior_day()
#> [1] "2018-03-25"
prior_day(edate)
#> Error in py_call_impl(callable, dots$args, dots$keywords): TypeError: unsupported operand type(s) for -: 'float' and 'datetime.timedelta'
#> 
#> Detailed traceback: 
#>   File "<string>", line 4, in prior_day
return_param(edate)
#> [1] 17616
```

Created on 2018-03-26 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

kevinushey:
Thanks for the bug report! This is primarily because we transform R `Date` objects into NumPy `datetime64` arrays. See e.g.

```
> library(reticulate)
> r_to_py(Sys.Date())
['2018-03-26T00:00:00.000000000']
```

I believe this is typically the behaviour you want when manipulating e.g. Dates that live as part of a Pandas DataFrame, or as a vector / list of dates, but it does fail when one attempts to manipulate with plain old Python datetime objects.

I'm still not 100% sure whether we've made the right choice, though. Perhaps Date objects should be coerced into Python datetime objects? Maybe only times (e.g. `POSIXt`) should be converted into NumPy `datetime64` objects?

kevinushey:
For context, we make the conversion here:

https://github.com/rstudio/reticulate/blob/897c65dbf62fb570f86c9f6d54bc79a0fd98135a/R/conversion.R#L125-L146

Perhaps there's a better way we could decide whether we should produce a Python datetime vs. a NumPy datetime64?

[...]

PR #259, opened 2018-05-11
call 'r_to_py', 'py_to_r' recursively for lists

kevinushey:
This PR makes two changes:

1. Now, we always convert R Dates to Python datetime objects. If users want to use NumPy datetime64 objects, they should explicitly work with times / POSIXcts. This is recommended for efficiency, and having type stability makes it easier to program against `reticulate`. (see also: https://github.com/rstudio/reticulate/issues/198)

2. For R list objects, calling `r_to_py()` will call `r_to_py()` on the inner elements recursively. This is a bit slower, but is necessary to ensure that e.g. lists of Date objects are coerced properly.


terrytangyuan:
Would it better to do this recursively on Date objects? I am a little concerned that this will introduce overhead to downstream packages.

kevinushey:
I don't quite understand what you mean.

terrytangyuan:
Sorry I meant "would it better to do this recursively **only** on Date objects instead of doing this for lists items in general" .

kevinushey:
Vectors of date objects aren't an issue:

```
> library(reticulate)
> dates <- rep(Sys.Date(), 3)
> r_to_py(dates)
['2018-05-11T00:00:00.000000000' '2018-05-11T00:00:00.000000000'
 '2018-05-11T00:00:00.000000000']
```

It's just for `list` objects where our only choice is to recursively visit each list element to figure out its type, and then do something sensible after learning about all the types. Or that's probably what you meant?

Recursively converting in this way would mean we could support conversions of things like this, though:

```
> r_to_py(list(1L, 2, "3", Sys.Date(), Sys.time()))
[1, 2.0, '3', datetime.date(2018, 5, 11), array(['2018-05-11T21:48:46.328669184'], dtype='datetime64[ns]')]
```

kevinushey:
It also seems to me that if one uses Python lists + Python datetime objects, they're likely not placing efficiency as the top priority -- if someone absolutely needs efficient operations, then they're likely to stick with NumPy arrays (and so `datetime64`).

It's really tough to balance ergonomics + performance in this space :-/

terrytangyuan:
Ok that makes sense to me now. Thanks for the clarification!

kevinushey:
@jjallaire it's been a while since we submitted this PR but I think it's still worth taking. What do you think?

jjallaire:
Will merge now. Could you commit a NEWS change to master and bump the development version?

Issue #270, opened 2018-05-18
py_load_object can't read DataFrames with duplicates in index


holken1:
As `py_load_object` tries to insert the Pandas DataFrame's index as `row.names` in the R `data.frame`, it will fail whenever the index contains duplicate values (which is allowed in a Pandas DataFrame).

Here's a sample Python code to create a pickle file with a duplicate value in the index:
```
d = {'one' : [1., 2., 3., 4.],
     'two' : [4., 3., 2., 1.]}
df = pd.DataFrame(d, index=['a', 'a', 'c', 'd'])
pickle.dump(df, open("test.pkl", "wb"))
```
And then this R code will fail:
```
library(reticulate)
pandas <- import("pandas")
df <- py_load_object("test.pkl", pickle = "pickle")
```
The exact error message:
```
Error in `row.names<-.data.frame`(`*tmp*`, value = value) : 
  duplicate 'row.names' are not allowed
In addition: Warning message:
non-unique value when setting 'row.names': ‘a’  
```

IMHO, ideally there would be an option like `use.row.names=TRUE/FALSE` that would allow the user to switch between using row.names and a column, either to work around this issue, or simply depending on what is preferred by the user.

PR #271, opened 2018-05-18
handle duplicates in Pandas index

kevinushey:
This PR makes two changes:

- `py_to_r()` now tags the returned R data.frame with the Pandas index, as an attribute called `pandas.index`. This is for users who need to post-process the index in some way, because our default converter didn't do the right thing.

- We detect duplicate values in the index; when discovered, we don't attempt to set the index and instead emit a warning.

Fixes https://github.com/rstudio/reticulate/issues/270.

@jjallaire let me know if this feels like the right change.


jjallaire:
Maybe add a NEWS item for this?

PR #279, opened 2018-05-27
check if python symbols are present in the current process

randy3k:
When R is embedded in a Python environment, it is possible that the python symbols get loaded into the main process. This PR first checks if the current process contains the python symbols. If found, it uses the python symbols found instead of the given `libPath`.

It would be more robust to check if the library found in the main process is consistent with `libPath`, and throw an error if they do not match. But it will require some work and I currently don't have an elegant way to implement the check.

related to #98


src/libpython.cpp
```diff
@@ -53,14 +53,32 @@ void lastDLErrorMessage(std::string* pError)
...
 #else
-  *ppLib = ::dlopen(libPath.c_str(), RTLD_NOW|RTLD_GLOBAL);
+  int symbols_present = 0;
+  void* pLib = NULL;
+  // check if the current process contains python symbols
+  pLib = ::dlopen(NULL, RTLD_NOW|RTLD_GLOBAL);
+  loadSymbol(pLib, "Py_IsInitialized", (void**) &Py_IsInitialized, pError);
+  ::dlclose(pLib);
+
+  if (Py_IsInitialized != NULL) {
```
jjallaire: Is there a way to do this for Windows as well?

jjallaire:
This looks good to me at first reading. I am going to wait to merge it until after the next CRAN release (which we are just now locking down for)

src/libpython.cpp
randy3k: we might use [this](https://github.com/JuliaPy/PyCall.jl/blob/eec4a0e810b9f87234569d742f6664ce16ee134d/src/startup.jl#L21-L36) from PyCall of Julia. I don't have a proper windows machine to test it though. 

This PR is needed for me to use the latest conda python (an older version of conda python doesn't load symbols into the main process). I am not sure if python symbols will be loaded into the main process in Windows as well.

randy3k:
I have updated the code. Now it checks the python symbols in `py_discover_config`. It returns the following in [rtichoke](https://github.com/randy3k/rtichoke) for instance.
```
r$> reticulate::py_discover_config()
python:         /Users/Randy/miniconda3/bin/python
libpython:      /Users/Randy/miniconda3/bin/python
pythonhome:     /Users/Randy/miniconda3:/Users/Randy/miniconda3
version:        3.6.5 |Anaconda, Inc.| (default, Mar 29 2018, 13:14:23)  [GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
numpy:           [NOT FOUND]

NOTE: Python version was forced by main process

```

jjallaire:
@randy3k Thanks for all of this work. I am going to wait to merge until after the next CRAN release.

randy3k:
bump

randy3k:
I can see that a new release was made. Maybe it's the time to review this PR?

jjallaire:
Yes, apologize for not reviewing as part of v1.10 but there is a lot of code to review and it touches some sensitive parts of initialization so there is definitely a risk of breakage. I will try to review for the v1.11 release (which should be in another ~ 30 days).

docbiz:
(Bumping as it's still an issue that is tripping people up.)

randy3k:
This is no longer needed for radian as we are patching reticulate directly https://github.com/randy3k/radian/commit/88e0943ef2aa6820d2ed5c85227ca297ef04324f.

PR #282, opened 2018-05-29
fix conversion of boolean matrices (closes #280)

### LINKED ISSUE #280: Type conversion for boolean matrix
Hi, 

When passing a boolean matrix from R to python, something unexpected seems to happen. Suppose I have a simple python module (in a file "myfun.py")
```
def myfun(x):
    print(x)
```
Now when importing this module 
```
py_module <- reticulate::import_from_path("myfun", path = ".")

a <- c(TRUE, TRUE, TRUE)
A <- cbind(a, a)
B <- 1 * A
```
and applying this function to R objects `a, A, B`
```
py_module$myfun(a)

[True, True, True]
```

```
py_module$myfun(A)

[[ True False]
 [False  True]
 [False False]]
```

```
py_module$myfun(B)

[[1. 1.]
 [1. 1.]
 [1. 1.]]
```
Conversion for `a` and `B` happens as expected, whereas for `A` something strange happens. 

### SessionInfo()
```
R version 3.5.0 (2018-04-23)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Fedora 26 (Twenty Six)

Matrix products: default
BLAS: /usr/local/lib64/R-3.5/lib/libRblas.so
LAPACK: /usr/local/lib64/R-3.5/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8       
 [4] LC_COLLATE=en_GB.UTF-8     LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8   
 [7] LC_PAPER=en_GB.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

 [1] Rcpp_0.12.17     rstudioapi_0.7   magrittr_1.5     knitr_1.20       whisker_0.3-2   
 [6] devtools_1.13.5  debugme_1.1.0    lattice_0.20-35  R6_2.2.2         stringr_1.3.1   
[11] tools_3.5.0      grid_3.5.0       withr_2.1.2      htmltools_0.3.6  yaml_2.1.19     
[16] digest_0.6.15    assertthat_0.2.0 rprojroot_1.3-2  crayon_1.3.4     processx_3.1.0  
[21] Matrix_1.2-14    callr_2.0.4      formatR_1.5      memoise_1.1.0    evaluate_0.10.1 
[26] rmarkdown_1.9    reprex_0.1.2     stringi_1.2.2    compiler_3.5.0   backports_1.1.2 
[31] reticulate_1.7   jsonlite_1.5    

python2.7
```

kevinushey:
This PR fixes #280 by converting the (`int`-flavored) R logical vector into a plain array of `bool`s before creating the NumPy array.

We set the `NPY_ARRAY_OWNDATA` flag to tell NumPy that it should own + manage the data passed in to the constructor when doing this, so that Python / NumPy itself can free the malloc'ed data when the associated Python object goes out of scope.

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -255,7 +255,7 @@ int narrow_array_typenum(PyArray_Descr* descr) {
 bool is_numpy_str(PyObject* x) {
   if (!isPyArrayScalar(x))
     return false; // ndarray or other, not string
-  
+
   PyArray_DescrPtr descrPtr(PyArray_DescrFromScalar(x));
   int typenum = narrow_array_typenum(descrPtr);
   return (typenum == NPY_STRING || typenum == NPY_UNICODE);
@@ -274,7 +274,7 @@ bool is_python_str(PyObject* x) {
 
   else if (is_numpy_str(x))
     return true;
-  
+
   else
     return false;
 }
@@ -995,6 +995,21 @@ PyObject* r_to_py(RObject x, bool convert) {
            "converted");
     }
 
+    int flags = NPY_ARRAY_FARRAY_RO;
+
+    // because R logical vectors are just ints under the
+    // hood, we need to explicitly construct a boolean
+    // vector for our Python array. note that the created
+    // array will own the data so we do not free it after
+    if (typenum == NPY_BOOL) {
+      R_xlen_t n = XLENGTH(sexp);
+      bool* converted = (bool*) std::malloc(n * sizeof(bool));
+      for (R_xlen_t i = 0; i < n; i++)
+        converted[i] = LOGICAL(sexp)[i];
+      data = converted;
+      flags |= NPY_ARRAY_OWNDATA;
+    }
+
     // create the matrix
     PyObject* array = PyArray_New(&PyArray_Type,
                                    nd,
@@ -1003,7 +1018,7 @@ PyObject* r_to_py(RObject x, bool convert) {
                                    NULL,
                                    data,
                                    0,
-                                   NPY_ARRAY_FARRAY_RO,
+                                   flags,
                                    NULL);
 
     // check for error
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -1003,7 +1003,7 @@ PyObject* r_to_py(RObject x, bool convert) {
     // array will own the data so we do not free it after
     if (typenum == NPY_BOOL) {
       R_xlen_t n = XLENGTH(sexp);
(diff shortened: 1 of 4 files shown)


src/python.cpp
```diff
@@ -995,6 +995,21 @@ PyObject* r_to_py(RObject x, bool convert) {
...
 
+    int flags = NPY_ARRAY_FARRAY_RO;
+
+    // because R logical vectors are just ints under the
+    // hood, we need to explicitly construct a boolean
+    // vector for our Python array. note that the created
+    // array will own the data so we do not free it after
+    if (typenum == NPY_BOOL) {
+      R_xlen_t n = XLENGTH(sexp);
+      bool* converted = (bool*) std::malloc(n * sizeof(bool));
```
jjallaire: Is there a numpy memory allocation function or macro? (just to be 100% sure that the same heap is used for malloc and free)

kevinushey: You're right; it looks like there's a `PyArray_malloc()` function (https://docs.scipy.org/doc/numpy-1.7.0/reference/c-api.array.html#memory-management). I'll update the PR.

kevinushey: Reading a bit here:

https://github.com/numpy/numpy/blob/f5758d6fe15c2b506290bfc5379a10027617b331/numpy/core/include/numpy/ndarraytypes.h#L348-L367

It looks like NumPy explicitly uses the 'raw' allocation APIs in Python. That is, while it is possible to register custom allocators in Python (appears to be new in Python 3.6), NumPy explicitly requests the 'raw' allocators when possible. Note that these effectively just delegate to `malloc()`:

https://github.com/python/cpython/blob/28f713601d3ec80820e842dcb25a234093f1ff18/Objects/obmalloc.c#L66-L76

I think this implies the code here should be good 'as-is' since all the indirection in NumPy is around making sure the raw `malloc()` routine is used.

jjallaire:
Okay, should we still define a PyArray_malloc macro (with a link to the
those source code locations) to make it explicit that we are attempting to
synchronize with the NumPy heap?

kevinushey:
Done!

Issue #298, opened 2018-06-26
With Shiny dataframe gets converted to dictionary in Python


ggsdc:
Hello,

I'm testing this library to be used in a Shiny application, right now I'm having a isse trying to pass a data.frame from R to Python.

The main idea is the following:

1. Generate a random data.frame with an `numericInput` in R.
2. Choose a multiplier with another `numeriInput`.
3. Have a python function that multiplies all the columns in the data.frame in Python.

This is not the final aim of the application, just a small example to show the error I am encountering.

This would be the code so you can try and replicate the error.

global.R:
```
library(reticulate)
library(shiny)
library(shinydashboard)
use_virtualenv("./env")
source_python("main.py")
```

The Python enviroment is just a clean virtual enviroment in the same folder as the RStudio project and the .R and .py files. It just has pandas installed, its requirements.txt file would be:

```
numpy==1.14.3
pandas==0.23.0
python-dateutil==2.7.3
pytz==2018.4
six==1.11.0
```
The rest of the scripts are:

ui.R:
```
header <- dashboardHeader(title="Example")

sidebar <- dashboardSidebar(
  sidebarMenu(
    menuItem("Main Page", tabName="main", icon=icon("info"))
  )
)

body <- dashboardBody(
  tabItems(
    tabItem(
      tabName="main",
      column(numericInput(label=h3("Number of rows"), "rows", value=10),
             br(),
             tableOutput("rows"),
             width=3),
      column(numericInput(label=h3("Multiplier"), "mult", value=1),
             br(),
             tableOutput("dataframe"),
             width=3)
    )
  )
)

dashboardPage(header = header, sidebar = sidebar, body = body)
```

server.R:
```
shinyServer(function(input, output, session){
  
  dfR <- reactive({
    x <-data.frame(a=runif(input$rows,0,100), b=runif(input$rows,0,100), c=runif(input$rows,0,100))
    return(x)
    })
  
  output$rows <- renderTable(dfR())
  
  output$dataframe <- renderTable(mult_df(dfR(), input$mult))
  
})
```

main.py:
```
import pandas as pd

def mult_df(dfA, value):
  import pandas as pd
  dfB = pd.DataFrame.from_dict(dfA) * value
  return dfB

```
With these files it works but as you can see in the python function I have to use `pd.DataFrame.from_dict(dfA)` to convert the input data.frame as it gets interpreted as a dictionary.

If I change the main.py to this:
```
import pandas as pd

def mult_df(dfA, value):
[...]

PR #299, opened 2018-06-27
Always call r_to_py S3 method when converting objects from Python to R

jjallaire:
Addresses https://github.com/rstudio/reticulate/issues/298 

@kevinushey Could you review this please?

I'm wondering if the `static` references I introduced to the R reticulate environment and `r_to_py()` functions could cause any problems? (e.g. being feed at the end of a process lifetime when R is already gone?).

The other issue is ensuring that we have all of the memory semantics right (i.e. we won't either leak objects or leave an object without a reference causing a crash). This is about the lowest level change we could make in reticulate so has the potential to wreak havoc if it's not done correctly!

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -935,9 +935,30 @@ SEXP py_to_r(PyObject* x, bool convert) {
   }
 }
 
+PyObject* r_to_py(RObject x, bool convert) {
+
+  // get a static reference to the R version of r_to_py
+  static Rcpp::Environment pkgEnv = Rcpp::Environment::namespace_env("reticulate");
+  static Rcpp::Function r_to_py_fn = pkgEnv["r_to_py"];
+
+  // call the R version and hold the return value in a PyObjectRef (SEXP wrapper)
+  // this object will be released when the function returns
+  PyObjectRef ref(r_to_py_fn(x, convert));
+
+  // get the underlying Python object and call Py_IncRef before returning it
+  // this allows this function to provide the same memory sematnics as the
+  // previous C++ version of r_to_py (which is now r_to_py_cpp), which always
+  // calls Py_IncRef on Python objects before returning them
+  PyObject* obj = ref.get();
+  Py_IncRef(obj);
+
+  // return the Python object
+  return obj;
+}
+
 // convert an R object to a python object (the returned object
 // will have an active reference count on it)
-PyObject* r_to_py(RObject x, bool convert) {
+PyObject* r_to_py_cpp(RObject x, bool convert) {
 
   int type = x.sexp_type();
   SEXP sexp = x.get__();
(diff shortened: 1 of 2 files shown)


kevinushey:
Given that `reticulate` has already picked up some reverse dependencies I could try running a revdepcheck to see if this PR breaks anything downstream.

I think that having static references to these Rcpp objects may be dangerous -- the destructor will attempt to unprotect the associated SEXP, which may no longer exist at shutdown time. I think leaving them as non-static would be okay.

jjallaire:
Okay, I got rid of the static.

I ran the tests against keras and tensorflow and both had no issues.

I think it would be good to also do some diagnostics along these lines:

1) Run tests with GC_TORTURE
2) Run tests with whatever CRAN uses for memory error checking

jjallaire:
Beyond that. I believe that I've gotten the memory semantics right here but it's definitely worth another pair of eyes tracking back the implications of things.

Would also be good to just empirically confirm that we aren't leaking any memory with the new implementation.

kevinushey:
If there were a memory leak, I would expect code like this to cause the R session's memory usage to blow up:

```r
library(reticulate)
df <- data.frame(alpha = 1:1E3)
py_run_string('def identity(x): return x')
main <- import_main()
repeat main$identity(df)
```

All seems well in this case.

jjallaire:
Okay, I'm running right now with `gctorture()` set to the most stringent level to check that angle.

src/python.cpp
```diff
@@ -935,9 +935,30 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
 
+PyObject* r_to_py(RObject x, bool convert) {
+
+  // get a static reference to the R version of r_to_py
+  Rcpp::Environment pkgEnv = Rcpp::Environment::namespace_env("reticulate");
+  Rcpp::Function r_to_py_fn = pkgEnv["r_to_py"];
+
+  // call the R version and hold the return value in a PyObjectRef (SEXP wrapper)
+  // this object will be released when the function returns
+  PyObjectRef ref(r_to_py_fn(x, convert));
```
kevinushey: Just to make sure I understand the semantics, `PyObjectRef` itself is an R environment containing the associated Python wrapper object. A C finalizer is set on the Python object when this object is constructed, and that C finalizer will call `Py_DecRef()` when the parent R environment is cleaned up. To ensure that the object then lives beyond that cleanup, we need to explicitly increase the reference count (as is done below).

jjallaire: Yes, that's exactly correct (note that py_to_r_cpp effectively does the same thing, incrementing the reference count before returning any Python object).

R CMD check with `gctorture()` is still running, no problems yet....

kevinushey:
LGTM! I think we're ready to merge.

ijlyttle:
Just a wee comment to say thanks for this: things get a little easier in [altair](https://github.com/vegawidget/altair).

We can now remove the `r_to_py()` call in the chart constructor:

```R
alt$Chart(r_to_py(vega_data$cars()))
```

I was getting ready to file an issue, but no need! 

cc @AliciaSchep

Issue #303, opened 2018-07-02
Converting R raw vector to python


jeffwong-nflx:
I am trying to serialize some complex R objects to python.  I want to serialize the R object, and pass the raw bytes over. I've found some odd behavior when using RawVectors with reticulate.

In R I can generate raw using

`writeBin('hello', raw())`

Python can understand raw, for example

`"68656c6c6f00".decode("hex")`

This correctly decodes the raw to 'hello'. I can seamlessly pass these raw bytes around, for example

`writeBin('hello', raw()) %>% r_to_py %>% py_to_r

For more complex R objects I want to use `serialize`, because writeBin can only operate on vectors. For example

`serialize(list("hello", 1:3), NULL) %>% r_to_py() %>% py_to_r`

However, the python side picks this up as just "X\n". What could be going on here?

kevinushey:
My suspicion is that the null bytes here are causing issues. For example:

```r
library(reticulate)

roundtrip <- function(x) {
  py_to_r(r_to_py(x))
}

roundtrip(as.raw(c(48:57)))
#> [1] "0123456789"
roundtrip(as.raw(c(48:57, 0, 48:57)))
#> 0123456789
```

IIUC, the underlying Python string object _does_ contain the whole set of characters (including the null byte); however, printing truncates to the first discovered null byte. Python seems to handle this properly so this may be a bug in reticulate (?)

```python
>>> bytes("123\000456")
'123\x00456'
```

Also worth pointing out that calling `py_to_r()` on a string containing null bytes silently fails to convert. I suspect this is because we're converting the raw R vector to a Python string, and then when reticulate wants to convert the Python string back, it chooses to convert as an R string. Because R strings cannot contain null bytes, reticulate bails and gives the Python object as-is.

@jjallaire what are your thoughts? We might consider converting R raw vectors to Python [bytearray](https://docs.python.org/3.1/library/functions.html#bytearray) objects by default. One of the difficulties here is that with Python 2 strings and bytes are basically the same thing; in Python 3 they're separate types but it would be nice to choose a conversion scheme that's consistent across versions if possible.

jjallaire:
It looks to me like we do the conversion right here, and for Python 3 it's PyBytes and for Python 2 it's PyString (https://github.com/rstudio/reticulate/blob/master/src/python.cpp#L157-L162)

```r
PyObject* as_python_bytes(Rbyte* bytes, size_t len) {
  if (is_python3())
[...]

PR #305, opened 2018-07-03
Convert RAW vectors to Python bytearray; Convert Python bytearray to RAW

jjallaire:
Addresses https://github.com/rstudio/reticulate/issues/303

@kevinushey Could you review?

Trying to think through whether we should merge this now (CRAN release process will start in ~ 7 days) or wait. The only changes in behavior are:

- RAW is now always converted to bytearray (previously it was string for Python 2 which is plain wrong and bytes for Python 3). 

- bytearray is now converted to RAW (previously it had no conversion)

I considered whether we should convert bytes to RAW, however I think we can't do this for Python 2 (where bytes is the same thing as string) and I don't think we should introduce behavior that differs across Python 2 and Python 3.


jeffwong-nflx:
Thanks JJ! This is really helpful for me. I'm working on a use case where I serialize bytes from R to storage. There is a python client, motivating the `serialize() %>% r_to_py` call, and to retrieve the bytes from storage the opposite `py_to_r %>% unserialize` call

R/python.R
```diff
@@ -981,6 +981,15 @@ py_str.python.builtin.object <- function(object, ...) {
   str
 }
 
+#' @export
+py_str.python.builtin.bytearray <- function(object, ...) {
+  builtins <- import_builtins()
+  paste(
+    paste0("python.builtin.bytearray (length ", builtins$len(object), "):"),
+    capture.output(str(py_to_r(ba), give.length = FALSE, give.head = FALSE, vec.len = 10))
```
kevinushey: Looks like that `ba` should be `object`?

src/python.cpp
```diff
@@ -1167,7 +1167,8 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
   // bytes
   } else if (type == RAWSXP) {
 
-    return as_python_bytes(RAW(sexp), Rf_length(sexp));
+    Rcpp::RawVector raw(sexp);
+    return PyByteArray_FromStringAndSize((const char*)&raw[0], raw.size());
```
kevinushey: This is pedantic but `&raw[0]` could technically be an out-of-bounds access for an empty vector.

R/python.R
jjallaire: Good catch! I actually just changed this to just print the number of bytes and the type (as the initial implementation actually converted the bytearray to RAW which is probably too expensive for a print method)

jeffwong-nflx:
In this PR could we also address another issue?

[When reticulate converts RAW to bytes](https://github.com/rstudio/reticulate/blob/91ba73dd457ce111cb58631893d7b148e1a3b328/src/python.cpp#L1170), it measures the size of RAW using Rf_length(). This is limited to sizes of 2GB. An alternative that allows longs is Rf_xlength. See this issue from [Rcpp](https://github.com/RcppCore/Rcpp/issues/302)

Without this, a call to serialize %>% r_to_py will throw an error if the bytes are too large

src/python.cpp
jjallaire: Not pedantic at all as that's a real failure mode in both directions! (thanks for catching that). Just pushed a change, could you take a quick look (wrote tests but another pair of eyes always useful)

jeffwong-nflx:
I checked out your branch and everything worked well for my use case. Thanks!

jjallaire:
@jeffwong-nflx Thanks for verifying!

@kevinushey You can merge this after final review.

PR #309, opened 2018-07-11
Make variables defined using %as% operator in with() available after execution of the with block

jjallaire:
Same behavior as Python. See https://stackoverflow.com/questions/6432355/variable-defined-with-with-statement-available-outside-of-with-block

@kevinushey Could you review this?

cc @skeydan re: https://github.com/rstudio/tfsandbox/issues/4


kevinushey:
LGTM!

PR #312, opened 2018-07-12
Clear error if unable to find __module__ name.

akshaym:


jjallaire:
Thanks, made an alternate fix here: https://github.com/rstudio/reticulate/commit/b1728da7765b042dde04e07886595f5630286048#diff-849f590f08cba3094f269e0a4d69398d

Issue #313, opened 2018-07-13
Python exception IOError : Bad file descriptor under Windows


Aygos:
Hi,

With latest 1.9 CRAN or 1.9.0.9001 Github version of reticulate, every time repl_python() is used (manually via R Console, or automatically via RStudio (run selection)), reticulate package throw the following Python exception : 

```
> reticulate::repl_python()
Python 2.7.14 (E:\Mes_Documents\Programs\RStudio-1.2.792\bin\rsession.exe) Reticulate 1.9.0.9001 REPL -- A Python interpreter in R.
>>> print 'ok ?'

Error in py_call_impl(callable, dots$args, dots$keywords) :
   IOError: [Errno 9] Bad file descriptor
```
Normally Bad file descriptor is thrown when trying to read / write on an manually closed file object.

Under Linux, is ok, there is no error.

``

Small bug, but rather annoying.

Thx, in advance.

Cyril.

---

Here is my environment :

```
> sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

Matrix products: default

locale:
[1] LC_COLLATE=French_France.1252  LC_CTYPE=French_France.1252    LC_MONETARY=French_France.1252
[4] LC_NUMERIC=C                   LC_TIME=French_France.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] reticulate_1.9.0.9001

loaded via a namespace (and not attached):
[1] compiler_3.5.1  Matrix_1.2-14   tools_3.5.1     Rcpp_0.12.17    grid_3.5.1      jsonlite_1.5   
[7] lattice_0.20-35
```

And  :

```
> py_config()
python:         E:\MES_DO~1\Programs\Python27\python.exe
libpython:      E:/MES_DO~1/Programs/Python27/python27.dll
pythonhome:     E:\Mes_Documents\Programs\Python27
version:        2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
Architecture:   64bit
numpy:          E:\Mes_Documents\Programs\Python27\lib\site-packages\numpy
numpy_version:  1.14.3
```

Note : perhaps this is related to this previously fixed bug : 
[...]

PR #314, opened 2018-07-13
always remap output streams with Windows + Python 2.7

kevinushey:
Fixes https://github.com/rstudio/reticulate/issues/313.

@jjallaire can you review? I wonder if the intention was that this code would be firing more often (it seems like Python is always initialized with `stdout` and `stderr` but we might want to remap to R in those cases as well)

### DIFF
--- a/R/output.R
+++ b/R/output.R
@@ -5,7 +5,8 @@ remap_output_streams <- function() {
   output$remap_output_streams(
     write_stdout,
     write_stderr,
-    tty = interactive() || isatty(stdout())
+    tty = interactive() || isatty(stdout()),
+    force = is_windows() && !is_python3()
   )
 }
 
--- a/inst/python/rpytools/output.py
+++ b/inst/python/rpytools/output.py
@@ -54,10 +54,10 @@ class OutputRemap(object):
     return None
 
 
-def remap_output_streams(r_stdout, r_stderr, tty):
-  if (sys.stdout is None):
+def remap_output_streams(r_stdout, r_stderr, tty, force):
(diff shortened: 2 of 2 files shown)


randy3k:
I just realized that this PR breaks the usage of embedding `R` with Windows + Python 2.7. It creates a dead loop since `remap_output_streams` maps `sys.stdout` to R steam and the embedding R remaps the R stream to `sys.stdout`.

kevinushey:
What do you mean by "usage of embedding R"?

randy3k:
Sorry for not being clear. I meant using [`rpy2`][1] or [`rapi`][2] to embed R within an python environment.

[1]: https://rpy2.readthedocs.io/
[2]: https://github.com/randy3k/rapi
[3]: https://github.com/randy3k/rtichoke

This PR breaks any such usage. For instance, [`rtichoke`][3] no longer could load `reticulate` on Window + Python 2.7.

jjallaire:
As far I know we absolutely require remapping to make many types of Python output show up (i.e. reticulate itself is broken without remapping). Perhaps rtichoke could detect this state and then do something different.

randy3k:
I guess it depends who is being embedded. rtichoke is working perfectly with Windows + python 2.7. It means it’s got the stdout is not none and print is functioning. 

An alternative solution is to detect if python has previously initialized when reticulate is loaded and stop routing the streams if R is being embedded in python.

randy3k:
> `RETICULATE_REMAP_OUTPUT_STREAMS`

👍

PR #322, opened 2018-07-25
Expose find_conda() and get_virtualenv()

javierluraschi:
For users interested in having access to the python executables to interface over command line, exposing the path to the virtualenv and searching for conda installations would certainly be helpful.


R/conda.R
```diff
@@ -194,7 +194,8 @@ conda_version <- function(conda = "auto") {
 }
 
 
-
+#' @rdname conda-tools
+#' @export
 find_conda <- function() {
```
jjallaire: This function finds the global conda rather than the currently in use conda environment so I don't think it's what you want.

jjallaire:
Let's discuss the actual use case for this PR before we decide whether to merge (there may be a better way)

jjallaire:
This will get you the path to the python binary for the e.g. the "r-mlflow" conda environment:

```r
subset(conda_list(), name == "r-mlflow")$python
```

javierluraschi:
Sounds good. I think you might still want to consider taking the `get_virtualenv()` change to allow others using only virtualenvs to extract this, but since we will scope to conda only support in mlflow, I'll close this one. Thanks!

PR #328, opened 2018-08-06
don't drop blank lines in reticulate engine

kevinushey:
Fixes an issue noted by https://github.com/rstudio/rstudio/issues/1556#issuecomment-410789560.

Putting this up as a PR as I do not recall the original motivation for stripping these blank lines, but it seems it was done very early on (https://github.com/rstudio/reticulate/commit/93886117f4f414cb1b88d7932071d9e3b4ddbae7).

I don't seem to notice any ill effects from including the blank lines (and they are then properly reflected in the resulting rendered document) so I think this should be safe.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -60,7 +60,7 @@ eng_python <- function(options) {
   # helper function for extracting range of code, dropping blank lines
   extract <- function(code, range) {
     snippet <- code[range[1]:range[2]]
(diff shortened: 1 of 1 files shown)


jjallaire:
I don't have any insight either on why the blank lines are excluded.

gcalmettes:
Any specific tests needed before being able to merge this PR?

PR #335, opened 2018-08-13
allow configure of output stream remap

kevinushey:
cc: @randy3k; can you test this PR and let us know if it's sufficient for what you need?


randy3k:
Thanks for the quick work. It fixes my issue.

PR #344, opened 2018-08-22
ensure modules in pwd can be imported

### LINKED ISSUE #340: Import local module fails in REPL (repl_python())
Simple reproducible example:

Under C directory I have "utils" directory with `__init__.py` file and `utils.py` file:

C\:
*utils
**`__init__.py`
**`utils.py`

In Python this should be enough for "utils" to be a "package", and for `from utils import *` to work in REPL, say IDLE:

image

This doesn't work in `reticulate::repl_python()`:

image

Am I doing something wrong, is it a bug or not yet a feature?

Rstudio v1.2.830

```
> sessionInfo()
R version 3.5.1 (2018-07-02)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=English_Israel.1252  LC_CTYPE=English_Israel.1252    LC_MONETARY=English_Israel.1252
[4] LC_NUMERIC=C                    LC_TIME=English_Israel.1252    

attached base packages:

other attached packages:
[1] reticulate_1.10

[1] compiler_3.5.1  Matrix_1.2-14   tools_3.5.1     Rcpp_0.12.18    grid_3.5.1      jsonlite_1.5   
[7] png_0.1-7       lattice_0.20-35
```

kevinushey:
Closes #340.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -44,11 +44,6 @@ ensure_python_initialized <- function(required_module = NULL) {
      }
     .globals$py_config <- initialize_python(required_module, use_environment)
 
-    # ensure modules can be imported from the current working directory
-    sys <- import("sys", convert = FALSE)
-    if (sys$path$count("") == 0L)
-      sys$path$insert(0L, "")
-
     # generate 'R' helper object
     py_inject_r(envir = globalenv())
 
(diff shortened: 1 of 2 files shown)


kevinushey:
Yes, I think it's worth just making this happen by default (since this is what usually happens when you launch a Python session from the terminal, I think it's the behavior users will expect universally)

jjallaire:
We do other manipulation of the Python sys path within the `initialize_python()` function here: https://github.com/rstudio/reticulate/blob/a4dcf63c8c0e7695524b06eea1b15fd9ce8f4e95/R/package.R#L132-L135

Should we have this code live in the same place?

kevinushey:
Ah, I didn't realize that. I'll move the code there!

PR #359, opened 2018-09-28
ensure code is auto-printed in Python chunks

kevinushey:
This PR fixes an issue where running a chunk e.g.

`````
```{python}
1 + 1
```
`````

would not produce any output when run in RStudio. Note that the execution model here is effectively the same as what we do in `repl_python()`; ie it uses the same tricks to ensure that Python auto-prints the evaluated expressions.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -119,23 +119,11 @@ eng_python <- function(options) {
   # synchronize state R -> Python
   eng_python_synchronize_before()
 
-  # determine if we should capture errors (don't capture in rmarkdown::render())
-  capture_errors <- local({
-
-    if (identical(options$error, TRUE))
-      return(TRUE)
-
-    if (!requireNamespace("rmarkdown", quietly = TRUE))
-      return(TRUE)
-
-    rmarkdown <- asNamespace("rmarkdown")
-    context <- rmarkdown$.render_context
-    if (is.null(context))
-      return(TRUE)
-
-    is.null(context$peek())
-
-  })
+  # determine if we should capture errors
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by kevinushey** [comment]
Let's not merge or review just yet (need to test + verify a little bit more)

**Comment by jjallaire** [comment]
Okay, sounds good. Let me know when it's a good time to review.

**Comment by kevinushey** [comment]
Okay, this PR should be ready for review.

One bit I added that's a bit hacky is a check whether the `eng_python()` function is being called within `rmarkdown::render()` or not. Do we have a publicly documented way of accomplishing this?

File: R/knitr-engine.R [hunk 1]
```diff
@@ -118,34 +110,45 @@ eng_python <- function(options) {
...
+  capture_errors <- local({
+
+    if (identical(options$error, TRUE))
+      return(TRUE)
+
+    if (!requireNamespace("rmarkdown", quietly = TRUE))
+      return(TRUE)
+
+    rmarkdown <- asNamespace("rmarkdown")
+    context <- rmarkdown$.render_context
```
jjallaire: We can use some R options that are set only during knit and/or render as described here: https://github.com/yihui/knitr/issues/926

So for knitr:

```r
isTRUE(getOption('knitr.in.progress'))
```

And for R Markdown, you can check for any one of the following values set in opts_knit:

```r
https://github.com/rstudio/rmarkdown/blob/master/R/render.R#L488-L496
```

kevinushey:
Should be ready for merge -- @jjallaire let me know if now's a good time to merge.

PR #361, opened 2018-10-06
Change: table size

bjungbogati:


jjallaire:
This file is generated by pandoc. To make a change like this you'd need to change the corresponding Rmd file. In this case, we got rid of introduction.Rmd so there is no Rmd file to change!

kevinushey:
Should we just remove the article? Note that it's the first result on Google right now (although the redirect to homepage you added does work)

screen shot 2018-10-09 at 2 42 17 pm

kevinushey:
Ah, I see the page itself instructs the browser to redirect so that's fine.

PR #370, opened 2018-10-23
detect virtual environments created with venv

kevinushey:
This PR is the first step in support for Python virtual environments created by `venv`.

Note that these virtual environments are almost identical to those created with `virtualenv`. The only difference is that the `activate_this.py` script is missing. My understanding is that `activate_this.py` is a hack that tries to allow an existing Python session to bind to an existing virtual environment; whereas the preferred route is to instead run e.g.

    source bin/activate && python

which works as expected for virtual environments created by both `virtualenv` and `venv`.

Something we should discuss then is whether we need any other code changes to support virtual environments that don't contain this script. I see we use it at e.g.

https://github.com/rstudio/reticulate/blob/93c9ee9bceee360a9aa3a431735a4c78379c383e/src/python.cpp#L1433-L1458

It's worth noting that `bin/activate` effectively does two things:

1. Sets the `VIRTUAL_ENV` environment variable; its path points to the virtual environment,
2. Updates the PATH, and puts the virtual environment bin directory first.

Assuming that embedded Python sessions also respect the `VIRTUAL_ENV` environment variable, we should validate that setting that before initializing Python is enough to ensure that the newly-launched Python session uses the virtual environment as requested.

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -377,6 +377,13 @@ python_config <- function(python, required_module, python_versions, forced = NUL
   else
     numpy <- NULL
 
+  # check to see if this is a Python virtualenv
+  root <- dirname(dirname(python))
+  virtualenv <- if (is_python_virtualenv(root))
+    root
+  else
+    ""
+
   # check for virtualenv activate script
   activate_this <- file.path(dirname(python), "activate_this.py")
   if (file.exists(activate_this))
@@ -392,6 +399,7 @@ python_config <- function(python, required_module, python_versions, forced = NUL
     python = python,
     libpython = libpython,
     pythonhome = pythonhome,
+    virtualenv = virtualenv,
     virtualenv_activate = virtualenv_activate,
     version_string = version_string,
     version = version,
--- a/R/package.R
+++ b/R/package.R
@@ -108,6 +108,12 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
                           Sys.getenv("PATH"),
                           sep = .Platform$path.sep))
 
+  # if we're a virtual environment then set VIRTUAL_ENV (need to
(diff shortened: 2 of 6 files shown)


kevinushey:
Note that this PR does indeed ensure Python sets `sys.prefix` and `sys.exec_prefix` as we would hope for a virtual environment:

``` r
reticulate::use_virtualenv("~/scratch/venv", required = TRUE)
reticulate::py_config()
#> python:         /Users/kevin/scratch/venv/bin/python
#> libpython:      /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7/lib/python3.7/config-3.7m-darwin/libpython3.7.dylib
#> pythonhome:     /usr/local/opt/python/Frameworks/Python.framework/Versions/3.7:/usr/local/opt/python/Frameworks/Python.framework/Versions/3.7
#> version:        3.7.0 (default, Sep 18 2018, 18:47:22)  [Clang 9.1.0 (clang-902.0.39.2)]
#> numpy:           [NOT FOUND]
#> 
#> NOTE: Python version was forced by use_python function

sys <- reticulate::import("sys")
sys$prefix
#> [1] "/Users/kevin/scratch/venv"
sys$exec_prefix
#> [1] "/Users/kevin/scratch/venv"
```

<sup>Created on 2018-10-23 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

So at least this PR passes the smoke tests. Can you think of anything else I should double-check just to be sure?

jjallaire:
Does the `PATH` end up getting updated by a call we make to `Sys.setenv()` ? (I seem to recall that we do this).

jjallaire:
Will merge now, let’s be sure to add a NEWS item

PR #371, opened 2018-10-24
allow creation of virtual environments with 'venv'

kevinushey:
This PR teaches reticulate how to create Python virtual environments when the `venv` module is available. When the `venv` module is available we'll use that to build the virtual environment; if not we'll fall back to trying the `virtualenv` utility on the PATH.

This PR does make one relatively big change: rather than using the version of Python discovered on the PATH, we instead use the version of Python that reticulate was configured with. This is necessary because the `venv` module only creates virtual environments associated with its own Python installation; e.g. one cannot run `python3 -m venv --python /usr/bin/python` to use Python 3 to create a Python 2 virtual environment.

One other alternate would be for us to allow users to simply provide the path to a Python binary, and we could then figure out the rest from there.

Also worth stating that virtual environments created by `virtualenv` and `python -m venv` are fundamentally the same; the only real difference is the absence of an `activate_this.py` script with Python 3 (but it seems unneeded as long as `VIRTUAL_ENV` is set before Python is launched)

### DIFF
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -52,18 +52,26 @@ virtualenv_create <- function(envname) {
 
   if (!utils::file_test("-d", virtualenv_path) || !file.exists(virtualenv_bin("activate"))) {
     cat("Creating virtualenv at ", virtualenv_path, "\n")
-    result <- system2(config$virtualenv, shQuote(c(
-      "--system-site-packages",
-      "--python", config$python,
-      path.expand(virtualenv_path)))
-    )
+
+    prefix <- if (isTRUE(config$venv))
+      paste(config$python, "-m venv")
+    else
+      config$virtualenv
+
+    args <- "--system-site-packages"
+    if (!isTRUE(config$venv))
+      args <- c(args, "--python", config$python)
+    args <- c(args, path.expand(virtualenv_path))
+
+    command <- paste(prefix, paste(shQuote(args), collapse = " "), sep = " ")
+    result <- system(command)
+
     if (result != 0L)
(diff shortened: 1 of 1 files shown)


R/virtualenv.R
```diff
@@ -198,31 +206,31 @@ virtualenv_config <- function() {
...
+  # use currently configured version of Python
+  config <- py_config()
+  python <- config$python
+
+  # detect Python 2 vs Python 3
+  version <- package_version(config$version)
+  pyver <- if (version < "3") "" else "3"
 
   # find required binaries
   pip <- python_unix_binary(paste0("pip", pyver))
```
jjallaire: I'm wondering if all of our uses of `python_unix_binary()` now need to be replaced with the result of `py_config()` (because otherwise we might use a mismatched version of python and pip, or python and virtualenv).

kevinushey: That's a good idea. Perhaps instead of finding the associated binary on the PATH we should just check whether importing the associated Python module succeeds?

jjallaire: I think those aren't just tests though, they are attempts to actually locate the binary so that we can call it. Of course, we could actually change things so we don't call the binary anymore but rather do e.g `python -m pip ...`, that might be an improvement. In any case, I think we need to either find the binary by looking alongside `python` (assuming that's where pip and virtualenv always live) or we need to invoke them via the `-m` command line arg).

kevinushey: Ahh, got it -- I understand now. I'll spend some more time with this PR and see if I can clean this up.

kevinushey:
Implemented in a separate PR.

Issue #291, opened 2018-06-20
Possible to change libpython path in `py_discover_config()`?


stephaniehicks:
I'm trying to change the `libpython` path as listed in `py_discover_config()`. I know you can force a change in the python path with 
```
Sys.setenv('RETICULATE_PYTHON'='/yourpathhere/bin/python')
py_discover_config()
```
But is it possible to force a change in the library path (`libpython`)? Looking at line 333 in the [config.R](https://github.com/rstudio/reticulate/blob/master/R/config.R) file, it looks like it should be possible to set `config$LIBPL`, but I couldn't figure out how based on the vignettes?

jjallaire:
No, there isn't a way to change the path to libpython (that's something we deduce from the python binary by interrogating it). It's not 100% foolproof though so if there's a scenario where we don't successfully detect the path let us know and we can perhaps work around it.

stephaniehicks:
I'm working on our cluster and I have installed `sklearn` in my home directory (`/users/shicks1/.local/lib/python2.7/site-packages`) 

```
[compute-043 /users/shicks1]$ module load python/2.7.9
[compute-043 /users/shicks1]$ python
Python 2.7.9 (default, May  6 2015, 15:41:33) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sklearn
>>> import sys
>>> print '\n'.join(sys.path)
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python2.7
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python27.zip
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python2.7/plat-linux2
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python2.7/lib-tk
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python2.7/lib-old
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python2.7/lib-dynload
/users/shicks1/.local/lib/python2.7/site-packages
/jhpce/shared/community/compiler/gcc/4.4.7/python/2.7.9/lib/python2.7/site-packages
```

[...]

PR #374, opened 2018-11-01
Search for any unconventional Python library extensions in Linux

iambaim:
In some Linux distributions such as Centos 7, Python library is only available as `libpython2.7.so.1.0` (if `python-devel` package is not installed). This patch ensures that these library files are found.

Related issue: https://github.com/rstudio/reticulate/issues/291#issuecomment-424299557, https://github.com/rocker-org/rocker/issues/296 .

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -335,23 +335,19 @@ python_config <- function(python, required_module, python_versions, forced = NUL
     python_libdir_config <- function(var) {
       python_libdir <- config[[var]]
       ext <- switch(Sys.info()[["sysname"]], Darwin = ".dylib", Windows = ".dll", ".so")
-      libpython <- file.path(python_libdir, paste0("libpython" , version, c("", "m"), ext))
-      libpython_exists <- libpython[file.exists(libpython)]
-      if (length(libpython_exists) > 0)
-        libpython_exists[[1]]
-      else
-        libpython[[1]]
+      pattern <- paste0("^libpython", version, "m?", ext)
+      libpython <- list.files(python_libdir, pattern = pattern, full.names = TRUE)
     }
-    if (!is.null(config$LIBPL)) {
-      libpython <- python_libdir_config("LIBPL")
-      if (!file.exists(libpython)) {
-        if (!is.null(config$LIBDIR))
-          libpython <- python_libdir_config("LIBDIR")
+    for (libsrc in c("LIBPL", "LIBDIR")) {
+      if (!is.null(config[[libsrc]])) {
+        libpython <- python_libdir_config(libsrc)
+        if (length(libpython))
+          break
         else
           libpython <- NULL
+      } else {
+        libpython <- NULL
       }
-    } else {
-      libpython <- NULL
     }
   }
 
@@ -397,7 +393,7 @@ python_config <- function(python, required_module, python_versions, forced = NUL
   # return config info
   structure(class = "py_config", list(
     python = python,
-    libpython = libpython,
+    libpython = libpython[1],
     pythonhome = pythonhome,
     virtualenv = virtualenv,
     virtualenv_activate = virtualenv_activate,
--- a/R/package.R
+++ b/R/package.R
@@ -65,8 +65,8 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   # check for basic python prerequsities
   if (is.null(config)) {
     stop("Installation of Python not found, Python bindings not loaded.")
-  } else if (!is_windows() && (is.null(config$libpython) || !file.exists(config$libpython))) {
-    stop("Python shared library '", config$libpython, "' not found, Python bindings not loaded.")
+  } else if (!is_windows() && is.null(config$libpython)) {
+    stop("Python shared library not found, Python bindings not loaded.")
   } else if (is_incompatible_arch(config)) {
     stop("Your current architecture is ", current_python_arch(), " however this version of ",
          "Python is compiled for ", config$architecture, ".")


skeydan:
Just for curiosity's sake, do you happen to know if this was a dependency for some other package, or did it come with the distro?

I just checked on my (Fedora, not CentOS) system and interestingly, I have a 

```libpython3.so```

which came with the distro and in addition, that same

```libpython2.7.so.1.0``` (with no symlink)

that came on a later date.

**Comment by iambaim** [comment]
> Just for curiosity's sake, do you happen to know if this was a dependency for some other package, or did it come with the distro?

@skeydan Interesting question... As far as I know, both are provided by the distro. I did some digging, in Centos 7, the `libpython2.7.so` is provided by the `python-devel` package:
```console
$ yum whatprovides *libpython2.7.so
Loaded plugins: auto-update-debuginfo, fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: ftp.uninett.no
 * epel: mirror.nsc.liu.se
 * epel-debuginfo: mirror.nsc.liu.se
 * extras: ftp.uninett.no
 * updates: ftp.uninett.no
python-devel-2.7.5-68.el7.x86_64 : The libraries and header files needed for Python development
Repo        : base
Matched from:
Filename    : /usr/lib64/libpython2.7.so
Filename    : /usr/lib64/python2.7/config/libpython2.7.so
...
```
While "out-of-the-box" Python provides only the `libpython2.7.so.1.0` file:
```console
$ yum deplist python
Loaded plugins: auto-update-debuginfo, fastestmirror, langpacks
Loading mirror speeds from cached hostfile
 * base: ftp.uninett.no
 * epel: ftp.acc.umu.se
 * epel-debuginfo: mirror.nsc.liu.se
 * extras: ftp.uninett.no
 * updates: ftp.uninett.no
package: python.x86_64 2.7.5-69.el7_5
  dependency: libc.so.6(GLIBC_2.2.5)(64bit)
   provider: glibc.x86_64 2.17-222.el7
  dependency: libdl.so.2()(64bit)
   provider: glibc.x86_64 2.17-222.el7
  dependency: libm.so.6()(64bit)
   provider: glibc.x86_64 2.17-222.el7
  dependency: libpthread.so.0()(64bit)
   provider: glibc.x86_64 2.17-222.el7
  dependency: libpython2.7.so.1.0()(64bit)
   provider: python27-python-libs.x86_64 2.7.13-5.el7
   provider: python-libs.x86_64 2.7.5-69.el7_5
  dependency: libutil.so.1()(64bit)
   provider: glibc.x86_64 2.17-222.el7
  dependency: python-libs(x86-64) = 2.7.5-69.el7_5
   provider: python-libs.x86_64 2.7.5-69.el7_5
  dependency: rtld(GNU_HASH)
   provider: glibc.x86_64 2.17-222.el7
   provider: glibc.i686 2.17-222.el7

$ repoquery -l python-libs | grep libpython2.7
/usr/lib/libpython2.7.so.1.0
/usr/share/systemtap/tapset/libpython2.7-32.stp
/usr/lib64/libpython2.7.so.1.0
/usr/share/systemtap/tapset/libpython2.7-64.stp
...
```

skeydan:
Interesting!

jjallaire:
@kevinushey Could you take a quick look at this just to sanity check the implementation?

R/config.R
```diff
@@ -336,6 +336,8 @@ python_config <- function(python, required_module, python_versions, forced = NUL
       python_libdir <- config[[var]]
       ext <- switch(Sys.info()[["sysname"]], Darwin = ".dylib", Windows = ".dll", ".so")
       libpython <- file.path(python_libdir, paste0("libpython" , version, c("", "m"), ext))
+      # look for any unconventional library extensions (e.g., in Centos 7)
+      libpython <- c(libpython, unlist(lapply(libpython, function(x) list.files(path = dirname(x), pattern = paste0(basename(x), "*"), full.names = TRUE))))
```
kevinushey: I think this would be simpler as:

``` r
pattern <- paste("^libpython", version, "m?", ext)
libpython <- list.files(python_libdir, pattern = pattern, full.names = TRUE)
```

iambaim: > pattern <- paste("^libpython", version, "m?", ext)

Agree. Change the above to `paste0` and it's perfect!

jjallaire:
@kevinushey I'll leave this to you to merge to make sure everything is buttoned up the way you wanted it to be.

R/config.R
iambaim: I've updated this request as per @kevinushey suggestion. I had to add several other changes so that we only need `list.files()` to find the Python library files. As the new `for` loop, it's actually not that necessary but can simplify the nested `if`/`else`.

If this is too much I can revert back to only change two lines from my initial pull request :-)

R/config.R
```diff
@@ -335,23 +335,18 @@ python_config <- function(python, required_module, python_versions, forced = NUL
...
-    if (!is.null(config$LIBPL)) {
-      libpython <- python_libdir_config("LIBPL")
-      if (!file.exists(libpython)) {
-        if (!is.null(config$LIBDIR))
-          libpython <- python_libdir_config("LIBDIR")
-        else
+    for (libsrc in c("LIBPL", "LIBDIR")) {
+      if (!is.null(config[[libsrc]])) {
+        libpython <- python_libdir_config(libsrc)[1]
+        if (is.na(libpython))
```
kevinushey: When could this be `NA`? Won't `list.files()` always return a character vector of length >= 0?

R/config.R
```diff
@@ -335,23 +335,18 @@ python_config <- function(python, required_module, python_versions, forced = NUL
...
           libpython <- NULL
-      }
-    } else {
-      libpython <- NULL
+        else
+          break
+      } else
```
kevinushey: Is there a missing `{` here?

R/package.R
```diff
@@ -65,8 +65,8 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   # check for basic python prerequsities
   if (is.null(config)) {
     stop("Installation of Python not found, Python bindings not loaded.")
-  } else if (!is_windows() && (is.null(config$libpython) || !file.exists(config$libpython))) {
-    stop("Python shared library '", config$libpython, "' not found, Python bindings not loaded.")
+  } else if (!is_windows() && is.null(config$libpython)) {
```
kevinushey: Can you explain why this change was needed?

R/config.R
kevinushey: Oh, because indexing a zero-length character will return `NA`, e.g.

``` r
> character()[1]
[1] NA
```

I think it would be more clear to just use e.g.

``` r
libpython <- python_libdir_config(libsrc)[1]
if (length(libpython)
  break
```

and then use the first entry of `libpython` downstream as appropriate. But the existing code should be correct too.

R/package.R
iambaim: `config$libpython` is populated by `python_libdir_config()`. Before the proposed change to the function, the library filename is pre-determined. However, this is not the case now. Therefore:
1. I don't think we need to check if the library file exists anymore (`list.files` does this).
2. If no library found, we don't need to print `NULL`.

R/config.R
iambaim: Not really. I was just trying to save one line there, my bad :)

R/config.R
iambaim: Good idea, I'll update this pull.

randy3k:
An unexpected effect of the PR is that it allows `libpython` to be an empty string. 
Doing `dlopen("")` will search the symbols in the main process, and it is exactly what I need in #479.

PR #384, opened 2018-11-14
allow for functions called with 'r' helper

### LINKED ISSUE #383: calling r.plot(...) from python does not work
Hi there!

I like reticulate so much!

however this fails:
```R
reticulate::py_eval("r.plot([1,2,3,2])")
```

I was expecting a plot to be displayed ( I am using  RStudio)

I got my expected behavior by modifying `py_inject_r` line 1347:
https://github.com/rstudio/reticulate/blob/196f5b7bae5d376fbe5b13405823a0498248f92e/R/python.R#L1347
```R
getter <- function(self, code) {        
        str_code <- reticulate:::as_r_value(code)
        r_variable <- eval(parse(text = str_code), envir = envir)
        
        if ( is.function(r_variable) ) {
            main_f <- reticulate::import_main(convert = T) # convert required for functions
            main_f[[paste0("_r_",str_code)]]  <-  r_variable
            return(reticulate::py_eval(paste0("_r_",str_code)))
        }
        
        return(r_to_py(r_variable))
    }

```

This would also work from R Notebook (and it is so great to call r.plot from python there!)

I hope you can make a cleaner fix, but at least this drives you to the issue.

Keep going on  !

kevinushey:
Closes #383.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@
 
 Install the development version with: `devtools::install_github("rstudio/reticulate")`
 
+- The `r` helper object (used for evaluating R code from Python) now better
+  handles conversion of R functions. (#383)
+
 - The `use_virtualenv()` function now understands how to bind to virtual
   environments created by the Python `venv` module.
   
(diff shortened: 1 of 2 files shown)


battmanux:
Hello,

I tried your fix and it works in my case.

Thanks !

PR #401, opened 2018-12-05
revamp python virtualenv helper internals

kevinushey:
This PR revamps the internals around how Python virtual environments are created + used in reticulate.

Primary goals here:

1. Refactor the existing code, as it was getting a bit difficult to manage;
2. Create a separate `pip` interface that can be used internally for installing / removing packages,
3. Better handle the nuances between `virtualenv`, `venv`, and where they might (might not) be available for different Python installations.

One major change this PR makes: we now attempt to use `virtualenv` and `venv` as modules whenever possible. This avoids some potential ambiguities -- for example, when one sees a `virtualenv` executable on the PATH, it's hard to ascertain which Python installation it's actually associated with, and in general it's safer to use Python x.y.z to create a virtual environment associated with Python version x.y.z. This advice comes from https://docs.python.org/3/library/venv.html:

> Note The pyvenv script has been deprecated as of Python 3.6 in favor of using python3 -m venv to help prevent any potential confusion as to which Python interpreter a virtual environment will be based on.

We still use the `pip` binary as needed, though -- since virtual environments always create this as a companion to the regular Python binary + other infrastructure.

### DIFF
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -204,7 +204,7 @@ virtualenv_default_python <- function(python) {
   if (!is.null(python))
     return(python)
 
-  config <- py_config()
+  config <- py_discover_config()
   normalizePath(config$python, winslash = "/")
 }
 
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -2,21 +2,27 @@
 #'
 #' R functions for managing Python [virtual environments](https://virtualenv.pypa.io/en/stable/).
 #'
-#' @details
-#' Virtual environments are by default located at `~/.virtualenvs`. You can change this
-#' behavior by defining the `WORKON_HOME` environment variable.
+#' Virtual environments are by default located at `~/.virtualenvs` (accessed
+#' with the `virtualenv_root` function). You can change the default location by
+#' defining defining the `WORKON_HOME` environment variable.
 #'
 #' Virtual environment functions are not supported on Windows (the use of
 #' [conda environments][conda-tools] is recommended on Windows).
 #'
-#' @param envname The name of, or path to, a Python virtual environment.
+#' @param envname The name of, or path to, a Python virtual environment. If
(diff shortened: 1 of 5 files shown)


kevinushey:
Also worth noting -- IIUC, it's no longer necessary to invoke the `activate` scripts since those are primarily intended for interactive usages. (`activate` basically just ensures that the `python` executable in that virtual environment is placed first on the PATH).

We might want to bring back something similar if we intend to allow for users to call e.g. `system()` commands using a virtual environment's python, but as far as I can see we don't currently do this.

kevinushey:
Finally, the `virtualenv_create()` function has gained an argument `python = NULL`, where `NULL` means "use the version of Python that reticulate is currently bound to" and is otherwise the user-supplied path to an existing Python binary.

R/virtualenv.R
```diff
@@ -1,309 +1,260 @@
...
-  matches <- regexec("^[^ ]+\\s+(\\d+)\\.(\\d+).*$", result)
-  matches <- regmatches(result, matches)[[1]]
-  if (length(matches) != 3)
-    stop("Unable to parse Python version '", result[[1]], "'", call. = FALSE)
+  if (!is.null(python))
+    return(python)
 
-  # return as R numeric version
-  numeric_version(paste(matches[[2]], matches[[3]], sep = "."))
+  config <- py_config()
```
jjallaire: We may want to use `py_discover_config()` here. One of the goals of the installation functions is to avoid initializing reticulate in the current process -- this is because if you install into a virtualenv using a default python session then you can't immediately import the Python library (you'd need to restart R). This affects the simplicity of getting started materials (e.g. `install_tensorflow()`, then restart R, then do `library(tensorflow)`.

That's one reason we put in `py_discover_config()` -- it will tell you the version of Python that *would* be loaded without actually loading it.

Issue #208, opened 2018-03-31
reticulate + rpy2 sharing R process


SemanticBeeng:
Cannot get to work `rpy2` v 2.9.2 and `reticulate` v 1.6 with R v 3.4.3 and Python 3.5.5 (or 3.6.5) to share the same instance of R embedded process / engine instance.

context - reviewed these but still cannot make this work
https://github.com/rstudio/reticulate/issues/98
https://github.com/rstudio/reticulate/pull/104
https://github.com/rstudio/reticulate/issues/138

Environment
```
$ uname -a
Linux bigdata-vm 4.4.0-83-generic #106-Ubuntu SMP Mon Jun 26 17:54:43 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```
Python and R setup with `conda` `environment.yml`

```
name: dev-ds
channels:
- r
- conda-forge
- anaconda 
- defaults
dependencies:
- conda-forge::python=3.5.5
- conda-forge::numpy=1.14.2
- conda-forge::pandas=0.22.0
# ==============
# R core
# ==============
- r-base=3.4.3
- r-magrittr=1.5
- r-reticulate    # is 1.3 but need 1.6* so re-installed manually from R
- r-dplyr=0.7.4
- r-dbi=0.7
- r-r6=2.2.2
- r-magrittr=1.5
- r-jsonlite=1.5
- r-dplyr=0.7.4
- r-purrr=0.2.4
- r-bindr=0.1
# ==============
# R application runtime 
# ==============
- r-readr=1.1.1
- r-config=0.2
- r-xml
- r-reshape2
- r-data.table
- r-forecast=8.2
- r-tseries
- nlopt
- r-nloptr=1.0.4
# ==============
# R application dev
# ==============
- r-git2r     # 0.20.0
- r-rcpp=0.12.14 # 0.12.16
- r-devtools=1.13.4
- r-testthat=2.0.0
- r-roxygen2
- r-sparklyr  # 0.7.0

# ==============
# Python core
# ==============
- pip:
  - rpy2==2.9.2 
```
Using R from under the `conda`  env:

```
library(reticulate)
reticulate::import("sys")
rpy2 <- import('rpy2')
robjects <- import('rpy2.robjects')
```
exists with _"R is already initialized"_ caused by last line.

From a Python script `some.py`

```
from rpy2.robjects import pandas2ri
from rpy2.robjects.packages import importr
import rpy2.robjects as robjects

def main():
    r = robjects.r

[...]

PR #429, opened 2019-02-10
Set env variable before initialization

lgautier:
Set environment variable before initializing Python (see https://github.com/rstudio/reticulate/issues/208#issuecomment-462089957). Also linked to issue #98 (use environment variables to track initialization state).

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -113,15 +113,31 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   # set as appropriate)
   if (nzchar(config$virtualenv))
     Sys.setenv(VIRTUAL_ENV = config$virtualenv)
+  
+  # set R_SESSION_INITIALIZED flag (used by rpy2)
+  curr_session_env <- Sys.getenv("R_SESSION_INITIALIZED", unset = NA)
+  Sys.setenv(R_SESSION_INITIALIZED = sprintf('PID=%s:NAME="reticulate"', Sys.getpid()))
 
   # initialize python
-  py_initialize(config$python,
-                config$libpython,
-                config$pythonhome,
-                config$virtualenv_activate,
-                config$version >= "3.0",
-                interactive(),
-                numpy_load_error)
+  tryCatch(
+    {
+      py_initialize(config$python,
+                    config$libpython,
+                    config$pythonhome,
+                    config$virtualenv_activate,
+                    config$version >= "3.0",
+                    interactive(),
+                    numpy_load_error)
+    },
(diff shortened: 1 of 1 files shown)


R/package.R
```diff
@@ -113,15 +113,31 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
...
+                    config$virtualenv_activate,
+                    config$version >= "3.0",
+                    interactive(),
+                    numpy_load_error)
+    },
+    error = function(e) {
+      if (is.na(curr_session_env)) {
+        Sys.unsetenv("R_SESSION_INITIALIZED")
+      } else {
+        Sys.setenv(curr_session_env)
```
jjallaire: Does this want to be: 

```r
Sys.setenv(R_SESSION_INITIALIZED = curr_session_env)
```

R/package.R
```diff
@@ -113,15 +113,31 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   # set as appropriate)
   if (nzchar(config$virtualenv))
     Sys.setenv(VIRTUAL_ENV = config$virtualenv)
+  
+  # set R_SESSION_INITIALIZED flag (used by rpy2)
+  curr_session_env <- Sys.getenv("R_SESSION_INITIALIZED", unset = NA)
+  Sys.setenv(R_SESSION_INITIALIZED = sprintf('PID=%s:NAME="reticulate"', Sys.getpid()))
```
jjallaire: I wonder if we could manage the failure case with an `on.exit()`. Something like this:

```r
on.exit({
  if (!config$available) {
    if (is.na(curr_session_env))
      Sys.unsetenv("R_SESSION_INITIALIZED")
    else 
      Sys.setenv(R_SESSION_INITIALIZED = curr_session_env)
  }
}, add = TRUE)
```

That way we wouldn't have to mess with the control flow below at all.

R/package.R
lgautier: Oops. Yes, that's a mistake in my PR. Fixed.

R/package.R
lgautier: I thought about but the `on.exit()` would have to add the handle the case of a successful initialization of Python and it did not seem much of a gain since the control of flow is in the `on.exit()` anyway.

Either of the two is a stylistic choice, I'd say. The PR's main purpose is to have the environment variable set before Python is initialized, I am happy with any way to handle failure during initialization you like :-) .

jjallaire: Okay, I'm good with your implementation then.

Issue #437, opened 2019-02-27
Error initializing python with virtualenv > 16.1


mmarchetti:
When trying to run python from an existing virtualenv created with virtualenv > 16.1, the following error appears:
```
Error in py_initialize(config$python, config$libpython, config$pythonhome,  :
  NameError: name 'prev' is not defined

Detailed traceback:
  File "<string>", line 46, in <module>
  File "<string>", line 46, in <listcomp>
```

Virtualenv and python work fine outside of reticulate. Downgrading to virtualenv 16.1 resolves the issue and allows reticulate to work normally. 

This seems to be coming from [this line](https://github.com/pypa/virtualenv/commit/971d341ce383420af342a8f9d531ce77242fda0e#diff-60f1b0e736e751c0ba89c96ee165766eR46) in virtualenv, called from reticulate [here](https://github.com/rstudio/reticulate/blob/7869d13544ecc13284a8a3f12025643e80c1e6a1/src/python.cpp#L1455).

A reproducible test run using docker is below.

```
$ docker run -it python:3.7 bash
root@35a5ef2449db:/# pip --version
pip 19.0.1 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)
root@35a5ef2449db:/# python --version
Python 3.7.2
root@35a5ef2449db:/# apt update
Ign:1 http://deb.debian.org/debian stretch InRelease
Get:2 http://deb.debian.org/debian stretch-updates InRelease [91.0 kB]
Get:3 http://security.debian.org/debian-security stretch/updates InRelease [94.3 kB]
Get:4 http://deb.debian.org/debian stretch Release [118 kB]
Get:5 http://deb.debian.org/debian stretch Release.gpg [2434 B]
Get:6 http://deb.debian.org/debian stretch-updates/main amd64 Packages [11.1 kB]
Get:7 http://security.debian.org/debian-security stretch/updates/main amd64 Packages [474 kB]
Get:8 http://deb.debian.org/debian stretch/main amd64 Packages [7084 kB]
Fetched 7874 kB in 2s (3051 kB/s)
Reading package lists... Done
Building dependency tree
Reading state information... Done
22 packages can be upgraded. Run 'apt list --upgradable' to see them.
root@35a5ef2449db:/# apt install r-base
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
[...]

PR #439, opened 2019-02-27
avoid use of local dict (#437)

kevinushey:
Fixes https://github.com/rstudio/reticulate/issues/437. For whatever reason, the use of a local dict causes some issues with name resolution when attempting to source `activate_this.py`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: reticulate
 Type: Package
 Title: Interface to 'Python'
-Version: 1.10.0.9004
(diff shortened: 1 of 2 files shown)


jjallaire:
This looks like a good fix. We should probably try to get this onto CRAN right away. Both b/c of this fix but also b/c we haven't had a CRAN release since last August and we've got quite some work that we'll want for the IDE release (e.g. work on matplotlib). @kevinushey Can you pull a release together?

kevinushey:
Will do!

Issue #441, opened 2019-02-28
`use_python()` should throw an error if Python is already embedded and the path doesn't match


s-u:
`use_python()` will silently succeed without having any effect if the session is already initialized and thus the path cannot be changed. The user typically has no control over whether Python has already been initialized or not (since any other package could have initialized it), so it is important to know if the requested path cannot be honored.

The most common pitfall is that some package initialized python2 and you need python3, but reticulate won't warn that you're not getting what you asked for (and thus nothing will work).

This was partially discussed in #418 but the answer there ("we'll find the version of Python you need") makes no sense to me since the whole issue is that you get the wrong version and won't know it.

jjallaire:
The original concept behind `use_python()` is that it is advisory (i.e. the same script running on different systems might find Python in different places/environments, so all would be listed). There is a `required` flag you can add to make sure that the specified version of Python actually exists (an error is thrown if it doesn't), but that doesn't cover your case of Python being already initialized. Perhaps a better behavior would be: if the `required` flag is used AND Python is already initialized and not pointing to that version, then you get an error.

s-u:
I agree, that sounds to me like a reasonable behavior.

jjallaire:
PR which implements the new behavior: https://github.com/rstudio/reticulate/pull/442

PR #442, opened 2019-03-03
use_python: verify required version when python is already initialized

jjallaire:
Addresses #441

@kevinushey Could you do a code review?

### DIFF
--- a/R/use_python.R
+++ b/R/use_python.R
@@ -18,6 +18,17 @@ use_python <- function(python, required = FALSE) {
   if (required && !file_test("-f", python) && !file_test("-d", python))
     stop("Specified version of python '", python, "' does not exist.")
 
+  # if required == TRUE and python is already initialized then confirm that we
+  # are using the correct version
+  if (required && is_python_initialized()) {
+    if (!identical(normalize_python_path(py_config()$python),
+                   normalize_python_path(python))) {
+      stop("Required version of Python '", python ,"' cannot be used ",
+           "because another version of Python ('", py_config()$python,
+           "') is already initialized ", "for this process.")
+    }
+  }
+
   if (required)
     .globals$required_python_version <- python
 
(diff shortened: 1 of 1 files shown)


kevinushey:
One minor worry is that since `normalize_python_path()` doesn't call `normalizePath()` different ways of referring to the same path will cause this to err; e.g.

```r
> python <- Sys.which("python")
> python
                                                                    python 
"C:\\Users\\kevin\\AppData\\Local\\Programs\\Python\\Python37\\python.exe" 
> 
> reticulate::use_python(python, required = TRUE)
> reticulate::py_config()
python:         C:\Users\kevin\AppData\Local\Programs\Python\Python37\python.exe
libpython:      C:/Users/kevin/AppData/Local/Programs/Python/Python37/python37.dll
pythonhome:     C:\Users\kevin\AppData\Local\Programs\Python\Python37
version:        3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)]
Architecture:   64bit
numpy:          C:\Users\kevin\AppData\Roaming\Python\Python37\site-packages\numpy
numpy_version:  1.16.2

NOTE: Python version was forced by use_python function
> reticulate::use_python(python, required = TRUE)
> 
> reticulate::use_python(normalizePath(python, winslash = "/"), required = TRUE)
Error in reticulate::use_python(normalizePath(python, winslash = "/"),  : 
  Required version of Python 'C:/Users/kevin/AppData/Local/Programs/Python/Python37/python.exe' cannot be used because another version of Python ('C:\Users\kevin\AppData\Local\Programs\Python\Python37\python.exe') is already initialized for this process.
```

I don't think that will be a problem in practice since I doubt user's will attempt to refer to the same Python binary in different ways but worth considering if there's a scenario where a user could inadvertently trip this.

jjallaire:
Good catch. Just pushed a change to address this.

s-u:
Great, thanks, works for me!

PR #444, opened 2019-03-04
allow the use of channel in conda_install

### LINKED ISSUE #443: conda_install from different channels
When installing a package, conda allows to choose different channels.
However, the function `conda_install` in reticulate supports only `conda-forge` as channel. 
Therefore, for some packages we are forced to use the command-line for installing them. 

Is it possible to add the argument `channel` to the function `conda_install`? 

From the conda 4.6.7 help:
```
  -c CHANNEL, --channel CHANNEL
                        Additional channel to search for packages. These are
                        URLs searched in the order they are given (including
                        file:// for local directories). Then, the defaults or
                        channels from .condarc are searched (unless
                        --override-channels is given). You can use 'defaults'
                        to get the default packages for conda. You can also
                        use any name and the .condarc channel_alias value will
                        be prepended. The default channel_alias is
                        http://conda.anaconda.org/.
```

dimitri-mar:
Add the argument channel to the `conda_install` function. Fix #443 

It would translate to `conda install -c my_channel package`.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -113,6 +113,7 @@ conda_remove <- function(envname, packages = NULL, conda = "auto") {
 }
 
 #' @param forge Include the [Conda Forge](https://conda-forge.org/) repository.
+#' @param channel Include additional channel to search for packages.
 #' @param pip_ignore_installed Ignore installed versions when using pip. This is `TRUE` by default
 #'   so that specific package versions can be installed even if they are downgrades. The `FALSE`
 #'   option is useful for situations where you don't want a pip install to attempt an overwrite
@@ -124,7 +125,7 @@ conda_remove <- function(envname, packages = NULL, conda = "auto") {
 #' @keywords internal
 #'
 #' @export
-conda_install <- function(envname, packages, forge = TRUE, pip = FALSE, pip_ignore_installed = TRUE, conda = "auto") {
+conda_install <- function(envname, packages, forge = TRUE, channel = NULL, pip = FALSE, pip_ignore_installed = TRUE, conda = "auto") {
 
   # resolve conda binary
   conda <- conda_binary(conda)
@@ -151,6 +152,8 @@ conda_install <- function(envname, packages, forge = TRUE, pip = FALSE, pip_igno
     args <- c("install")
     if (forge)
       args <- c(args, "-c", "conda-forge")
+    if (! is.null(channel))
+      args <- c(args, "-c", channel)
     args <- c(args, "--yes", "--name", envname, packages)
     result <- system2(conda, shQuote(args))
   }


VallinP:
An easy way is to add only one specific channel in the arguments, like bioconda which is very usefull for linking R/Bioconductor functions with Python Machine Learning possibilities.
My suggestion : https://github.com/VallinP/reticulate/blob/master/R/conda.R#L136

ShixiangWang:
Hope this feature can be implemented.

R/conda.R
```diff
@@ -151,6 +152,8 @@ conda_install <- function(envname, packages, forge = TRUE, pip = FALSE, pip_igno
     args <- c("install")
     if (forge)
```
TomKellyGenetics:
```suggestion
    if (forge & ! is.null(channel) )
```
Required to run `conda -c channel` rather than `conda -c -conda-forge -c channel` which returns an error.

kevinushey: Are you sure? https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/channels.html suggests this should be okay, e.g.

```
$ conda install scipy --channel conda-forge --channel bioconda
```
``

TomKellyGenetics: I've installed @dimitri-mar's branch in R and running it does not work on my branch unless this change is made. I got an error calling it on a different channel. It seems that in this case, there was not windows build on conda-forge and only on another channel.

BenWiseman:
Any progress? As it is, reticulate is just not usable for me with the gimped version of conda being used.

kevinushey:
I've committed a similar version of this PR to master now.

R/conda.R
TomKellyGenetics: To clarify, this is required to install "leidenalg" on windows systems (as an example for testing). Running `conda -c conda-forge -c vtraag leidenalg` returns an error. It needs to call `conda -c vtraag leidenalg` as there is no windows build in conda-forge in this case.

TomKellyGenetics: This is resolved by this PR #752. Thanks! https://github.com/rstudio/reticulate/pull/752/files#diff-ed0960af4336904d09c7d5e7e02472d2R133-R142

I'm having trouble installing it from CRAN or GitHub in version 1.15 though.

```
Error: package or namespace load failed for ‘reticulate’ in get(Info[i, 1], envir = env):
 lazy-load database '/Users/tom/Library/R/3.6/library/reticulate/R/reticulate.rdb' is corrupt
```

As a workaround this installs a working version:

```sh
git clone https://github.com/rstudio/reticulate.git
cd reticulate
```
```r
library("devtools")
document()
install()
```

PR #447, opened 2019-03-05
modify PATH before discovering Python config

### LINKED ISSUE #367: reticulate marks numpy as missing in Anaconda installation
```R
library(reticulate)
reticulate::use_python("~/Anaconda3/python.exe", required = TRUE)
reticulate::py_config()
numpy <- import("numpy", convert = TRUE)
numpy[["__path__"]]
```

I see:

```
> library(reticulate)
> reticulate::use_python("~/Anaconda3/python.exe", required = TRUE)
> reticulate::py_config()
python:         C:/Users/kevin/Anaconda3/python.exe
libpython:      C:/Users/kevin/Anaconda3/python37.dll
pythonhome:     C:\Users\kevin\ANACON~1
version:        3.7.0 (default, Jun 28 2018, 08:04:48) [MSC v.1912 64 bit (AMD64)]
Architecture:   64bit
numpy:           [NOT FOUND]

NOTE: Python version was forced by use_python function
> numpy <- import("numpy", convert = TRUE)
> numpy[["__path__"]]
[1] "C:\\Users\\kevin\\ANACON~1\\lib\\site-packages\\numpy"
```

This was with Anaconda 3 just installed today. If I try to import `numpy` from the console directly, I see:

```
>>> import numpy
  File "C:\Users\kevin\Anaconda3\lib\site-packages\numpy\core\__init__.py", line 16, in <module>
    from . import multiarray
ImportError: DLL load failed: The specified module could not be found.

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "C:\Users\kevin\Anaconda3\lib\site-packages\numpy\__init__.py", line 142, in <module>
    from . import add_newdocs
  File "C:\Users\kevin\Anaconda3\lib\site-packages\numpy\add_newdocs.py", line 13, in <module>
    from numpy.lib import add_newdoc
  File "C:\Users\kevin\Anaconda3\lib\site-packages\numpy\lib\__init__.py", line 8, in <module>
    from .type_check import *
  File "C:\Users\kevin\Anaconda3\lib\site-packages\numpy\lib\type_check.py", line 11, in <module>
    import numpy.core.numeric as _nx
  File "C:\Users\kevin\Anaconda3\lib\site-packages\numpy\core\__init__.py", line 26, in <module>
    raise ImportError(msg)
ImportError:
Importing the multiarray numpy extension module failed.  Most
likely you are trying to import a failed build of numpy.
If you're working with a numpy git repo, try `git clean -xdf` (removes all
files not under version control).  Otherwise reinstall numpy.

Original error was: DLL load failed: The specified module could not be found.
```

But attempting to import `numpy` works fine if run from the Anaconda Prompt. I'm guessing it's because it puts a whole bunch of stuff on the PATH:

```
(base) C:\Users\kevin>echo %PATH%
C:\Users\kevin\Anaconda3;C:\Users\kevin\Anaconda3\Library\mingw-w64\bin;C:\Users\kevin\Anaconda3\Library\usr\bin;C:\Users\kevin\Anaconda3\Library\bin;C:\Users\kevin\Anaconda3\Scripts;C:\Users\kevin\Anaconda3\bin;<...>
```

Do we need to do that as well before running the config script? (It's worth noting that all of this stuff becomes available on the PATH _after_ Python has been initialized)

kevinushey:
Closes #367.

The fix here effectively moves some of the PATH modification we do to before the attempt to discover the Python configuration + load numpy.

There's a bit of extra orchestration required here as we don't want the PATH changes to persist if we're only attempting to discover Python; ie, we only want to save the PATH changes if we're truly initializing Python. I think I got the semantics right here but it's worth double-checking that I'm doing things correctly.

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -294,9 +294,7 @@ python_environments <- function(env_dirs, required_module = NULL) {
     envs
 }
 
-
-
-python_config <- function(python, required_module, python_versions, forced = NULL) {
+python_munge_path <- function(python) {
 
   # add the python bin dir to the PATH (so that any execution of python from
   # within the interpreter, from a system call, or from within a terminal
@@ -326,15 +324,15 @@ python_config <- function(python, required_module, python_versions, forced = NUL
       python_dirs <- c(python_dirs, normalizePath(python_library_bin))
   }
 
-  # restore the PATH when we're done if we're not explicitly initializing Python
-  # (need to differentiate between a plain config discovery and true init)
-  PATH <- Sys.getenv("PATH")
-  if (!.globals$py_initializing)
-    on.exit(Sys.setenv(PATH = PATH), add = TRUE)
+  path_prepend(python_dirs)
 
-  Sys.setenv(PATH = paste(paste(python_dirs, collapse =  .Platform$path.sep),
-                          PATH,
-                          sep = .Platform$path.sep))
+}
+
+python_config <- function(python, required_module, python_versions, forced = NULL) {
+
+  # update and restore PATH when done
+  oldpath <- python_munge_path(python)
+  on.exit(Sys.setenv(PATH = oldpath), add = TRUE)
 
   # collect configuration information
   if (!is.null(required_module)) {
--- a/R/config.R
+++ b/R/config.R
@@ -94,7 +94,6 @@ py_module_available <- function(module) {
 #' @export
 py_discover_config <- function(required_module = NULL, use_environment = NULL) {
 
-
   # if PYTHON_SESSION_INITIALIZED is specified then use it without scanning
   # further (this is a "hard" requirement because an embedding process may
   # set this to indicate that the python interpreter is already loaded)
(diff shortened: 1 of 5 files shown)


R/virtualenv.R
```diff
@@ -204,7 +204,7 @@ virtualenv_default_python <- function(python) {
   if (!is.null(python))
     return(python)
 
-  config <- py_discover_config()
+  config <- py_config()
```
jjallaire: This can't be `py_config()`, because that will actually cause Python to load (and the `virtualenv` family of functions shouldn't do that).

R/config.R
```diff
@@ -299,11 +298,49 @@ python_environments <- function(env_dirs, required_module = NULL) {
...
 python_config <- function(python, required_module, python_versions, forced = NULL) {
 
+  # add the python bin dir to the PATH (so that any execution of python from
+  # within the interpreter, from a system call, or from within a terminal
+  # hosted within the front end will use the same version of python.
+  #
+  # we do this up-front in python_config as otherwise attempts to discover
+  # and load numpy can fail, especially on Windows
+  # https://github.com/rstudio/reticulate/issues/367
+  python_home <- python_dirs <- dirname(python)
```
jjallaire: Did we lose the call to `normalizePath(python_dirs)` here?

kevinushey: You're right, I missed that. I'll bring it back.

R/package.R
```diff
@@ -32,16 +33,22 @@ is_python_initialized <- function() {
...
+    if (!is.null(.globals$delay_load_module)) {
+      required_module <- .globals$delay_load_module
+      use_environment <- .globals$delay_load_environment
+      .globals$delay_load_module <- NULL # one shot
+      .globals$delay_load_environment <- NULL
+      .globals$delay_load_priority <- 0
+    }
+
+    # mark as initializing (so we know to save any PATH changes)
+    .globals$py_initializing <- TRUE
```
jjallaire: `py_disover_config()` can potentially call `py_config()` multiple times, which would in turn cause the `PATH` to be amended multiple times during initialization. I think that rather than using the `py_initializing` flag we should package the PATH manipulating code into a helper function and call it from both places.

kevinushey:
I've updated the PR -- let me know if this looks like what you were imagining.

I extracted the path munging code into its own helper function that's a bit smarter about how it munges the PATH, and make sure it's called both during config discovery (and reset after) and also during Python initialization (and not unset if initialization was successful).

PR #448, opened 2019-03-05
allow default environment for py_install

kevinushey:
This PR teaches `py_install()` to install packages into the currently active virtual environment (if any). This would mean that e.g.

```r
library(reticulate)
virtualenv_create("test")
use_virtualenv("test", required = TRUE)
py_available(TRUE)
py_install("numpy")
```

would install into the active `test` environment, rather than the default `r-reticulate` environment.

The intention in this PR is that (e.g. `renv`) could then automatically initialize `reticulate` when R is launched, and then `py_install()` would then just "do the right thing" based on the environment that `renv` has told `reticulate` to use.

Some questions worth answering if we like this PR:

- Do we want to implement similar handling for Conda environments?
- What about the case where the user has called `use_virtualenv()` but has not yet initialized Python?

### DIFF
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -77,11 +77,13 @@ virtualenv_create <- function(envname, python = NULL) {
 #' @inheritParams virtualenv-tools
 #' @rdname virtualenv-tools
 #' @export
-virtualenv_install <- function(envname, packages, ignore_installed = TRUE) {
+virtualenv_install <- function(envname = NULL, packages, ignore_installed = TRUE) {
 
   path <- virtualenv_path(envname)
   if (file.exists(path))
-    writeLines(paste("Using virtual environment", shQuote(envname), "..."))
+    writeLines(paste("Using virtual environment", shQuote(basename(path)), "..."))
+  else if (is.null(envname))
+    stop("virtualenv_install() called without active virtual environment")
   else
     path <- virtualenv_create(envname)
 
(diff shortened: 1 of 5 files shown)


R/virtualenv.R
```diff
@@ -180,6 +182,18 @@ virtualenv_exists <- function(envname) {
 
 virtualenv_path <- function(envname) {
 
+  # if envname is NULL but Python is already active, then
+  # use the path to the current virtual environment
+  if (is.null(envname) && py_available()) {
+    config <- py_config()
```
jjallaire: This again violates the idea that installation functions shouldn't load Python (especially problematic on Windows where there can be DLL conflicts). I don't think py_install should key off of what's currently loaded, rather it should see whether it's running inside an renv.

jjallaire:
I also don't think that renv should be initializing reticulate when it loads (this should be deferred to first use by a package).

jjallaire:
We can discuss realtime tomorrow if that's easier than going back and forth here.

R/virtualenv.R
kevinushey: Sorry, that makes sense -- I've made that change.

kevinushey:
> I also don't think that renv should be initializing reticulate when it loads (this should be deferred to first use by a package).

I've changed this on the `renv` side so that all we do is set `RETICULATE_PYTHON` on load as appropriate.

R/virtualenv.R
```diff
@@ -180,6 +182,21 @@ virtualenv_exists <- function(envname) {
 
 virtualenv_path <- function(envname) {
 
+  # if envname is NULL, check for an active renv environment
+  # and use its virtual environment if available
+  if (is.null(envname)) {
+    renv <- Sys.getenv("RENV_PROJECT", unset = NA)
+    if (!is.na(renv)) {
+      path <- file.path(renv, "renv/r-reticulate")
```
jjallaire: We will probably want to make the `py_install()` envname also `NULL`, and then automatically create this virtualenv during py_install. There should also be an option to auto-create the environment when the renv is created.

kevinushey:
Superceded by new PR #450.

PR #450, opened 2019-03-06
implement notion of 'default' env for python actions

kevinushey:
This PR implements the notion of a 'default' Python environment, as managed through the `RETICULATE_PYTHON_ENV` environment variable. The idea is that all of the various functions `virtualenv_create()`, `py_install()`, `conda_install()` and so on would use this environment by default if `envname = NULL` (the default) and the aforementioned environment variable is set.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -76,21 +76,25 @@ conda_list <- function(conda = "auto") {
 
 #' @rdname conda-tools
 #' @export
-conda_create <- function(envname, packages = "python", conda = "auto") {
+conda_create <- function(envname = NULL, packages = "python", conda = "auto") {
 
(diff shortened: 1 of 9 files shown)


jjallaire:
Do we need the related change to `py_discover_config()` here: https://github.com/rstudio/reticulate/blob/master/R/config.R#L118 (i.e. give preference to `RETICULATE_PYTHON_ENV` when it is defined). I guess we'd want it to have preference over everything except `PYTHON_SESSION_INITIALIZED` and `RETICULATE_PYTHON`.

kevinushey:
Yes, sorry -- I've added that change now.

PR #451, opened 2019-03-11
prefer bin/conda to condabin/conda

kevinushey:
This is a fix for https://github.com/rstudio/keras/issues/691 that basically re-routes a request to use `condabin/conda` back to `bin/conda`, as the required conda utilities (e.g. `activate`) are only installed in that `bin` folder.

Note that making this change in `reticulate` does fix the issue in `keras` since it relies on `conda_binary()` to resolve the path to `conda`.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -177,6 +177,16 @@ conda_binary <- function(conda = "auto") {
     conda <- conda[[1]]
   }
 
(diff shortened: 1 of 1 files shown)


skeydan:
thanks Kevin!

PR #459, opened 2019-03-14
fix for conversion of date objects

### LINKED ISSUE #458: Error when passing Date to python
On a fresh system install, passing an `R` `Date` object to a `python` function gives the error message:

     Error in py_call_impl(callable, dots$args, dots$keywords) : 
       Cannot convert object to an environment: [type=double; target=ENVSXP].

I have installed `datetime`, `numpty` and `pandas` with `pip`. Am I missing a necessary python library? `r_to_py` works fine with `Date` objects, btw.

MRE: 

Python: `example.py`
 
    def fn(x):
      return 0

R:

    library(reticulate)
    
    module <- import(example)
    module[['fn']](Sys.Date())

kevinushey:
Closes #458.

@jjallaire I think my error here was passing along the `convert` argument during the imports of the `datetime` module, when (if I understand correctly) it's normally used for the Python capsules and especially for handling conversion of R functions. Let me know if this change looks right to you.

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -136,7 +136,7 @@ r_to_py.POSIXt <- function(x, convert = FALSE) {
   if (py_module_available("numpy"))
     return(np_array(as.numeric(x) * 1E9, dtype = "datetime64[ns]"))
 
-  datetime <- import("datetime", convert = convert)
+  datetime <- import("datetime", convert = FALSE)
   datetime$datetime$fromtimestamp(as.double(x))
 }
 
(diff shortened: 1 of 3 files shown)


R/conversion.R
```diff
@@ -154,7 +154,7 @@ py_to_r.datetime.datetime <- function(x) {
 #' @export
 r_to_py.Date <- function(x, convert = FALSE) {
 
-  datetime <- import("datetime", convert = convert)
+  datetime <- import("datetime", convert = FALSE)
```
jjallaire: Yes, that's correct. If you always want the conversion to yield an R object then you should explicitly pass `convert = FALSE`

PR #463, opened 2019-03-19
Normalizing path before py_save/load_object.

jcrodriguez1989:
reticulate's `py_save_object` and `py_load_object` functions use the `open` function from `builtin` Python lib.
For some reason, the `open` function gives an error if the file path uses '~' as a shortcut for the home dir:
``` r
library("reticulate")
py_run_string("my_object = [8,8,1,8]")

dir.exists("~/mytmp")
```

    ## [1] TRUE

``` r
unlink("~/mytmp/my_object.pkl")
file.exists("~/mytmp/my_object.pkl")
```

    ## [1] FALSE

``` r
try(py_save_object(py$my_object, "~/mytmp/my_object.pkl"))[[1]]
```

    ## [1] "Error in py_call_impl(callable, dots$args, dots$keywords) : 
    ## FileNotFoundError: [Errno 2] No such file or directory: '~/mytmp/my_object.pkl'\n"

However, if using the same path but in full path (or regular path) it works.

``` r
py_save_object(py$my_object, "/home/jcrodriguez/mytmp/my_object.pkl")
```

``` r
try(my_object <- py_load_object("~/mytmp/my_object.pkl"))[[1]]
```

    ## [1] "Error in py_call_impl(callable, dots$args, dots$keywords) : 
    ## FileNotFoundError: [Errno 2] No such file or directory: '~/mytmp/my_object.pkl'\n"

``` r
my_object <- py_load_object("/home/jcrodriguez/mytmp/my_object.pkl")
my_object
```

    ## [1] 8 8 1 8

## Fix

If in reticulate code we use `normalizePath` function then we get it work:

``` r
library("reticulate")
py_run_string("my_object = [8,8,1,8]")

dir.exists("~/mytmp")
```

    ## [1] TRUE

``` r
unlink("~/mytmp/my_object.pkl")
file.exists("~/mytmp/my_object.pkl")
```

    ## [1] FALSE

``` r
py_save_object(py$my_object, "~/mytmp/my_object.pkl")
```

``` r
my_object <- py_load_object("~/mytmp/my_object.pkl")
my_object
```

    ## [1] 8 8 1 8

### DIFF
--- a/R/pickle.R
+++ b/R/pickle.R
@@ -11,7 +11,8 @@
 py_save_object <- function(object, filename, pickle = "pickle") {
   builtins <- import_builtins()
   pickle <- import(pickle)
-  handle <- builtins$open(filename, "wb")
+  norm_filename <- normalizePath(filename, mustWork = FALSE)
+  handle <- builtins$open(norm_filename, "wb")
   on.exit(handle$close(), add = TRUE)
   pickle$dump(object, handle, protocol = pickle$HIGHEST_PROTOCOL)
 }
@@ -21,7 +22,8 @@ py_save_object <- function(object, filename, pickle = "pickle") {
 py_load_object <- function(filename, pickle = "pickle") {
   builtins <- import_builtins()
   pickle <- import(pickle)
-  handle <- builtins$open(filename, "rb")
+  norm_filename <- normalizePath(filename)
+  handle <- builtins$open(norm_filename, "rb")
   on.exit(handle$close(), add = TRUE)
   pickle$load(handle)
 }
--- a/R/pickle.R
+++ b/R/pickle.R
@@ -11,8 +11,9 @@
 py_save_object <- function(object, filename, pickle = "pickle") {
   builtins <- import_builtins()
   pickle <- import(pickle)
-  norm_filename <- normalizePath(filename, mustWork = FALSE)
-  handle <- builtins$open(norm_filename, "wb")
+  # do tilde expansion here, as Python doesnt do it
+  exp_filename <- path.expand(filename)
+  handle <- builtins$open(exp_filename, "wb")
   on.exit(handle$close(), add = TRUE)
   pickle$dump(object, handle, protocol = pickle$HIGHEST_PROTOCOL)
 }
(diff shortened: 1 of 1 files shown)


R/pickle.R
```diff
@@ -11,7 +11,8 @@
 py_save_object <- function(object, filename, pickle = "pickle") {
   builtins <- import_builtins()
   pickle <- import(pickle)
-  handle <- builtins$open(filename, "wb")
+  norm_filename <- normalizePath(filename, mustWork = FALSE)
```
jjallaire: @kevinushey Would it be better to use `path.expand` here? (not sure if there are other side effects of `normalizePath` that we'd want to avoid?)

kevinushey: I would also lean towards using `path.expand()` here as well and let any path normalization (beyond tilde expansion) that needs to happen, happen on the Python side.

jcrodriguez1989: Well, Python doesn't do tilde expansion, it has to be done manually with `os.path.expanduser`. On the other hand, R does tilde expansion, for instance, any `file.*` or `dir.*` base function.
So I thought, as py_load/save_object are R functions, it would be nice to have tilde expansion automatically done.
However, I agree with any suggestion @kevinushey might have, his knowledge of Python must be highly greater than mine 😅.

I didn't know about the `path.expand` function, `normalizePath` uses it:
```r
> path.expand
function (path) 
.Internal(path.expand(path))
<bytecode: 0x55e51aff80a0>
<environment: namespace:base>
> normalizePath
function (path, winslash = "\\", mustWork = NA) 
.Internal(normalizePath(path.expand(path), winslash, mustWork))
<bytecode: 0x55e51affb408>
<environment: namespace:base>
```
So, thanks for the knowledge! ☺️

kevinushey: The main thing I was thinking about was Windows. Python does understand the different forms of Windows paths (forward slash, backslash, short path names) and may understand UNC paths as well. If we perform normalization on the R side we risk losing some user intent or getting some subtly incompatible normalization between what R does and what Python would have done, so `path.expand()` feels a bit safer.

kevinushey: But since only R does tilde expansion by default we of course want to do it before passing any paths to Python. :-)

jcrodriguez1989: Great!
Good to learn the difference between both functions. I have already modified the pull request code.

Just out of curiosity, in the conda.R file, around line 258, there are several `path.expand` with absolute paths (hard-coded) as input, would this have any effect on other OS or architecture?

```r
      path.expand("/anaconda/bin/conda"),
      path.expand("/anaconda2/bin/conda"),
      path.expand("/anaconda3/bin/conda"),
      path.expand("/anaconda4/bin/conda"),
      path.expand("/miniconda/bin/conda"),
      path.expand("/miniconda2/bin/conda"),
      path.expand("/miniconda3/bin/conda"),
      path.expand("/miniconda4/bin/conda")

kevinushey: Those calls to `path.expand()` are effectively un-needed / superfluous; I'm guessing they're there just for symmetry with other parts of the code. They won't have any effect on the passed-in path.

kevinushey:
Thanks, and sorry for taking so long to take a look. I made a similar set of changes + documentation updates on main.

PR #476, opened 2019-04-08
don't disable convert when converting lists

kevinushey:
Fixes the issue described in https://github.com/rstudio/keras/issues/732.

I wanted to add a unit test for this behavior but for the life of me couldn't produce a a standalone example that exhibits the same behavior :/

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@
 
 Install the development version with: `devtools::install_github("rstudio/reticulate")`
 
+- Fixed an issue where Python objects within Python lists would not be
+  converted to R objects as expected.
+
 - Fixed an issue where single-row data.frames with row names could not
   be converted. (#468)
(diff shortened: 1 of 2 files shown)


jjallaire:
@kevinushey Should this be in the CRAN update as well?

@skeydan and @dfalbel, Does this fix an issue nasty enough that we should make sure that CRAN reticulate has the fix?

skeydan:
I think so, yes. It does not crop up often but when it does, it's hard to sort out what's going on, especially if it's the first time you run into it.
Thanks!

kevinushey:
I think we should take this for the CRAN update as well. I'll submit later after verifying the Conda JSON issue on all platforms.

kevinushey:
CI failures are unrelated (looks like interim issue in devtools?)

PR #479, opened 2019-04-12
open main program when libPath is empty

randy3k:
It is a non-invasive version of my previous PR #279 for handling the case that R is embedded in a Python environment.

Basically, it allows me to hijack `reticulate::py_discover_config` 

```r
py_discover_config <- funciton(...) {
    return list(
        python = "/usr/bin/python",
        libpython = "",
        ...
        ...
    )
}
```
so that reticulate would load the symbols from the main program instead. It virtually makes no changes to regular use. I hope it can get merged quicker. (PS: My last PR sat here for 9 months and I decided to close it and look for other options)

### DIFF
--- a/src/libpython.cpp
+++ b/src/libpython.cpp
@@ -57,9 +57,17 @@ bool loadLibrary(const std::string& libPath, void** ppLib, std::string* pError)
 {
   *ppLib = NULL;
 #ifdef _WIN32
-  *ppLib = (void*)::LoadLibraryEx(libPath.c_str(), NULL, 0);
+  if (libPath.empty()) {
+    *ppLib = (void*)::LoadLibraryEx(NULL, NULL, 0);
+  } else {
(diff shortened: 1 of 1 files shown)


src/libpython.cpp
```diff
@@ -57,9 +57,17 @@ bool loadLibrary(const std::string& libPath, void** ppLib, std::string* pError)
 {
   *ppLib = NULL;
 #ifdef _WIN32
-  *ppLib = (void*)::LoadLibraryEx(libPath.c_str(), NULL, 0);
+  if (libPath.empty()) {
+    *ppLib = (void*)::LoadLibraryEx(NULL, NULL, 0);
```
randy3k: Actually, I am not sure if it will work in Windows.

randy3k:
Um, it seems that
```
dlopen("", RTLD_NOW|RTLD_GLOBAL);
```
is the same as
```
dlopen(NULL, RTLD_NOW|RTLD_GLOBAL);
```

If that's really the case, the PR is actually not needed.

PR #482, opened 2019-04-17
Flowery support?

dfalbel:
In `tfdatasets` we are probably using [`flowery`](https://github.com/lionel-/flowery) as the backend to iterate over datasets (which are python iterators)  since it provides a nice syntax for R users and a lot of methods, docs, etc on how to deal with iterators.

It turns out that it's not that hard to support directly in reticulate in a backward compatible way. It would allow us to iterate over any python iterator:

``` r
library(reticulate)
builtins <- import_builtins(convert = FALSE)
range <- builtins$range(1L, 10L)

flowery::iterate(
  for(i in as_iterator(range)){
    print(i)
  }
)
#> 1
#> 2
#> 3
#> 4
#> 5
#> 6
#> 7
#> 8
#> 9
#> NULL
```

<sup>Created on 2019-04-17 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Do you think it makes sense to implement it?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,7 +34,8 @@ Imports:
     graphics,
     jsonlite,
     Rcpp (>= 0.12.7),
-    Matrix
+    Matrix,
+    flowery
 Suggests:
     testthat,
     knitr,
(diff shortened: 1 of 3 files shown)


kevinushey:
Is this still something we want to consider for reticulate?

dfalbel:
We can close it! `coro` now supports it directly: https://github.com/r-lib/coro#compatibility-with-the-reticulate-package

PR #483, opened 2019-04-18
this part is about virtualenv

cderv:
just a typo during a copy paste from above 😉

---
This change is [Reviewable](https://reviewable.io/reviews/rstudio/reticulate/483)


kevinushey:
Thanks!

PR #484, opened 2019-04-25
Adding important documentation to py_config man.

jcrodriguez1989:
I added to the `py_config` man pages that it initializes Python, i.e., `py_available()` will return `TRUE` after its execution.
In the `autokeras::install_autokeras()` function from my `autokeras` package I was testing with `py_config()` if the Python version was 3.6 (as required by the Python Auto-Keras lib), and then I was calling `keras::install_keras()` which fails if `py_available()`. I had to figure out that I should use `py_discover_config()` instead of `py_config()` by debugging. Hence, I think the adding Ive made to the man page would result useful :) .

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -2,7 +2,9 @@
 
 #' Python configuration
 #'
-#' Information on Python and Numpy versions detected
+#' Information on Python and Numpy versions detected.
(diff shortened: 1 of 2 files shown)


kevinushey:
Sorry for taking so long to review. It looks like similar changes have already landed in the development version of `reticulate`.

Issue #487, opened 2019-04-28
`caught segfault, cause 'memory not mapped'` after calling reticulate functions on NixOS


aswan89:
I'm attempting to use Reticulate (via keras/tensorflow) on NixOS and running into a segfault.

Nix has it's own way of doing things involving a centralized store of binaries that are then referenced in user environments via symlinks, meaning a lot of tools used to construct environments like virtualenv fail since the binaries are not explicitly where the tools are expected. In this case I am able to drop into a `nix-shell` as outlined in heading [9.14.1.1.3.1](https://nixos.org/nixpkgs/manual/#using-python) that has python, keras, and tensorflow available. I can successfully launch a python REPL from this shell and import/assign keras objects. 

I can also launch an R REPL and load reticulate, but attempting to use any reticulate functions causes the following segfault and traceback:

```
*** caught segfault ***
address 0x8, cause 'memory not mapped'

Traceback:
 1: py_run_string_impl(paste0("import sys; sys.path.append('", system.file("python",     package = "reticulate"), "')"))
 2: initialize_python(required_module, use_environment)
 3: ensure_python_initialized()
 4: import_builtins(convert = FALSE)
 5: repl_python()
```

R is aware of python's and reticulate's location in the nix-store:

```
> Sys.which('python')
                                                                    python 
"/nix/store/m2br92794yw09hjwh3vzadbphckagmzv-python3-3.7.3-env/bin/python" 

> paste0("import sys; sys.path.append('", system.file("python",     package = "reticulate"), "')")
[1] "import sys; sys.path.append('/nix/store/v64aywkbbr4qj92xv19fxmhdv66icl3q-r-reticulate-1.10/library/reticulate/python')"
```

This appears to be some kind of error with how reticulate is calling `py_run_string_impl`. It may be that I have more work to do with pulling in the correct dependencies in a Nix-friendly way but I don't think a segfault is the appropriate way for this to fail.

deliciouslytyped:
I was trying to help when this got posted, I got as far as getting gdb to tell me the crash is in `0x00007fffdae5b46b in PyModule_GetDict () from /nix/store/0n8slcq8p5x31kc9hncabsqq9y3fpkzp-python3-3.7.3/lib/libpython3.7m.so `, but that's as far as I got. I think that's the following line: https://github.com/rstudio/reticulate/blob/7be2d8462801c271f12577dd65334ed029cf0053/src/python.cpp#L2055

ecoughlan:
[...]

PR #497, opened 2019-05-05
Fix crash on failed numpy import

ecoughlan:
This fixes the crash in #487 when numpy is not in the search path. This is a new issue after upgrading to python 3.7, so there might be other occurrences (although I didn't see anything specific in the changelog).


kevinushey:
Thanks!

Issue #498, opened 2019-05-11
py_config() unable to detect system-installed python in R-3.6.0 on Windows 10


yangchino1:
and gives the error message
> Error in mapply(function(from, to) paste(lines[from:to], collapse = "\n"),  : 
  zero-length inputs cannot be mixed with those of non-zero length

`py_available` and `use_python` are unable either, although `Sys.which("python")` can find the python executable. However, they all work for  R-3.5.3 on the same machine. I guess this is due to changes in R-3.6.0. I wonder which would fix this first, R for Windows or reticulate?

viebrix:
I have the same issue with 
Windows 7, R 3.6, Anaconda and RStudio/keras,
see also: https://github.com/rstudio/keras/issues/776

viebrix:
Problem seems to be in function python_config following line:
`config <- system2(command = python, args = paste0('"', config_script, '"'), stdout = TRUE)`
This command returns in R3.6 nothing (ends in length=0)
if you change to:
`config <- system2(command = paste0('',python,''), args = paste0('"', config_script, '"'), stdout = TRUE, stderr=T)`
Information about the config will be returned. 
I don't know why the answer from 'python.exe /reticulate/config/config.py" is a stderr and not a stdout, but maybe the contributor of the package can follow this hint?

Answer is if I use stderr=T following information:
```
[1] "Version: 3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)]"                   "VersionNumber: 3.7"                                                                           
[3] "PREFIX: C:\\Users\\xxxxx~1\\AppData\\Local\\CONTIN~1\\ANACON~1"                               "EXEC_PREFIX: C:\\Users\\xxxxx~1\\AppData\\Local\\CONTIN~1\\ANACON~1"                         
[5] "Architecture: 64bit"                                                                           "NumpyPath: C:\\Users\\xxxxx~1\\AppData\\Local\\CONTIN~1\\ANACON~1\\lib\\site-packages\\numpy"
[7] "NumpyVersion: 1.16.3"                                                                          "RequiredModule: tensorflow" 
```

viebrix:
@yangchino1 : You can try it with devtools::install_github("viebrix/reticulate")

viebrix:
[...]

PR #505, opened 2019-05-21
R3.6 Windows Problems with reticulate

viebrix:
It seems that Windows in R3.6 using system2 call, does not return a stdout from an python call. It seems to me that it only returns data as stderr. Therefore py_config() does not return the configuration from the python file. It provides configuration detail only in stderr and not in stdout. This leeds the an mapply error, because the config is only of length=0. 

See also issue: https://github.com/rstudio/reticulate/issues/498

Additional python_version and python_module_version in python-tools.R use also the system2 command. 
I have changed those calls in my fork and it seems now it's working on my PC, maybe someone can check if this solution is a possible solution for the whole package or if there is a bug which should be corrected in R 3.6

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -354,7 +354,7 @@ python_config <- function(python, required_module, python_versions, forced = NUL
 
   config_script <- system.file("config/config.py", package = "reticulate")
   # It seems that Windows in R3.6 returns the configuration from the python file only in stderr and not in stdout
-  config <- system2(command = python, args = paste0('"', config_script, '"'), stdout = TRUE, stderr = TRUE)
+  config <- system2(command = python, args = paste0('"', config_script, '"'), stdout = TRUE, stderr = FALSE)
   status <- attr(config, "status")
   if (!is.null(status)) {
     errmsg <- attr(config, "errmsg")
--- a/R/config.R
+++ b/R/config.R
@@ -353,7 +353,8 @@ python_config <- function(python, required_module, python_versions, forced = NUL
   }
 
   config_script <- system.file("config/config.py", package = "reticulate")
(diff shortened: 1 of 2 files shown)


kevinushey:
Strange! It works fine for me in RStudio, but not within RGui. It also works fine in an R session launched from the terminal.

Can you confirm whether that's the case for you as well?

kevinushey:
Looks like this is a bug in R 3.6.0; I've filed it here: https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17568

I think if you explicitly set `stderr = FALSE` that would work around the issue, but I think I would prefer if we scoped this workaround to R 3.6.0.

viebrix:
Hi Kevin,
thanks for investigating this issue!
I can confirm that Rscript.exe is reacting differently than RGUI (my RStudio version is currently not up to date, therefore I haven't made tests with it).

I adapted your example from bugzilla R:
test_system2.R:
```
print("test1")
print(system2("cmd.exe", "/c echo hi", stdout = TRUE))
print("test2")
print(system2("cmd.exe", "/c echo hi", stdout = TRUE, stderr=F))
```

calling in RGUI 3.6:
```
> source("test_system2.R")
[1] "test1"
character(0)
[1] "test2"
[1] "hi"
```
calling with Rscript:
```
C:\Program Files\R\R-3.6.0\bin\x64>Rscript test_system2.R
[1] "test1"
[1] "hi"
[1] "test2"
[1] "hi"
```

I understand, that you want to resolve a problem directly where it occurs, in this case directly in RGUI.
For me it's ok because I currently can work with my fork, until RGUI will be corrected.

kevinushey:
Thanks for the PR!

PR #525, opened 2019-06-17
recursively convert tuples as we do lists

skeydan:
Before modifying `test.py` I tried this R-only test 

```
 builtins <- import_builtins(convert = FALSE)
 collections <- import("collections", convert = FALSE)
 t <- builtins$tuple(list(22, collections$OrderedDict(list(list(777, "b"), list(1, "a")))))
 py_to_r(t)
 ```

but this didn't convert the ``OrderedDict``:

```
[1]]
[1] 22

[[2]]
OrderedDict([(777.0, 'b'), (1.0, 'a')])

```

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -74,6 +74,9 @@ py_to_r.python.builtin.list <- function(x) {
   converted
 }
 
+#' @export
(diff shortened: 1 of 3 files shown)


kevinushey:
LGTM; thanks!

PR #541, opened 2019-07-04
fixes conda environment being ignored if numpy not present

hchandola:
Passed environment in current `delay_load` code fails if `numpy` is not present in the passed environment. This fix ignores the check for numpy version if numpy is not installed in the environment.

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -207,7 +207,8 @@ py_discover_config <- function(required_module = NULL, use_environment = NULL) {
     if (has_compatible_arch && (is.null(config$numpy) || has_preferred_numpy))
       valid_python_versions <- c(valid_python_versions, python_version)
     has_required_module <- is.null(config$required_module) || !is.null(config$required_module_path)
-    if (has_python_gte_27 && has_compatible_arch && has_preferred_numpy && has_required_module)
+    if (has_python_gte_27 && has_compatible_arch && (is.null(config$numpy) || has_preferred_numpy) &&
+      has_required_module)
       return(config)
   }
 
--- a/R/config.R
+++ b/R/config.R
@@ -204,7 +204,7 @@ py_discover_config <- function(required_module = NULL, use_environment = NULL) {
     has_python_gte_27 <- as.numeric_version(config$version) >= "2.7"
     has_compatible_arch <- !is_incompatible_arch(config)
     has_preferred_numpy <- !is.null(config$numpy) && config$numpy$version >= "1.6"
-    if (has_compatible_arch && has_preferred_numpy)
+    if (has_compatible_arch && (is.null(config$numpy) || has_preferred_numpy))
       valid_python_versions <- c(valid_python_versions, python_version)
     has_required_module <- is.null(config$required_module) || !is.null(config$required_module_path)
     if (has_python_gte_27 && has_compatible_arch && has_preferred_numpy && has_required_module)


jjallaire:
One problem here is that w/o NumPy we can't convert R Matrices to Python. While it may be true that use of your particular package doesn't require Matrices, this is setting the Python version for the entirety of the session so any other package that does need matrices will then fail to work.

The pattern of using a special environment for a package (like we do in tensorflow) is actually somewhat of an anti-pattern, but was necessitated by the fact that TF often needed to be completely removed (as in, it's environment obliterated) before upgrading. It's much better if all R packages can share a single Python environment (e.g. the reticulate default "r-reticulate"). In this case all the R packages can work in the same R session vs. require the user to create different sessions for different combinations of Python libraries they need to use. We are in fact right now changing tensorflow to use the default "r-reticulate" environment (for the above reasons) and crossing our fingers that the upgrading problems have gotten better on the TF side.

So I'd recommend having your installation functions just use the default `py_install()` function w/ no special environment arguments. This will have the added bonus of making them play well with the forthcoming *renv* package (https://rstudio.github.io/renv/articles/python.html).

hchandola:
The problem with current setup is it makes no guarantees of loading from
the user’s environment. I think we should rather give warning for missing
numpy (mesg: the user won’t be able to use matrixes) or error in case of
wrong architecture but not load a different python than what the user asked
for.

The current method of loading from any other environment than what the user
asked for is wrong. We should not treat an argument like a suggestion.

Forcing all packages to share an environment will run into problems when
there are dependency conflicts between packages.

I will vote that we put the PR in with added warnings and errors. Please
let me know your thoughts.

jjallaire:
> Forcing all packages to share an environment will run into problems when
there are dependency conflicts between packages.

That's actually how R works (there is a single R session and dependency conflicts between packages cause problems, i.e. you can't have multiple versions of a package loaded in the same session). The solution to this is Python virtual environments (or in the case of R, renv as noted before). If you force an envname in your .onLoad you will foil the user's ability to use renv/virtualenv to create separate dependency trees for different projects on their system).

I agree that it's not natural for arguments to be advisory. I think what we should have done in the first place is make NumPy a requirement for reticulate (that's actually how the package worked in the beginning). I think the solution on your end is to just add NumPy to the packages installed by your install function.

@kevinushey LMK if you have a different viewpoint here.

hchandola:
With respect to what exists today, I would say either giving a warning or bailing out on error if `numpy` is missing is a must (instead of loading whatever first environment that works).

jjallaire:
Okay, I agree that a warning indicating that the requested environment was
skipped due to no NumPy is absolutely necessary (so there is no mystery
about why the argument was bypassed). Error might break some existing
configurations (although I do regret it wasn't an error in the first
place). Kevin let us know what you think.

kevinushey:
I agree that we should warn if the version of Python actually loaded does not match the version that was requested by the user. I think this is the largest source of confusion for users of reticulate -- finding out that the version of Python requested by `use_python()` was ignored, but not knowing why. (And it seems most new users are unaware of the `required` argument; ie, they expect it to be mandatory rather than advisory)

Perhaps sometime we could prototype a change where we flip the default `required = FALSE` to `required = TRUE` and see how much fallout that causes.

jjallaire:
The issue w/ required = TRUE is that I originally imagined that people
would be attempting to run the same reticulated R script on multiple
systems that might have slightly different Python installations. Thus the
notion that it's an advisory hint (look here but if it's not there proceed
to the next hint). I think if we flipped it to TRUE we would break a
non-trivial number of existing scripts (but I could be wrong).

kevinushey:
That's a good point, and that could be true. I think then a warning rather than an error is the best approach so at least users can more easily see that reticulate didn't load the requested version of Python.

I think some of the tooling we're implementing in `renv` will help here. When you activate Python with a project, we track the Python version (e.g. 3.7 vs. 2.7) and the type of environment (virtualenv vs. conda); if you try to restore an renv project on a separate machine, we'll then try to also restore the same environment with the same version of Python (searching in the PATH + other common install locations) + environment type.

kevinushey:
Do we still want to take this PR now that we're explicitly warning when a different environment than the one requested by the user is loaded?

hchandola:
I was planning to discard the PR since from the thread sounded like we
would be giving a warning. Sounds good?

kevinushey:
Sounds good! Thanks for the quick response.

PR #544, opened 2019-07-09
revamp py_install()

### LINKED ISSUE #543: revamp py_install()
- Handle venv (with Python 3) versus virtualenv (with Python 2)
- Update `python_unix_binary()` to select Python on the PATH and use it in `py_install()`
- Prefer Python 3 in `py_install()` (but allow opt out in some way)

kevinushey:
Closes #543.

This PR makes the following changes:

- When type is "auto", we try to detect and infer the environment type directly if an envname was supplied.

- `python_unix_binary()` is removed, and we now instead query `RETICULATE_PYTHON` as well as the PATH for the version of Python to use when creating a new virtual environment. This is done in the `virtualenv_default_python()` function. (note that this detection is not required for Conda environments)

Effectively, this takes some of the logic that was inlined in `py_install()` and delegates back to the already-existing `virtualenv_` and `condaenv_` functions for accomplishing the same.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -1,14 +1,57 @@
 
+py_install_method_detect <- function(envname, conda = "auto") {
 
+  # on Windows, we always use conda environments
+  if (is_windows()) {
+
+    # validate that conda is available
+    conda <- tryCatch(conda_binary(conda = conda), error = identity)
+    if (inherits(conda, "error")) {
+      msg <- paste(
+        "Conda installation not found (failed to locate conda binary)",
+        "Please install Anaconda 3.x for Windows (https://www.anaconda.com/download/#windows) before proceeding.",
+        sep = "\n"
+      )
+      stop(msg, call. = FALSE)
+    }
+
+    # return it
+    return("conda")
+
+  }
+
+  # try to find an existing virtualenv
+  if (virtualenv_exists(envname))
+    return("virtualenv")
+
+  # try to find an existing condaenv
+  if (condaenv_exists(envname, conda = conda))
+    return("conda")
+
+  # check to see if virtualenv or venv is available
+  python <- virtualenv_default_python()
+  if (python_has_module(python, "virtualenv") || python_has_module(python, "venv"))
+    return("virtualenv")
+
+  # check to see if conda is available
+  conda <- tryCatch(conda_binary(conda = conda), error = identity)
+  if (!inherits(conda, "error"))
+    return("conda")
+
+  # default to virtualenv
+  "virtualenv"
+
+}
 
 #' Install Python packages
 #'
-#' Install Python packages into a virtualenv or conda env.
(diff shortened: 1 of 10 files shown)


R/install.R
```diff
@@ -1,14 +1,57 @@
...
+
+  # on Windows, we always use conda environments
+  if (is_windows()) {
+
+    # validate that conda is available
+    conda <- tryCatch(conda_binary(conda = conda), error = identity)
+    if (inherits(conda, "error")) {
+      msg <- paste(
+        "Conda installation not found (failed to locate conda binary)",
+        "Please install Anaconda 3.x for Windows (https://www.anaconda.com/download/#windows) before proceeding.",
```
jjallaire: Maybe get rid of the 3.x qualifier here?

R/install.R
```diff
@@ -23,145 +66,31 @@
...
-      have_virtualenv <- virtualenv_available && !is.null(virtualenv)
-
-      # if we don't have pip and virtualenv then try for conda if it's allowed
-      if ((!have_pip || !have_virtualenv) && have_conda) {
-
-        conda_install(envname, packages = packages, conda = conda, ...)
-
-      # otherwise this is either an "auto" installation w/o working conda
-      # or it's an explicit "virtualenv" installation
-      } else {
```
jjallaire: My biggest worry about this change is that we are losing the printing of bootstrapping commands at the console. But perhaps these aren't necessary as we can assume they have pip / venv in most cases? In would be interesting to see what happens on a brand new OS X machine -- can the user get a working Python environment w/o help from us?

kevinushey: Apple is actually planning to deprecate Python in the next release of macOS (https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_beta_3_release_notes) so we might want to consider suggesting to the user that they should install their own copy of Python.

I've had good luck with Homebrew Python and I think it's in general the way most people install Python + other software on macOS nowadays, so perhaps we should consider recommending that?

kevinushey: I'll see if I can bring back installation instructions as well though (I agree they're useful to have for users)

jjallaire:
Or should we recommend conda rather than homebrew? (As homebrew requires
the installation of not just Python but also homebrew, and we already
require conda on Windows)

R/install.R
kevinushey: It looks like we do still provide recommendations on how pip / virtualenv / venv can be installed, but we defer it to environment creation time now:

https://github.com/rstudio/reticulate/blob/feature/py-install-revamp/R/virtualenv.R#L273-L295

jjallaire:
Ok, that's great that we have that code. It seems like we should just make
sure that there is a way for these messages to be triggered / printed
during py_install should they be needed (i.e. in the case of an early exit
due to not finding the requisite components)

PR #545, opened 2019-07-11
warn if request for specific Python not honored

kevinushey:
This PR checks whether the user has requested a specific version of Python, and warns if the version of Python that was actually loaded differs from their requested version.

cc: @jjallaire

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -128,19 +128,28 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   # as the requested version of python
   local({
 
+    # nothing to do if user didn't request any version
     requested_versions <- reticulate_python_versions()
     if (length(requested_versions) == 0)
-      return(TRUE)
+      return()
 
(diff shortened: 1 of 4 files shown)


R/package.R
```diff
@@ -124,6 +124,29 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
   # ensure modules can be imported from the current working directory
   py_run_string_impl("import sys; sys.path.insert(0, '')")
 
+  # notify the user if the loaded version of Python isn't the same
+  # as the requested version of python
+  local({
+
+    requested_versions <- reticulate_python_versions()
+    if (length(requested_versions) == 0)
+      return(TRUE)
```
jjallaire: Are these just early returns? (i.e. does the `TRUE` do anything here?)

kevinushey: Just early returns (no meaning for return value)

jjallaire:
Okay, LGTM.

PR #546, opened 2019-07-13
allow reticulate to bind to debug Python builds

kevinushey:
This PR makes it possible for reticulate to bind to debug Python builds.

The main challenge here is that the PyObject struct layout (whose layout is needed for the PyModuleDef struct) is different when debugging is enabled. Because of that, we need to include two types of definitions, and choose the appropriate one at runtime.

This also forces us to change a couple macros which tried to use the PyObject struct directly (we instead call back into libpython whenever possible to ensure we do the right thing)

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -392,7 +392,7 @@ python_config <- function(python, required_module, python_versions, forced = NUL
     python_libdir_config <- function(var) {
       python_libdir <- config[[var]]
       ext <- switch(Sys.info()[["sysname"]], Darwin = ".dylib", Windows = ".dll", ".so")
-      pattern <- paste0("^libpython", version, "m?", ext)
+      pattern <- paste0("^libpython", version, "d?m?", ext)
       libpython <- list.files(python_libdir, pattern = pattern, full.names = TRUE)
     }
     for (libsrc in c("LIBPL", "LIBDIR")) {
--- a/src/libpython-types.h
+++ b/src/libpython-types.h
@@ -1,0 +1,101 @@
+
+#ifndef LIBPYTHON_TYPES_H
+#define LIBPYTHON_TYPES_H
+
+namespace libpython {
+
+#if _WIN32 || _WIN64
+#if _WIN64
+typedef __int64 Py_ssize_t;
+#else
+typedef int Py_ssize_t;
+#endif
+#else
+typedef long Py_ssize_t;
+#endif
+
+#define METH_VARARGS  0x0001
+#define METH_KEYWORDS 0x0002
+
+#define Py_file_input 257
+#define Py_eval_input 258
+
+
+#ifdef LIBPYTHON_DEBUG
+#define _PyObject_HEAD_EXTRA struct _object *_ob_next; struct _object *_ob_prev;
+#define _PyObject_EXTRA_INIT 0, 0,
+#else
+#define _PyObject_HEAD_EXTRA
+#define _PyObject_EXTRA_INIT
+#endif
+
+
+#define PyObject_HEAD   \
+_PyObject_HEAD_EXTRA    \
+  Py_ssize_t ob_refcnt; \
+struct _typeobject *ob_type;
+
+#define PyObject_VAR_HEAD               \
+PyObject_HEAD                           \
+  Py_ssize_t ob_size;
+
+typedef struct _typeobject {
+  PyObject_VAR_HEAD
+  const char *tp_name;
+  Py_ssize_t tp_basicsize, tp_itemsize;
+} PyTypeObject;
+
+typedef struct _object {
+  PyObject_HEAD
+} PyObject;
+
+typedef PyObject *(*PyCFunction)(PyObject *, PyObject *);
+
+struct PyMethodDef {
+  const char	*ml_name;
+  PyCFunction  ml_meth;
+  int		 ml_flags;
+  const char	*ml_doc;
+};
+typedef struct PyMethodDef PyMethodDef;
+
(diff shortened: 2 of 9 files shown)


jjallaire:
This looks great! So happy you were able to figure all of this out :-)

I would however say that I'd hold off merging this until after the upcoming release (as it has enough changes where I wouldn't want to jam it in at the last minute).

skeydan:
Hi @kevinushey, thanks so much, that's awesome!! I'll try it out on Monday.
(Evidently I won't need to symlink the executable anymore either :-))

skeydan:
Just tested with my debug version of Python. I get an assertion error early in the code (before the thing I actually would need to execute), to be reproduced by

```
reticulate::use_python("/home/key/python/python3.7/tf/bin/python", required = TRUE)
library(tensorflow)
x <- tf$cast(1, tf$float32)
```

```
R: Objects/typeobject.c:3670: excess_args: Assertion `PyTuple_Check(args)' failed.
```

Unfortunately, there seems to be no way where Python executes the assertions and prints them, but does _not_ stop... so probably I'll need to think of something else for my problem, but still, I think this is very useful to have!

skeydan:
Just noting that with a debug python, after sourcing python-gdb.py, one can actually nicely inspect the python frames for the above example:

```
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	  return ret;
[Current thread is 1 (Thread 0x7f01bb691300 (LWP 19136))]
Missing separate debuginfos, use: dnf debuginfo-install hdf5-1.8.20-6.fc30.x86_64 libaec-1.0.4-1.fc30.x86_64 openblas-Rblas-0.3.6-2.fc30.x86_64 sssd-client-2.2.0-3.fc30.x86_64
(gdb) py-bt
Undefined command: "py-bt".  Try "help".
(gdb) source /home/key/code/cpython/python-gdb.py
(gdb) py-bt
Traceback (most recent call first):
  <built-in method TF_NewBuffer of module object at remote 0x7f01a68a0690>
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/c_api_util.py", line 169, in tf_buffer
    buf = c_api.TF_NewBuffer()
  <built-in method next of module object at remote 0x7f01a8b013c0>
  File "/home/key/python/python3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 3736, in _get_op_def
    with c_api_util.tf_buffer() as buf:
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 1710, in __init__
    op_def = self._graph._get_op_def(node_def.op)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 3348, in _create_op_internal
    op_def=op_def)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 3279, in create_op
    attrs, op_def, compute_device)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/util/deprecation.py", line 507, in new_func
    return func(*args, **kwargs)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/constant_op.py", line 290, in _constant_impl
    name=name).outputs[0]
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/constant_op.py", line 246, in constant
    allow_broadcast=True)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/constant_op.py", line 305, in _constant_tensor_conversion_function
    return constant(v, dtype=dtype, name=name)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 1239, in internal_convert_to_tensor
    ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 1160, in convert_to_tensor_v2
    as_ref=False)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 1102, in convert_to_tensor
    return convert_to_tensor_v2(value, dtype, preferred_dtype, name)
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/ops/math_ops.py", line 693, in cast
    x = ops.convert_to_tensor(x, name="x")
  File "/home/key/python/python3.7/tf/lib/python3.7/site-packages/tensorflow_core/python/util/dispatch.py", line 180, in wrapper
    return target(*args, **kwargs)

```

kevinushey:
One thing I just realized over the weekend is that this PR won't handle the numpy + interpreter-specific functions we have, and handling that properly will be a little more tricky.

I think I might just take the following approach instead:

1. Allow a compile flag (e.g. `-DRETICULATE_DEBUG`) to set reticulate to use the 'debug' structs. In that case reticulate will only be able to work with debug versions of Python, but since this is for advanced usages for now I think that's fine.

2. Sometime later, see if we can get things to "just work" transparently regardless of whether it's a debug / release version of Python. This is still possible, but just trickier.

PR #548, opened 2019-07-15
allow reticulate to be compiled to use debug Python

kevinushey:
This is a simpler version of #546, which simply allows the user to choose between debug / release versions of Python at compile time for reticulate. The user can define `RETICULATE_PYTHON_DEBUG` to declare that they're building a version of reticulate that should work with debug versions of Python.

This is less useful than #546, but safer, and since this is already a feature intended for 'expert' users it's probably okay to hide it behind a pre-processor define (ie I imagine we'd only use this internally when attempting to debug crashes / segfaults)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,10 @@
 
 Install the development version of reticulate with `remotes::install_github("rstudio/reticulate")`.
 
+- It is now possible to compile `reticulate` with support for debug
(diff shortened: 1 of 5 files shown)


jjallaire:
I like this change much better and fully agree that have the user build reticulate from source to enable this is the right way to go.

PR #549, opened 2019-07-16
accept 'python_version' in py_install()

kevinushey:
This PR allows users of `py_install()` and `conda_install()` to request a specific Python version, to be installed in the associated Conda environment. (it currently ignored for virtual environments)

This should allow e.g. `install_tensorflow()` to work with Conda environments with a specific Python version requirement.

### DIFF
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -15,13 +15,17 @@
 #'   environment within `virtualenv_root()`. When `NULL`, the virtual environment
 #'   as specified by the `RETICULATE_PYTHON_ENV` environment variable will be
 #'   used instead.
+#'
 #' @param packages A character vector with package names to install or remove.
+#'
 #' @param ignore_installed Boolean; ignore previously-installed versions of the
 #'   requested packages? (This should normally be `TRUE`, so that pre-installed
 #'   packages available in the site libraries are ignored and hence packages
 #'   are installed into the virtual environment.)
+#'
 #' @param confirm Boolean; confirm before removing packages or virtual
 #'   environments?
+#'
 #' @param python The path to a Python interpreter, to be used with the created
 #'   virtual environment. When `NULL`, the Python interpreter associated with
 #'   the current session will be used.
@@ -45,6 +49,7 @@ virtualenv_list <- function() {
 #' @rdname virtualenv-tools
 #' @export
 virtualenv_create <- function(envname = NULL, python = NULL) {
+
   path <- virtualenv_path(envname)
   name <- if (is.null(envname)) path else envname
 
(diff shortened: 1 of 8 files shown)


R/virtualenv.R
```diff
@@ -69,6 +76,11 @@ virtualenv_create <- function(envname = NULL, python = NULL) {
     stop(msg, call. = FALSE)
   }
 
+  # upgrade pip and friends after creating the environment
```
jjallaire: This I used to do in tensorflow but found that it caused problems when pip upgrades had breaking issues (e.g. upgrade to pip 10 was really rocky). I ended up changing the logic to only upgrade pip/wheel/setuptools when they were "too old" (don't recall what the threshold was, maybe v8?). 

Anyway, there definitely could be a good reason you are doing this, but I wanted to alert you to the fact that it has caused problems in the past. LMK what you think.

kevinushey: My understanding is that virtualenv / venv actually contain their own bundled versions of pip (different from what may be installed by the user in their Python library), and since those versions can become stale, one normally would want to upgrade them after creating a virtual environment.

The other issue here is that pip is pretty chatty when it's out of date (it tells you to upgrade every time you try and use it) so I thought it might be prudent to ensure it's updated right away.

That said, these are all just minor annoyances compared to "updated pips can sometimes be broken" so it might be better to just be conservative so maybe we should stick with the old behavior.

jjallaire: Well, I had a very bad experience with pip 10 but perhaps that is well behind us? If you understand this to be the best practice I think it's fine. Do the web-based docs for virtualenv / venv include/recommend this or not?

kevinushey: I think `venv` uses an old bundled version of `pip` just to ensure that it can work even without internet access, but the most common practice is to update pip immediately after creating the virtual environment. I don't see this practice officially blessed, but it is acknowledged in e.g. https://bugs.python.org/issue30628:

> It *might* make sense to offer a `--upgrade-pip` flag for `venv` creation, but it's not entirely clear how much benefit that really offers over just running `python -m pip install --upgrade pip` inside the venv yourself. That one's probably worth filing as a separate RFE though, as it's probably the single most common post-creation script for new virtual environments.

jjallaire:
Okay, I'm comfortable with this then. Does it make sense to treate venv and
virtualenv (i.e. Python 2) differently here? It seems to me I recall that
in virtualenv updating pip somehow affected the entire Python installation
of pip which in turn broke things. Perhaps venv is better isolated?

kevinushey:
> Okay, I'm comfortable with this then. Does it make sense to treate venv and
virtualenv (i.e. Python 2) differently here? It seems to me I recall that
in virtualenv updating pip somehow affected the entire Python installation
of pip which in turn broke things. Perhaps venv is better isolated?

I think this should be fine, but let me test to confirm.

Issue #551, opened 2019-07-17
segfault when converting OrderedDict object


kevinushey:
To reproduce:

```r
library(reticulate)

py_run_string('
from collections import OrderedDict
d = OrderedDict()
d[1] = 1
')

main <- import_main(convert = FALSE)
repeat r_to_py(py_to_r(main$d))
```

kevinushey:
Looks like `iterate()` is the culprit:

```
library(reticulate)

py_run_string('x = [1, 2, 3]')
main <- import_main(convert = FALSE)
x <- main$x

repeat iterate(x)
```

PR #552, opened 2019-07-17
avoid use of iterate in conversion of dict

kevinushey:
Part of https://github.com/rstudio/reticulate/issues/551.

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -188,17 +188,12 @@ py_to_r.collections.OrderedDict <- function(x) {
   disable_conversion_scope(x)
 
   keys <- py_dict_get_keys(x)
-
-  i <- 1
-  result <- vector("list", length(keys))
-  iterate(keys, function(key) {
-    result[[i]] <<- py_to_r(py_dict_get_item(x, key))
-    i <<- i + 1
+  result <- lapply(seq_len(length(keys)) - 1L, function(i) {
+    py_to_r(py_dict_get_item(x, keys[i]))
   })
 
   names(result) <- py_dict_get_keys_as_str(x)
   result
-
 }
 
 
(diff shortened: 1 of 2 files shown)


kevinushey:
Phew! Found the root cause -- we were accidentally decreasing the reference count on something that we intended to 'own'.

This should hopefully fix a whole class of segfaults that we might have been seeing the wild (ie anything built on top of `iterate()`).

kevinushey:
@jjallaire can you sanity-check this diff? https://github.com/rstudio/reticulate/pull/552/commits/e0597983a41ad97c1e8607d0b1d043e96d52d3d7

I believe that if we want to construct a `py_ref()` here then we need to detach the PyObjectPtr; does that sound right?

jjallaire:
Will want to take a thorough look at this (can probably do so tomorrow but if it's another day pls hold merging until I do)

skeydan:
Wow, thank you so much Kevin!!  :-) :-) :-)

I'm of course no judge as to the fix, _but_, I haven't seen this (triggering 3 builds in a row ;-) for quite some time now:

image

jjallaire:
It does indeed look like the semantics of `py_ref()` are that the object needs to have an active reference count on it that won't be freed by other means. In this case, that means the `PyObjectPtr` needs to forget about it so it fulfills this expectation (this `detach()`). In short, yes, this LGTM!

langmm:
This is great, thank you!

langmm:
One question, will these changes be pushed to CRAN/conda-forge and, if so, when will this occur? I'm a bit unfamiliar with the process on CRAN but noticed the version was published on 4/12/19, but conda-forge recipe was updated yesterday (7/17/19).

kevinushey:
We're hoping to submit to CRAN soon; likely next week.

Issue #522, opened 2019-06-13
conda_install(pip = TRUE) does not work with Conda >= 4.4


hriebl:
Hello everybody,

with Conda 4.4, the command to activate an environment has changed, see https://github.com/conda/conda/releases/tag/4.4.0. Unfortunately, `conda_install(pip = TRUE)` still uses the old command and does not work anymore.

For me, the new command (`conda activate ENVNAME`) also fails when called via R's `system()` function, while it works from the normal command line. The issue probably has to do with some configuration files not being read properly.

I am not sure what would be the best, most platform independent way to solve this issue, but for me calling pip via Python works, see below.

Here is the whole story in code:

```
> library(reticulate)
> conda_create("test")

*snip*

# To activate this environment, use
#
#     $ conda activate test
#
# To deactivate an active environment, use
#
#     $ conda deactivate

[1] "/home/hannes/.conda/envs/test/bin/python"
> conda_install("test", packages = "networkx", pip = TRUE)
/bin/bash: /usr/bin/activate: No such file or directory
Error: Error 1 occurred installing packages into conda environment test
> system("conda activate test")

CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
If your shell is Bash or a Bourne variant, enable conda for the current user with

    $ echo ". /usr/etc/profile.d/conda.sh" >> ~/.bashrc

or, for all users, enable conda with

    $ sudo ln -s /usr/etc/profile.d/conda.sh /etc/profile.d/conda.sh

The options above will permanently enable the 'conda' command, but they do NOT
put conda's base (root) environment on PATH.  To do so, run

    $ conda activate

in your terminal, or to put the base environment on PATH permanently, run

    $ echo "conda activate" >> ~/.bashrc

Previous to conda 4.4, the recommended way to activate conda was to modify PATH in
your ~/.bashrc file.  You should manually remove the line that looks like

    export PATH="/usr/bin:$PATH"

^^^ The above line should NO LONGER be in your ~/.bashrc file! ^^^


> system("bash -c 'conda activate test'")

*same error, snip*

Restarting R session...

> library(reticulate)
> use_condaenv("test", required = TRUE)
> py_run_string("import subprocess, sys
+ subprocess.call([sys.executable, '-m', 'pip', 'install', 'networkx'])")
Collecting networkx
Collecting decorator>=4.3.0 (from networkx)
[...]

PR #567, opened 2019-08-02
Switch to conda activate command

hriebl:
Based on #522, here is a pull request for the switch from the `source activate` to the `conda activate` command. Should work with Conda >= 4.6. Tested on Fedora 29 and Windows 10. Unfortunately, I do not have a macOS installation for testing.

Not sure if you want to merge this, but feel free to take a look and propose some changes.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -182,14 +182,13 @@ conda_install <- function(envname = NULL,
 
   if (pip) {
     # use pip package manager
-    condaenv_bin <- function(bin) path.expand(file.path(dirname(conda), bin))
-    cmd <- sprintf("%s%s %s && pip install --upgrade %s %s%s",
-                   ifelse(is_windows(), "", ifelse(is_osx(), "source ", "/bin/bash -c \"source ")),
-                   shQuote(path.expand(condaenv_bin("activate"))),
+    cmd <- sprintf("%s %s && pip install --upgrade %s %s%s",
+                   ifelse(is_windows(), paste(conda, "activate"),
+                          sprintf("bash -c \"eval $(%s shell.bash hook) && conda activate", conda)),
                    envname,
                    ifelse(pip_ignore_installed, "--ignore-installed", ""),
                    paste(shQuote(packages), collapse = " "),
-                   ifelse(is_windows(), "", ifelse(is_osx(), "", "\"")))
+                   ifelse(is_windows(), "", "\""))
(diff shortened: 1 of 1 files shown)


kevinushey:
Is `conda activate` backwards compatible with older versions of Conda? I wonder if we need to change the command invocation based on the version of Conda in use.

R/conda.R
```diff
@@ -182,14 +182,13 @@ conda_install <- function(envname = NULL,
 
   if (pip) {
     # use pip package manager
-    condaenv_bin <- function(bin) path.expand(file.path(dirname(conda), bin))
-    cmd <- sprintf("%s%s %s && pip install --upgrade %s %s%s",
-                   ifelse(is_windows(), "", ifelse(is_osx(), "source ", "/bin/bash -c \"source ")),
-                   shQuote(path.expand(condaenv_bin("activate"))),
+    cmd <- sprintf("%s %s && pip install --upgrade %s %s%s",
+                   ifelse(is_windows(), paste(conda, "activate"),
+                          sprintf("bash -c \"eval $(%s shell.bash hook) && conda activate", conda)),
```
kevinushey: For my own context, where is the `eval $(%s shell.bash hook)` coming from?

hriebl: This is because in recent versions of Conda the `conda` command is no longer an executable file but a function in your shell. You can check this by running `which conda`. That's okay as long as you are using your shell interactively, but functions are not exported to non-interactive subshells. The `conda shell.bash hook` command makes the function available in the subshell. I took it from this post: https://github.com/conda/conda/issues/7980#issuecomment-492784093.

hriebl:
> Is `conda activate` backwards compatible with older versions of Conda? I wonder if we need to change the command invocation based on the version of Conda in use.

No, this is not backwards compatible, it will only work with Conda versions >= 4.6. Honestly, for the moment I would recommend you to keep everything as it is. The old `activate` command still seems to work with most™ Conda distributions. And my impression is that there are no clean solutions for this issue. The Conda shell scripts seem to be a bit of a mess...

Anyway, in case the Conda developers decide to completely remove the `activate` command at some point in the future, you have an alternative solution at hand. :slightly_smiling_face:

kevinushey:
Sorry for taking so long to return to this -- I think this PR is now obviated by other recent changes in reticulate. If you're still seeing that this is necessary, please let me know.

PR #573, opened 2019-08-07
Autocomplete functions: Get formals from Python function signatures!

### LINKED ISSUE #553: Preserve python function signature for autocompletion
`py_callable_as_function` converts a callable to an R function, but leaves the formals as `...`:

https://github.com/rstudio/reticulate/blob/7da9ac7ea53b33d5fc2f8dfb2a85d08f1b85f0e0/R/python.R#L1224-L1239

It should be easy to construct some formals via [`inspect.signature()`](https://docs.python.org/3/library/inspect.html#inspect.signature).

Also the repr should be adapted. `<function myfunc>` isn’t super helpful.

flying-sheep:
I’m not very experienced in Rcpp *or* Python’s C API, but I’m there I think!

This PR converts signatures in a compatible way.

1. Defaults are all set to `NULL` for the time being. This can be improved in a further PR, but isn’t super important, as it doesn’t matter for autocompletion (as R and RStudio autocomplete `paramname = ` for you and therefore ignore that information).
2. `...` in R replaces `*args` and `**kw` at once, so the first occurrence of either is replaced by `...` in the formals.
3. R doesn’t support positional-only parameters such as x and y here. So e.g. the suggestions for `pow` will be wrong:

    ```r
    > builtins$pow(<tab>
          | x =    
          | y =    
          | z =    
    > pow(x = 2, y = 2)
    Error in py_call_impl(callable, dots$args, dots$keywords):
      TypeError: pow() takes no keyword arguments
    Traceback:

    1. pow(x = 2, y = 2)
    ```

Fixes #553

### DIFF
--- a/R/python.R
+++ b/R/python.R
@@ -1225,6 +1225,7 @@ py_callable_as_function <- function(callable, convert) {
   f <- function(...) {
     # Cannot use list(...), as we replace the formals() later.
     call <- sys.call()
+    cat('Call: '); print(call)
     call[[1]] <- as.name('list')
     dots <- py_resolve_dots(eval(call))
     result <- py_call_impl(callable, dots$args, dots$keywords)
@@ -1239,9 +1240,8 @@ py_callable_as_function <- function(callable, convert) {
       result
     }
   }
-  if (length(attr(f, "formals")) > 0)
-    formals(f) <- attr(f, "formals")
-  attr(f, "formals") <- NULL
+  formals(f) <- py_get_formals(callable, convert)
+  cat('Formals: ', paste(names(formals(f)), '=', formals(f), collapse = ', '), '\n')
   f
 }
 
--- a/R/python.R
+++ b/R/python.R
@@ -1226,7 +1226,7 @@ py_callable_as_function <- function(callable, convert) {
     # Cannot use list(...), as we replace the formals() later.
     call <- sys.call()
     call[[1]] <- as.name('list')
-    dots <- py_resolve_dots(eval(call))
+    dots <- py_resolve_dots(eval.parent(call))
     result <- py_call_impl(callable, dots$args, dots$keywords)
     if (convert) {
       result <- py_to_r(result)
@@ -1239,7 +1239,9 @@ py_callable_as_function <- function(callable, convert) {
       result
     }
   }
-  formals(f) <- py_get_formals(callable, convert)
+  sig <- py_get_formals(callable, convert)
+  if (!is.null(sig))
+    formals(f) <- sig
   f
 }
 
--- a/R/python.R
+++ b/R/python.R
@@ -1222,8 +1222,11 @@ py_eval <- function(code, convert = TRUE) {
 }
 
 py_callable_as_function <- function(callable, convert) {
-  function(...) {
-    dots <- py_resolve_dots(list(...))
+  f <- function(...) {
+    # Cannot use list(...), as we replace the formals() later.
+    call <- sys.call()
+    call[[1]] <- as.name('list')
+    dots <- py_resolve_dots(eval(call))
     result <- py_call_impl(callable, dots$args, dots$keywords)
     if (convert) {
       result <- py_to_r(result)
@@ -1236,6 +1239,9 @@ py_callable_as_function <- function(callable, convert) {
       result
     }
   }
+  formals(f) <- attr(f, "formals")
+  attr(f, "formals") <- NULL
+  f
 }
 
 py_resolve_dots <- function(dots) {
--- a/R/python.R
+++ b/R/python.R
@@ -1239,7 +1239,8 @@ py_callable_as_function <- function(callable, convert) {
       result
     }
   }
-  formals(f) <- attr(f, "formals")
+  if (length(attr(f, "formals")) > 0)
+    formals(f) <- attr(f, "formals")
   attr(f, "formals") <- NULL
   f
 }
--- a/R/python.R
+++ b/R/python.R
@@ -1239,9 +1239,12 @@ py_callable_as_function <- function(callable, convert) {
       result
     }
   }
-  sig <- py_get_formals(callable, convert)
-  if (!is.null(sig))
-    formals(f) <- sig
+  attr(f, 'get_formals_error') <- tryCatch({
+    sig <- py_get_formals(callable, convert)
+    if (!is.null(sig))
+      formals(f) <- sig
+    NULL
+  }, error = function(e) e)
   f
 }
 
--- a/R/python.R
+++ b/R/python.R
@@ -1225,7 +1225,6 @@ py_callable_as_function <- function(callable, convert) {
   f <- function(...) {
     # Cannot use list(...), as we replace the formals() later.
     call <- sys.call()
-    cat('Call: '); print(call)
     call[[1]] <- as.name('list')
     dots <- py_resolve_dots(eval(call))
     result <- py_call_impl(callable, dots$args, dots$keywords)
@@ -1241,7 +1240,6 @@ py_callable_as_function <- function(callable, convert) {
     }
   }
   formals(f) <- py_get_formals(callable, convert)
-  cat('Formals: ', paste(names(formals(f)), '=', formals(f), collapse = ', '), '\n')
   f
 }
 
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -602,8 +602,6 @@ bool py_is_callable(PyObjectRef x) {
     return py_is_callable(x.get());
 }
 
-Rcpp::Pairlist py_get_formals(PyObject* func, bool convert);
-
 
 // convert a python object to an R object
 SEXP py_to_r(PyObject* x, bool convert) {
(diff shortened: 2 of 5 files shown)


flying-sheep:
The failure of the last test is spurious btw, this is ready.

R/python.R
```diff
@@ -1222,8 +1222,11 @@ py_eval <- function(code, convert = TRUE) {
 }
 
 py_callable_as_function <- function(callable, convert) {
-  function(...) {
-    dots <- py_resolve_dots(list(...))
+  f <- function(...) {
+    # Cannot use list(...), as we replace the formals() later.
+    call <- sys.call()
+    call[[1]] <- as.name('list')
```
kevinushey: I've typically seen `alist()` used rather than `list()` for this sort of purpose. Would that work here?

src/python.cpp
```diff
@@ -942,6 +942,85 @@ SEXP py_to_r(PyObject* x, bool convert) {
   }
 }
 
+// [[Rcpp::export]]
+SEXP py_get_formals(PyObjectRef func, bool convert) {
+  PyObjectPtr inspect(py_import("inspect"));
+  if (inspect.is_null()) stop(py_fetch_error());
+  PyObjectPtr get_signature(PyObject_GetAttrString(inspect.get(), "signature"));
```
kevinushey: This is only available since Python 3.3, but I think we need to still support Python 2.7.

src/python.cpp
```diff
@@ -942,6 +942,85 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
+  if (var_pos.is_null()) stop(py_fetch_error());
+  PyObjectPtr var_kw(PyObject_GetAttrString(param_class.get(), "VAR_KEYWORD"));
+  if (var_kw.is_null()) stop(py_fetch_error());
+  PyObjectPtr kw_only(PyObject_GetAttrString(param_class.get(), "KEYWORD_ONLY"));
+  if (kw_only.is_null()) stop(py_fetch_error());
+
+  Rcpp::Pairlist formals;
+  PyObject* param;
+  bool var_encountered = false;
+  while ((param = PyIter_Next(param_iter.get()))) {
```
kevinushey: `PyIter_Next()` returns a new reference. Does that reference need to be cleaned up here?

R/python.R
```diff
@@ -1236,6 +1239,10 @@ py_callable_as_function <- function(callable, convert) {
       result
     }
   }
+  sig <- py_get_formals(callable, convert)
```
kevinushey: Maybe wrap in `tryCatch()` and ignore errors? (since otherwise we just use `...` and things work as expected, just without explicit formals)

kevinushey:
This PR is a bit scary since it affects _every_ wrapper R function we use, and I worry about the potential for breakage in downstream packages due to subtle differences in argument matching (e.g. `keras` and `tensorflow`).

To make things more complicated, there's also a recent PEP for positional-only arguments: https://www.python.org/dev/peps/pep-0570/. Would this PR still work in light of that PEP being implemented?

R/python.R
flying-sheep: No, as `alist` exists so people can manually create a structure similar to what `sys.call` returns.

`sys.call` returns an unevaluated `call` expression and we want to evaluate things in the end. What `sys.call` returns in my implementation when we call `py_func(a, b = c, d = 1)` is:

```r
structure(
  list(
    as.name('py_func'),
    as.name('a'),
    b = as.name('c'),
    d = 1
  ),
  class = 'call'
)
```

I then replace `py_func` with `list` and evaluate the thing in the right context.

What `alist` does is simply explained by its definition: `function(...) as.list(sys.call())[-1L]`. It does nothing that isn’t already done at that point.

I hope this explains it. I rephrased this far too many times, so if it doesn’t, I fear you have to dive into this yourself, it’s a bit mind-bending and hard to phrase.

src/python.cpp
flying-sheep: Python 2’s EOL is in 140 days. Would you like to sign the https://python3statement.org?

Until then we can make this conditional right? Just return NULL if the module isn’t importable?

src/python.cpp
flying-sheep: Thank you, right! Done in 54389e84a547a5207c0098d41362cb64750f047e

IDK if there’s a more idiomatic way to do this in your codebase: Maybe I should use `py_iter_next`?

R/python.R
flying-sheep: Sure! Can I log errors somewhere in case we want to debug some signature that doesn’t convert even when it should?

flying-sheep: OK, I’m storing them in an attribute now.

src/python.cpp
flying-sheep: I’m wrapping the thing in a PyObjectPtr now!

flying-sheep:
I know it’s scary, but there’s no way to get the desired behavior without modifying the formals (which influences how functions are called). But I can’t think of a way this can go wrong: we pass positional arguments as positional, named arguments as named, and add `...` on the first occurrence of `*args` or `**kw`.

I explained positional-only parameters in the first comment, you must have missed that :wink:

src/python.cpp
kevinushey: Right, we can just make it conditional.

We'd like to drop Python 2.7 support but as long as some of the larger Python projects like Tensorflow and Keras continue to support Python 2.7 we'll have to support it as well.

src/python.cpp
```diff
@@ -978,14 +978,14 @@ SEXP py_get_formals(PyObjectRef func, bool convert) {
   if (kw_only.is_null()) stop(py_fetch_error());
 
   Rcpp::Pairlist formals;
-  PyObject* param;
+  PyObjectPtr param;
   bool var_encountered = false;
-  while ((param = PyIter_Next(param_iter.get()))) {
-    PyObjectPtr param_name(PyObject_GetAttrString(param, "name"));
+  while (param.assign(PyIter_Next(param_iter.get())), !param.is_null()) {
```
kevinushey: This still isn't quite right -- `assign()` doesn't call `Py_DecRef()` on a previously-held reference to a Python object, so I think there would still only be one `Py_DecRef()` being called at the end when the (single) `PyObjectPtr` is destructed.

One solution would be to construct a new `PyObjectPtr` for each iterated item, e.g.

```
while (true) {
    PyObjectPtr param(PyIter_Next(param_iter.get()))
    if (param == nullptr)
        break;
    < ... >
}
```

flying-sheep: That’s of course less pretty than what I was going for, but sure!

Alternatively: Is there a reason `assign()` doesn’t call `Py_DecRef()`? Seem like an oversight, unless I’m missing something.

src/python.cpp
flying-sheep: Actually due to using `tryCatch`, this works now. Every function will simply have

```r
attr(f, 'get_formals_error') <- simpleError("AttributeError: module 'inspect' has no attribute 'signature'")
```

flying-sheep:
More thoughts about parameters:

There’s pos-only, kw-only, and hybrid params.

- R doesn’t have pos-only params, and allows partially matching hybrid params when called by name. `...` separates hybrid and kw-only params and soaks up additional positional or keyword arguments.
- Python has all three kinds of params. `/` separates pos-only from hybrid params, and `*` or `*args` separates hybrid from kw-only params. `*args` soaks up additional positional arguments. `**kw` comes at the end and soaks up additional keyword arguments.

None of the differences should be a problem:

- Pos-only params can’t be represented in R, so when calling a Python function from R, trying to pass a pos-only param by name will simply result in the same error as when doing this in Python.
- R’s partial matching doesn’t apply, since `sys.call` captures things exactly like we passed it. So calling a function with signature `def foo(long_name = 1, **kw): print(long_name, kw)` from R works as expected: `foo(long = 2)` → `1 {'long': 2}`
- The difference between `...` and `*`, `*args`, `**kw` is also no problem: Every parameter after any of those* has to be specified by name.

*except for `**kw` as there can’t be anything behind it

flying-sheep:
Great, thank you! I really honestly believe this covers all bases, but you never know…

If a corner case where this fails appears, it should be easy to detect and fix it (or skip replacing the signature) without rolling back the whole thing.

skeydan:
Hi @flying-sheep I just checked this out due to https://github.com/rstudio/reticulate/issues/581

What is the expected output of your example from above, 

```
formals(builtins$pow)
```

For me it is 

`$...`

Thanks!

flying-sheep:
Interesting! So some builtins have it and some don’t.

```py
>>> import inspect
>>> inspect.signature(print)
ValueError: no signature found for builtin <built-in function print>
>>> inspect.signature(pow)
<Signature (x, y, z=None, /)>
```

But #581 should be fixed by this PR:

```r
> reticulate::py_run_string('def add(x,y):\n  return x+y')
> formals(reticulate::py$add)                                                                                  
$x


$y


```

skeydan:
Hi @flying-sheep thanks, you're right! I was pretty sure I had updated from master, and still have no idea why it didn't work,  but after installing another time it works now :-)

PR #590, opened 2019-09-02
Add bit64 support

kdkavanagh:
Allow for 64bit integers to be passed between Py <-> R using the bit64 package, inspired by https://gallery.rcpp.org/articles/creating-integer64-and-nanotime-vectors/

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -201,17 +201,18 @@ int narrow_array_typenum(int typenum) {
   case NPY_SHORT:
   case NPY_USHORT:
   case NPY_INT:
+  case NPY_LONG:
     typenum = NPY_LONG;
     break;
     // double
   case NPY_UINT:
   case NPY_ULONG:
   case NPY_ULONGLONG:
-  case NPY_LONG:
   case NPY_LONGLONG:
   case NPY_HALF:
   case NPY_FLOAT:
   case NPY_DOUBLE:
+
     typenum = NPY_DOUBLE;
     break;
 
@@ -618,9 +619,18 @@ SEXP py_to_r(PyObject* x, bool convert) {
       return LogicalVector::create(x == Py_True);
 
     // integer
-    else if (scalarType == INTSXP)
-      return IntegerVector::create(PyInt_AsLong(x));
-
+    else if (scalarType == INTSXP) {
+      long val = PyInt_AsLong(x);
+      if(val > std::numeric_limits<int>::max()) {
+        Rcpp::NumericVector vec(1);
+        std::memcpy(&(vec[0]), &(val), sizeof(double));
+        vec.attr("class") = "integer64";
+        return vec;
+      }
+      else{
+        return IntegerVector::create(val);
+      }
+    }
     // double
     else if (scalarType == REALSXP)
       return NumericVector::create(PyFloat_AsDouble(x));
@@ -653,9 +663,24 @@ SEXP py_to_r(PyObject* x, bool convert) {
       return vec;
     } else if (scalarType == INTSXP) {
       Rcpp::IntegerVector vec(len);
-      for (Py_ssize_t i = 0; i<len; i++)
-        vec[i] = PyInt_AsLong(PyList_GetItem(x, i));
+      for (Py_ssize_t i = 0; i<len; i++) {
+        long num = PyInt_AsLong(PyList_GetItem(x, i));
+        if(num > std::numeric_limits<int>::max()) {
+          //We need to start over an interpret as 64 bit int
+          Rcpp::NumericVector nVec(len);
+          long long* res_ptr  = (long long*) dataptr(nVec);
+          for (Py_ssize_t j = 0; j<len; j++) {
+            res_ptr[j] = PyInt_AsLong(PyList_GetItem(x, j));;
+          }
+          nVec.attr("class") = "integer64";
+          return nVec;
+          break;
+        } else {
+          vec[i] = num;
+        }
+      }
       return vec;
+
     } else if (scalarType == CPLXSXP) {
       Rcpp::ComplexVector vec(len);
       for (Py_ssize_t i = 0; i<len; i++) {
@@ -769,8 +794,17 @@ SEXP py_to_r(PyObject* x, bool convert) {
       case NPY_LONG: {
         npy_long* pData = (npy_long*)PyArray_DATA(array);
         rArray = Rf_allocArray(INTSXP, dimsVector);
-        for (int i=0; i<len; i++)
+        for (int i=0; i<len; i++) {
+          if(pData[i] > std::numeric_limits<int>::max()) {
+            Rcpp::NumericVector nVec(len);
+            //Inspired by https://gallery.rcpp.org/articles/creating-integer64-and-nanotime-vectors/
+            std::memcpy(&(nVec[0]), &(pData[0]), len * sizeof(double));
+            nVec.attr("class") = "integer64";
+            nVec.attr("dim") = dimsVector;
+            return nVec;
+          }
           INTEGER(rArray)[i] = pData[i];
+        }
         break;
       }
       case NPY_DOUBLE: {
@@ -1083,7 +1117,7 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
         typenum = NPY_INT;
       data = &(INTEGER(sexp)[0]);
     } else if (type == REALSXP) {
-      typenum = NPY_DOUBLE;
+      typenum = x.inherits("integer64") ? NPY_LONG : NPY_DOUBLE;
       data = &(REAL(sexp)[0]);
     } else if (type == LGLSXP) {
       typenum = NPY_BOOL;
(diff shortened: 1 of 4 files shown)


src/python.cpp
```diff
@@ -653,9 +663,24 @@ SEXP py_to_r(PyObject* x, bool convert) {
       return vec;
     } else if (scalarType == INTSXP) {
       Rcpp::IntegerVector vec(len);
-      for (Py_ssize_t i = 0; i<len; i++)
-        vec[i] = PyInt_AsLong(PyList_GetItem(x, i));
+      for (Py_ssize_t i = 0; i<len; i++) {
+        long num = PyInt_AsLong(PyList_GetItem(x, i));
+        if(num > std::numeric_limits<int>::max()) {
```
kdkavanagh: Alternatively, could loop thru array once first to check max value, then create the right vector type the first time

kevinushey:
I don't think it's a good idea to have the output vector type depend on the runtime value -- I think we need to ensure type stability, and the user has to opt-in to requesting bit64 vectors in some way. This is especially true since there are operations that can be performed on 'native' R integers that will not work with a bit64 vector.

I don't yet know what this interface should look like, though. My instinct says that there should be some way for users to provide custom converters, that accept R objects and return Python objets, and vice versa.

Also worth saying: the underlying SEXP type of an `integer64` vector is `REALSXP`; we'd likely need to audit a lot of `reticulate` code to ensure that we're doing the right thing when we receive a `REALSXP` that happens to be an `integer64` vector.

kdkavanagh:
I don't disagree w.r.t opting in with some option.  I've never built an enterprise-grade R package before, so not sure how to be incorporate that, but could imagine that being configured by the user after loading the reticulate package

eddelbuettel:
It's the right call. I am using `bit64` aka `integer64` aka `nanotime` _very carefully_ at work to transfer full resolution timestamo, and some `int64_t` index values.  One has to be really careful to 'cast corectly' and treat it correctly on the receiving end or else one mistakes it for a standard `numeric` filled with ... garbage bits.   This works, but only with some care.  It is probably not the right tool for a general purpose converter.

kdkavanagh:
I will take a stab at implementing an option tonight

OssiLehtinen:
Any news on this one? Looking forward to a solution, as this would solve a bunch of problems!

kdkavanagh:
Added two options which default to `FALSE`
```
  options(reticulate.long_as_bit64=TRUE)
  options(reticulate.ulong_as_bit64=TRUE)
```

kevinushey:
Thanks -- I'll try to re-review this soon.

src/python.cpp
```diff
@@ -58,6 +58,27 @@ std::wstring s_python_v3;
...
 std::wstring s_pythonhome_v3;
 
+const std::string CONFIG_LONG_AS_BIT64="reticulate.long_as_bit64";
+const std::string CONFIG_ULONG_AS_BIT64="reticulate.ulong_as_bit64";
+
+template<typename T>
+T getConfig(std::string config, T defValue) {
+  Environment base( "package:base" ) ;
+  Function getOption = base["getOption"];
+  SEXP s = getOption(config, defValue);
```
kevinushey: I believe we could use `Rf_GetOption` directly here.

src/python.cpp
```diff
@@ -612,9 +650,18 @@ SEXP py_to_r(PyObject* x, bool convert) {
       return LogicalVector::create(x == Py_True);
 
     // integer
-    else if (scalarType == INTSXP)
-      return IntegerVector::create(PyInt_AsLong(x));
-
+    else if (scalarType == INTSXP) {
+      long val = PyLong_AsLong(x);
+      if((val > std::numeric_limits<int>::max() || val < std::numeric_limits<int>::min()) && convertLongToBit64()) {
```
kevinushey: I worry a bit that this could be confusing if values were converted to `integer64` only "sometimes". That is, I'm not sure if the runtime type should depend on the runtime value. Would it make sense to _always_ make this conversion if the user has opted in?

src/python.cpp
```diff
@@ -647,9 +694,23 @@ SEXP py_to_r(PyObject* x, bool convert) {
       return vec;
     } else if (scalarType == INTSXP) {
       Rcpp::IntegerVector vec(len);
-      for (Py_ssize_t i = 0; i<len; i++)
-        vec[i] = PyInt_AsLong(PyList_GetItem(x, i));
+      for (Py_ssize_t i = 0; i<len; i++) {
+        long num = PyLong_AsLong(PyList_GetItem(x, i));
+        if((num > std::numeric_limits<int>::max() || num < std::numeric_limits<int>::min()) && convertLongToBit64()) {
```
kevinushey: Same concern here as above.

src/python.cpp
```diff
@@ -1223,15 +1307,33 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
...
   } else if (type == REALSXP) {
+    bool isBit64  = x.inherits("integer64");
+    bool isUnsigned = x.inherits("np.ulong");
+    bool isInt64 = isBit64 && !isUnsigned;
+    bool isUint64 = isBit64 && isUnsigned;
     if (LENGTH(sexp) == 1) {
       double value = REAL(sexp)[0];
-      return PyFloat_FromDouble(value);
+      if(isInt64) {
+        return PyInt_FromLong(reinterpret_cast<long&>(value));
```
kevinushey: Is it really safe to reinterpret a value as a reference? I've usually seen type punning done as e.g.

```
*(long*)(&value)
```

This is still technically undefined behavior, though. My understanding is that the only standards-compliant way of performing type punning is with `memcpy()`. See: https://stackoverflow.com/a/17790026/1342082

(In C++20, we will have bitcast: https://en.cppreference.com/w/cpp/numeric/bit_cast)

src/python.cpp
```diff
@@ -58,6 +58,27 @@ std::wstring s_python_v3;
 std::string s_pythonhome;
 std::wstring s_pythonhome_v3;
 
+const std::string CONFIG_LONG_AS_BIT64="reticulate.long_as_bit64";
```
kevinushey: Should these be `static const`? (since they should have internal linkage)

src/python.cpp
kevinushey: See https://godbolt.org/z/VDCcVF for an example of a helper function that could be used for casting here.

src/python.cpp
```diff
@@ -240,12 +268,24 @@ int narrow_array_typenum(int typenum) {
...
+}
+
 int narrow_array_typenum(PyArrayObject* array) {
-  return narrow_array_typenum(PyArray_TYPE(array));
+  return narrow_array_typenum(typenum(array));
 }
 
 int narrow_array_typenum(PyArray_Descr* descr) {
-  return narrow_array_typenum(descr->type_num);
+  return narrow_array_typenum(typenum(descr->type_num));
```
kevinushey:
```suggestion
  return narrow_array_typenum(typenum(descr));
```

src/python.cpp
```diff
@@ -647,9 +694,23 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
     } else if (scalarType == INTSXP) {
       Rcpp::IntegerVector vec(len);
-      for (Py_ssize_t i = 0; i<len; i++)
-        vec[i] = PyInt_AsLong(PyList_GetItem(x, i));
+      for (Py_ssize_t i = 0; i<len; i++) {
+        long num = PyLong_AsLong(PyList_GetItem(x, i));
+        if((num > std::numeric_limits<int>::max() || num < std::numeric_limits<int>::min()) && convertLongToBit64()) {
+          //We need to start over an interpret as 64 bit int
+          Rcpp::NumericVector nVec(len);
+          long long* res_ptr  = (long long*) dataptr(nVec);
```
kevinushey: Note that using `long long*` explicitly requires us to use C++11. This is okay but I think we need to update `SystemRequirements` to indicate that.

kevinushey:
Thanks for the pull request! We require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

colearendt:
This is really exciting stuff! Unfortunately, I gave it a try and ran into a few painful points:

- something weird in my environment is causing certain operations to segfault (`import pickle`, create a big integer, etc.), so I haven't been able to create a reprex. (NOTE: this is happening for the CRAN released version of reticulate as well, so I suspect this problem is elsewhere).
- the print method if `bit64` has not been loaded/referenced is pretty confusing:
```
sync_res$items[[1]]$id
[1] 1.196682e-314
attr(,"class")
[1] "integer64"
```
- However, just calling a `bit64` function directly fixes:
```
> bit64::as.integer64(1234)
integer64
[1] 1234
> sync_res$items[[1]]$id
integer64
[1] 2422110596
```

There are definitely some nice things here, though! Thanks!

eddelbuettel:
> However, just calling a bit64 function directly fixes

Our `nanotime` package also wraps around `bit64`, and I have seen this too.  Now, it is a little tricky as we generally want an optional component to be truly optional, but generally speaking an `if (requireNamespace(...)))` test (and load of the package) before code blocks with 64-bit ints seems to help (and I think that is what `data.table` does too).  Making it a globally imported package helps too, but that may be too blunt.

GreenGrassBlueOcean:
Dear @kevinushey @kdkavanagh 

Are there any plans to accept this pull request in to reticulate?
This would be very useful for my own r-package as this api often returns large integers which then turn out to be -1.

spittssj:
Is there any update on this issue?

I am starting to use `rgee` which wraps the Python Google Earth Engine API through `reticulate` and I ran into this issue with the `time` field that comes with the ImageCollections of time series satellite data.

ssh352:
It has been many years since R has no native int64 support. Do the core-team have a plan to add native support before many users switch to Python.

PR #598, opened 2019-09-25
Add install_miniconda() helper

kevinushey:
This PR ports some of the work done by @hafen in [rminiconda](https://github.com/hafen/rminiconda) and adds a helper function that installs Miniconda.


R/use_python.R
```diff
@@ -87,7 +87,33 @@ use_condaenv <- function(condaenv = NULL, conda = "auto", required = FALSE) {
     stop("Unable to locate conda environment '", condaenv, "'.")
 
   if (!is.null(condaenv))
-    use_python(conda_env_python, required = required)
+    use_python(conda_env_python[[1]], required = required)
```
kevinushey: One thing that surprised me: even if you supply a Conda binary explicitly, the different Conda installations seem to know about each other and their environments. This means that `conda_list()` will return environments from _all_ Conda installations discovered on the system.

I wonder if we need to limit the check here to Conda environments explicitly associated with / contained within the associated Conda installation?

R/miniconda.R
```diff
@@ -0,0 +1,197 @@
+
+#' Install Miniconda
+#'
+#' Download the [Miniconda](https://docs.conda.io/en/latest/miniconda.html)
+#' installer, and use it to install Miniconda.
+#'
+#' @param path The path in which Miniconda will be installed. Note that the
+#'   installer does not support paths containing spaces.
+#' 
+#' @param version The version of Python associated with the Miniconda installer
```
hafen: I think this can only strictly be `2` or `3`. Users may think that they can specify something like `3.6` or `3.7`, in which case the installer URL won't be valid. It might be good to make that clear in the documentation and validate that it is only `2` or `3` in the code.

R/miniconda.R
```diff
@@ -0,0 +1,197 @@
...
+#'   to be used.
+#' 
+#' @param update Boolean; update to the latest version of Miniconda after install?
+#' 
+#' @param python The version of Python to use. You can request specific
```
hafen: These docs might be a little confusing - this is for switching to a different specific version of Python in the conda environment, right?

R/use_python.R
```diff
@@ -87,7 +87,33 @@ use_condaenv <- function(condaenv = NULL, conda = "auto", required = FALSE) {
...
 
   invisible(NULL)
 }
+
+#' @rdname use_python
+#' @export
+use_miniconda <- function(condaenv = NULL, required = FALSE) {
+  
+  # check that Miniconda is installed
+  path <- miniconda_path()
```
hafen: Not a big deal, but could use `miniconda_exists()` here?

kevinushey:
I think this PR (installation of Miniconda) is good to go. I'm going to merge this and then start working on the Python declaration + discover + documentation in a separate PR.

PR #605, opened 2019-10-08
initial work on python dependency discovery + install

kevinushey:
NOTE: not yet ready for merge; putting this up as a PR just to further facilitate discussion.

---

This PR intends to make it possible for R packages to declare their reticulate Python requirements, and to have `reticulate` automatically discover and install those declared dependencies when using a Miniconda environment.

The intention is that users will be able to run R packages that act as wrappers to Python packages with minimal overhead; that is, they will not need to worry about Python environment management, as `reticulate` will take care of that automagically.

R packages can declare the packages they require with a new field in the DESCRIPTION file. For example, a package wrapping TensorFlow might use:

```
reticulate@R: list(
    list(package = "tensorflow", pip = TRUE),
    list(package = "keras", pip = TRUE)
    )
```

When `reticulate` initializes Python, it will look at the set of loaded packages, find these declared Python dependencies, and then install the required dependencies.

Some open questions:

1. What about version requirements? Do we want to think about this now, or leave it for later?
2. How do we avoid package versions getting stale? Should we prompt for upgrades? Should we try to "live at head" when possible?

For what it's worth, one thing that might make our lives easier: for version requirements, we can delegate that responsibility to `pip` / `conda` as appropriate; e.g. we could use

```
pip install -r requirements.txt
```

where `requirements.txt` is generated based on the declared Python requirements. We may need to do some collapsing of version requirements but that's at least simpler than other problems.

A similar facility exists for `conda` with e.g.

```
conda env update -f packages.yml
```

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -51,6 +51,9 @@ ensure_python_initialized <- function(required_module = NULL) {
 
     # remap output streams to R output handlers
     remap_output_streams()
+    
+    # install required packages
+    configure_environment()
 
   }
 }
--- a/R/python-packages.R
+++ b/R/python-packages.R
@@ -1,0 +1,68 @@
+
+configure_environment <- function() {
+  
+  # only done if we're using miniconda for now
+  config <- py_config()
+  home <- miniconda_path()
+  if (substring(config$python, 1, nchar(home)) != home)
+    return(FALSE)
+  
+  # find package requirements
+  pkgreqs <- python_package_requirements()
+  reqs <- unlist(pkgreqs, recursive = FALSE, use.names = FALSE)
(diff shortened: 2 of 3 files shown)


kevinushey:
I'm going to leave this as-is for now and tackle version requirements in a separate PR (under the assumption that it will not be a hard requirement for the majority of users)

jjallaire:
Since we can't easily reconcile 2 packages that have diverging version requirements I think we need to be careful to discourage people from using whatever we put in place. Since pip often gives you a very recent version by default, it seems like we might want to encourage only requiring downlevel versions (i.e. b/c we know something is broken in the more recent version).

PR #612, opened 2019-10-17
provide readline hook for Python input

### LINKED ISSUE #610: Using python's input() function.
RStudio locks up when requesting user input with the input() function in a python script. Is there a way to get this to work or just a limitation of the implementation?

kevinushey:
Closes #610.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 
 Install the development version of reticulate with `remotes::install_github("rstudio/reticulate")`.
 
+- `input()` and `raw_input()` now work as expected within `repl_python()`. (#610)
+
 - `reticulate` now more consistently normalizes the paths reported by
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -3,6 +3,8 @@
 
 Install the development version of reticulate with `remotes::install_github("rstudio/reticulate")`.
 
+- `input()` and `raw_input()` now work as expected within `repl_python()`. (#610)
```
jjallaire: Will this also make them behave correctly when running from any piece of Python code? (e.g. code in a .py script or Python package)

kevinushey: Indeed, it works in all of these contexts. (we override the `input` builtin so our hook gets called in any context where Python's `input()` might get called)

jjallaire:
Okay, great!

PR #618, opened 2019-10-23
comma is missing and prevent installation

cderv:
You may have already see that 🤔

---
This change is [Reviewable](https://reviewable.io/reviews/rstudio/reticulate/618)


kevinushey:
Thanks! I just made this same fix on master :-)

PR #620, opened 2019-10-24
WIP: py_convert_pandas_df

skeydan:
Purpose: Convert `pandas` dataframes on c++ level instead of in R, to reduce number of round trips.

This is just to show the general approach; in this initial version, the only "non-automatic" data type (meaning, the only one for whom calling `py_to_r` will not be sufficient) is `pd.Categorical`. More data types will need specific handling.

I could also imagine that we'd want to split out these individual functions, and/or add them as branches to `py_to_r`?

Thanks for taking a look :-)

### DIFF
--- a/src/libpython.h
+++ b/src/libpython.h
@@ -307,6 +307,9 @@ LIBPYTHON_EXTERN void (*PySys_SetArgv_v3)(int, wchar_t **);
 
 LIBPYTHON_EXTERN void (*PySys_WriteStderr)(const char *format, ...);
 
+LIBPYTHON_EXTERN PyObject* (*PyObject_CallMethod)(PyObject *o, const char *name, const char *format, ...);
+LIBPYTHON_EXTERN PyObject* (*PySequence_GetItem)(PyObject *o, Py_ssize_t i);
+  
 #define PyObject_TypeCheck(o, tp) ((PyTypeObject*)Py_TYPE(o) == (tp)) || PyType_IsSubtype((PyTypeObject*)Py_TYPE(o), (tp))
 
 #define PyType_Check(o) PyObject_TypeCheck(o, PyType_Type)
--- a/src/libpython.h
+++ b/src/libpython.h
@@ -309,6 +309,7 @@ LIBPYTHON_EXTERN void (*PySys_WriteStderr)(const char *format, ...);
 
 LIBPYTHON_EXTERN PyObject* (*PyObject_CallMethod)(PyObject *o, const char *name, const char *format, ...);
 LIBPYTHON_EXTERN PyObject* (*PySequence_GetItem)(PyObject *o, Py_ssize_t i);
+LIBPYTHON_EXTERN int (*PyObject_IsTrue)(PyObject *o);
   
 #define PyObject_TypeCheck(o, tp) ((PyTypeObject*)Py_TYPE(o) == (tp)) || PyType_IsSubtype((PyTypeObject*)Py_TYPE(o), (tp))
 
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2295,8 +2295,12 @@ SEXP py_convert_pandas_df(PyObjectRef df) {
     // access Series in slot 1
     PyObjectPtr series(PySequence_GetItem(tuple, 1));
     Py_IncRef(series);
+    
+    // delegate to py_convert_pandas_series
     PyObjectRef series_ref(series, df.convert());
+    
     RObject R_obj = py_convert_pandas_series(series_ref);
+    
     Py_DecRef(series);
     
     list.push_back(R_obj);
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2215,6 +2215,7 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
   PyObjectPtr dtype(PyObject_GetAttrString(series, "dtype"));
   PyObjectPtr name(PyObject_GetAttrString(dtype, "name"));
   
+  
   // as well as values
   PyObjectPtr values(PyObject_GetAttrString(series, "values"));
   
@@ -2292,7 +2293,7 @@ SEXP py_convert_pandas_df(PyObjectRef df) {
     // access Series in slot 1
     PyObjectPtr series(PySequence_GetItem(tuple, 1));
     // delegate to py_convert_pandas_series
-    PyObjectRef series_ref(series, df.convert());
+    PyObjectRef series_ref(series.detach(), df.convert());
     RObject R_obj = py_convert_pandas_series(series_ref);
 
     list.push_back(R_obj);
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2248,11 +2248,8 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
     }
     
     // populate character vector to hold levels
-    n = Rf_xlength(R_levels);
-    CharacterVector factor_levels(n);
-    for (int i = 0; i < n; ++i) {
-      factor_levels[i] = STRING_ELT(R_levels, i);
-    }
+    CharacterVector factor_levels(R_levels);
+    factor_levels.attr("dim") = R_NilValue;
     
     factor.attr("class") = "factor";
     factor.attr("levels") = factor_levels;
@@ -2294,15 +2291,10 @@ SEXP py_convert_pandas_df(PyObjectRef df) {
     
     // access Series in slot 1
     PyObjectPtr series(PySequence_GetItem(tuple, 1));
-    Py_IncRef(series);
-    
     // delegate to py_convert_pandas_series
     PyObjectRef series_ref(series, df.convert());
-    
     RObject R_obj = py_convert_pandas_series(series_ref);
-    
-    Py_DecRef(series);
-    
+
     list.push_back(R_obj);
   }
 
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2215,10 +2215,6 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
   PyObjectPtr dtype(PyObject_GetAttrString(series, "dtype"));
   PyObjectPtr name(PyObject_GetAttrString(dtype, "name"));
   
-  
-  // as well as values
-  PyObjectPtr values(PyObject_GetAttrString(series, "values"));
-  
   RObject R_obj;
   
   // special treatment for pd.Categorical
@@ -2258,9 +2254,49 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
     
     R_obj = factor;
     
-    // default case
+  // special treatment for pd.TimeStamp
+  // if available, time zone information will be respected,
+  // but values returned to R will be in UTC
+  } else if (as_std_string(name) == "datetime64[ns]" ||
+    // if a time zone is present, dtype is "object"
+    PyObject_HasAttrString(series, "dt")) {
+
+    // pd.Series.items() returns an iterator over (index, value) pairs
+    PyObjectPtr items(PyObject_CallMethod(series, "items", NULL));
+    
+    std::vector<double> posixct;
+    
+    while (true) {
+      
+      // get next tuple
+      PyObjectPtr tuple(PyIter_Next(items));
+      if (tuple.is_null()) {
+        if (PyErr_Occurred())
+          stop(py_fetch_error());
+        else
+          break;
+      }
+    
+     // access value in slot 1
+     PyObjectPtr values(PySequence_GetItem(tuple, 1));
+     // convert to POSIX timestamp, taking into account time zone (if set)
+     PyObjectPtr timestamp(PyObject_CallMethod(values, "timestamp", NULL));
+     
+     Datetime R_timestamp = py_to_r(timestamp, true);
+     posixct.push_back(R_timestamp);
+    }
+    
+    DatetimeVector R_posixct(posixct.size());
+    for (int i = 0; i < posixct.size(); ++i) {
+      R_posixct[i] = posixct[i];
+    }
+    
+    return R_posixct;
+    
+  // default case
   } else {
     
+    PyObjectPtr values(PyObject_GetAttrString(series, "values"));
     R_obj = py_to_r(values, series.convert());
     
   }
@@ -2272,7 +2308,7 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
 // [[Rcpp::export]]
 SEXP py_convert_pandas_df(PyObjectRef df) {
   
-  // pd.DataFrame.items() returns an Iterator over (column name, Series) pairs
+  // pd.DataFrame.items() returns an iterator over (column name, Series) pairs
   PyObjectPtr items(PyObject_CallMethod(df, "items", NULL));
   if (! (PyObject_HasAttrString(items, "__next__") || PyObject_HasAttrString(items, "next")))
     stop("Cannot iterate over object");
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
   return result;
 
 }
+
+// [[Rcpp::export]]
+SEXP py_convert_pandas_df(PyObjectRef obj) {
+  
+  // pd.DataFrame.items() iterates over (column name, Series) pairs
+  PyObjectPtr items(PyObject_CallMethod(obj, "items", NULL));
+
+  List list;
+  
+  while (true) {
+
+    // get next tuple
+    PyObjectPtr tuple(PyIter_Next(items));
+    if (tuple.is_null()) {
+      if (PyErr_Occurred())
+        stop(py_fetch_error());
+      else
+        break;
+    }
+    
+    // access Series in slot 1
+    PyObjectPtr series(PySequence_GetItem(tuple, 1));
+    
+    // and extract dtype
+    PyObjectPtr dtype(PyObject_GetAttrString(series, "dtype"));
+    PyObjectPtr dtype_str(PyObject_GetAttrString(dtype, "name"));
+    
+    // as well as values
+    PyObjectPtr values(PyObject_GetAttrString(series, "values"));
+    
+    SEXP sexp;
+
+    // special treatment for pd.Categorical
+    if (as_std_string(dtype_str) == "category") {
+      
+      // get actual values and convert to R
+      PyObjectPtr codes_(PyObject_GetAttrString(
+        PyObject_GetAttrString(
+          PyObject_GetAttrString(series, "cat"),
+          "codes"),
+        "values"));
+      SEXP codes = py_to_r(codes_, true);
+      
+      // get levels and convert to R
+      PyObjectPtr levels_(PyObject_GetAttrString(
+        PyObject_GetAttrString(dtype, "categories"),
+        "values"));
+      SEXP levels = py_to_r(levels_, true);
+      
+      // get "ordered" attribute
+      PyObjectPtr ordered_(PyObject_GetAttrString(dtype, "ordered"));
+      SEXP ordered = py_to_r(ordered_, true);
+   
+      // populate integer vector to hold factor values
+      int* codes_int = INTEGER(codes);
+      int n = Rf_length(codes);
+      IntegerVector factor(n);
+      // values need to start at 1
+      for (int i = 0; i < n; ++i) {
+        factor[i] = codes_int[i] + 1;
+      }
+
+      // populate character vector to hold levels
+      n = Rf_length(levels);
+      CharacterVector levels_chr(n);
+      for (int i = 0; i < n; ++i) {
+        levels_chr[i] = STRING_ELT(levels, i);
+      }
+
+      factor.attr("class") = "factor";
+      factor.attr("levels") = levels_chr;
+      if (*(LOGICAL(ordered)) == 1) factor.attr("ordered") = ordered;
+      
+      sexp = factor;
+    
+    // default case
+    } else {
+      
+      sexp = py_to_r(values, obj.convert());
+      
+    }
+    
+    list.push_back(sexp);
+  }
+
+  return list;
+
+}
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2240,7 +2240,7 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
     
     // populate integer vector to hold factor values
     int* codes_int = INTEGER(R_values);
-    int n = Rf_length(R_values);
+    int n = Rf_xlength(R_values);
     IntegerVector factor(n);
     // values need to start at 1
     for (int i = 0; i < n; ++i) {
@@ -2248,7 +2248,7 @@ SEXP py_convert_pandas_series(PyObjectRef series) {
     }
     
     // populate character vector to hold levels
-    n = Rf_length(R_levels);
+    n = Rf_xlength(R_levels);
     CharacterVector factor_levels(n);
     for (int i = 0; i < n; ++i) {
       factor_levels[i] = STRING_ELT(R_levels, i);
@@ -2276,7 +2276,7 @@ SEXP py_convert_pandas_df(PyObjectRef df) {
   
   // pd.DataFrame.items() returns an Iterator over (column name, Series) pairs
   PyObjectPtr items(PyObject_CallMethod(df, "items", NULL));
-  if (! PyObject_HasAttrString(items, "__next__"))
+  if (! (PyObject_HasAttrString(items, "__next__") || PyObject_HasAttrString(items, "next")))
     stop("Cannot iterate over object");
 
   std::vector<RObject> list;
@@ -2306,9 +2306,7 @@ SEXP py_convert_pandas_df(PyObjectRef df) {
     list.push_back(R_obj);
   }
 
-  List rList(list.size());
-  for (size_t i = 0; i < list.size(); i++)
-    rList[i] = list[i];
+  List rList(list.begin(), list.end());
   return rList;
 
 }
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2209,10 +2209,73 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
 }
 
 // [[Rcpp::export]]
-SEXP py_convert_pandas_df(PyObjectRef obj) {
+SEXP py_convert_pandas_series(PyObjectRef series) {
+  
+  // extract dtype
+  PyObjectPtr dtype(PyObject_GetAttrString(series, "dtype"));
+  PyObjectPtr name(PyObject_GetAttrString(dtype, "name"));
+  
+  // as well as values
+  PyObjectPtr values(PyObject_GetAttrString(series, "values"));
+  
+  RObject R_obj;
+  
+  // special treatment for pd.Categorical
+  if (as_std_string(name) == "category") {
+    
+    // get actual values and convert to R
+    PyObjectPtr cat(PyObject_GetAttrString(series, "cat"));
+    PyObjectPtr codes(PyObject_GetAttrString(cat, "codes"));
+    PyObjectPtr code_values(PyObject_GetAttrString(codes, "values"));
+    RObject R_values = py_to_r(code_values, true);
+    
+    // get levels and convert to R
+    PyObjectPtr categories(PyObject_GetAttrString(dtype, "categories"));
+    PyObjectPtr category_values(PyObject_GetAttrString(categories, "values"));
+    RObject R_levels = py_to_r(category_values, true);
+    
+    // get "ordered" attribute
+    PyObjectPtr ordered(PyObject_GetAttrString(dtype, "ordered"));
+    //RObject ordered = py_to_r(ordered_, true);
+    
+    // populate integer vector to hold factor values
+    int* codes_int = INTEGER(R_values);
+    int n = Rf_length(R_values);
+    IntegerVector factor(n);
+    // values need to start at 1
+    for (int i = 0; i < n; ++i) {
+      factor[i] = codes_int[i] + 1;
+    }
+    
+    // populate character vector to hold levels
+    n = Rf_length(R_levels);
+    CharacterVector factor_levels(n);
+    for (int i = 0; i < n; ++i) {
+      factor_levels[i] = STRING_ELT(R_levels, i);
+    }
+    
+    factor.attr("class") = "factor";
+    factor.attr("levels") = factor_levels;
+    if (PyObject_IsTrue(ordered)) factor.attr("ordered") = true;
+    
+    R_obj = factor;
+    
+    // default case
+  } else {
+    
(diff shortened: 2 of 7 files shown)


src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
   return result;
 
 }
+
+// [[Rcpp::export]]
+SEXP py_convert_pandas_df(PyObjectRef obj) {
+  
+  // pd.DataFrame.items() iterates over (column name, Series) pairs
+  PyObjectPtr items(PyObject_CallMethod(obj, "items", NULL));
```
kevinushey: Can we double-check that this does indeed return an iterable Python object? (Ensuring it has a `next` or `__next__` attribute should be sufficient)

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      else
+        break;
+    }
+    
+    // access Series in slot 1
+    PyObjectPtr series(PySequence_GetItem(tuple, 1));
+    
+    // and extract dtype
+    PyObjectPtr dtype(PyObject_GetAttrString(series, "dtype"));
+    PyObjectPtr dtype_str(PyObject_GetAttrString(dtype, "name"));
```
kevinushey: nit: why not just call it `name` to match other object <-> attribute name mappings?

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+    
+    // as well as values
+    PyObjectPtr values(PyObject_GetAttrString(series, "values"));
+    
+    SEXP sexp;
```
kevinushey: nit: for safety, maybe initialize this with `R_NilValue`? (It looks like all paths below will set this but it's nice to "know" that it will never have an invalid or undefined state)

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      PyObjectPtr codes_(PyObject_GetAttrString(
+        PyObject_GetAttrString(
+          PyObject_GetAttrString(series, "cat"),
+          "codes"),
+        "values"));
+      SEXP codes = py_to_r(codes_, true);
+      
+      // get levels and convert to R
+      PyObjectPtr levels_(PyObject_GetAttrString(
+        PyObject_GetAttrString(dtype, "categories"),
```
kevinushey: Similar unwrapping of calls here.

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+    // special treatment for pd.Categorical
+    if (as_std_string(dtype_str) == "category") {
+      
+      // get actual values and convert to R
+      PyObjectPtr codes_(PyObject_GetAttrString(
+        PyObject_GetAttrString(
+          PyObject_GetAttrString(series, "cat"),
+          "codes"),
+        "values"));
+      SEXP codes = py_to_r(codes_, true);
```
kevinushey: If you're working with raw SEXPs, then they need to be protected. Consider using RObject, or `Shield<SEXP>`, to get that protection for free.

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      
+      // get levels and convert to R
+      PyObjectPtr levels_(PyObject_GetAttrString(
+        PyObject_GetAttrString(dtype, "categories"),
+        "values"));
+      SEXP levels = py_to_r(levels_, true);
```
kevinushey: SEXP protection?

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      
+      // get "ordered" attribute
+      PyObjectPtr ordered_(PyObject_GetAttrString(dtype, "ordered"));
+      SEXP ordered = py_to_r(ordered_, true);
```
kevinushey: SEXP protection?

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+   
+      // populate integer vector to hold factor values
+      int* codes_int = INTEGER(codes);
```
kevinushey: I think you could take advantage of Rcpp here: just use `IntegerVector factor = codes + 1`. Does that work?

kevinushey: You should also validate that this truly is an integer vector (`INTSXP`) (and may consider handling numeric vectors as well (`REALSXP`)).

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      int n = Rf_length(codes);
+      IntegerVector factor(n);
+      // values need to start at 1
+      for (int i = 0; i < n; ++i) {
+        factor[i] = codes_int[i] + 1;
+      }
+
+      // populate character vector to hold levels
+      n = Rf_length(levels);
```
kevinushey: This seems like you could just immediately convert `levels` to a CharacterVector.

src/python.cpp
kevinushey: Or consider an appropriate Rcpp object.

kevinushey: Or, can you extract the underlying value using Python APIs directly?

src/python.cpp
kevinushey: Or consider an Rcpp object (e.g. RObject) so the underlying R object would automatically be protected for you.

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      CharacterVector levels_chr(n);
+      for (int i = 0; i < n; ++i) {
+        levels_chr[i] = STRING_ELT(levels, i);
+      }
+
+      factor.attr("class") = "factor";
+      factor.attr("levels") = levels_chr;
+      if (*(LOGICAL(ordered)) == 1) factor.attr("ordered") = ordered;
```
kevinushey: I think if you avoid converting from Python to R, you could likely use an appropriate Python API to get the underlying value out of the "ordered" attribute.

kevinushey:
> I could also imagine that we'd want to split out these individual functions, and/or add them as branches to py_to_r?

I agree it would be worth adding these as separate functions, to avoid making `py_convert_pandas_df` too large. A convention I often use is to just append type names or scope names to the "parent" function name; e.g. `py_convert_pandas_df_float64`; we could imagine a family of functions like this as needed.

skeydan:
Thanks a lot for the review Kevin! I'll go through the points and also, do some refactoring ...

> That said, py_to_r() already does a lot of this (with e.g. handlers for NumPy already there) so it may ultimately be unnecessary for the base types.

Yeah, I think what I had in mind was more like, for the `pandas`-specific types, starting from `Series` as every column is a `Series` ...

I just did a simple comment-out benchmark on `iris` where I just commented on these blocks in `conversion.R`


```
converted <- lapply(seq_along(columns) - 1L, function(i) {
    column <- columns[[i]]
    py_to_r(py_get_item(x, column)$values)
  })
```

resp.

```
converted <- py_convert_pandas_df(x)
```

And this is what I get (`py_to_r` only):

EXISTING:
```
Unit: milliseconds
        expr      min       lq     mean   median       uq
 py_to_r(y) 20.26531 22.99074 28.12521 24.05912 26.13646
      max neval
 154.3019  1000
```

resp.

NEW:

```
Unit: milliseconds
        expr      min       lq     mean   median       uq
 py_to_r(py) 5.312818 5.874213 6.492574 6.219017 6.888228
      max neval
 20.22062  1000
```

Looks like it makes sense to continue :-)

src/python.cpp
```diff
@@ -2207,3 +2207,91 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      sexp = factor;
+    
+    // default case
+    } else {
+      
+      sexp = py_to_r(values, obj.convert());
+      
+    }
+    
+    list.push_back(sexp);
```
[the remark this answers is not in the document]
skeydan: Oh, now I see why `py_iterate` (which I used as a starting point) does this :-)
I don't see a difference in time or memory usage for `iris` (doing 1000 conversions), but that's still just a tiny object... so changing in any case

src/python.cpp
skeydan: check

src/python.cpp
skeydan: done

src/python.cpp
skeydan: Using Rcpp in all cases now, this will be more readable than lots of protection/unprotection I guess...

src/python.cpp
skeydan: Regarding the addition, I tried that but got a compilation error. With the changes I made in the meantime (or even before), I do not get that same error anymore but a static assertion error, landing here (and feeling very bad about it ;-)):

```
/**
         * This is the worst case :
         * - not a primitive
         * - not implicitely convertible tp SEXP
         * - not iterable
         *
         * so we just give up and attempt to use static_assert to generate
         * a compile time message if it is available, otherwise we use
         * implicit conversion to SEXP to bomb the compiler, which will give
         * quite a cryptic message
         */
	template <typename T>
        inline SEXP wrap_dispatch_unknown_iterable(const T& object, ::Rcpp::traits::false_type) {
            RCPP_DEBUG_1("wrap_dispatch_unknown_iterable<%s>(., false )", DEMANGLE(T))
	    // here we know that T is not convertible to SEXP
	    #ifdef HAS_STATIC_ASSERT
                static_assert(!sizeof(T), "cannot convert type to SEXP");
```

This is for the current code (I renamed a lot of things, trying to get something self-consistent which wasn't so easy with all the local variables):

```
RObject R_values = py_to_r(code_values, true);
int* codes_int = INTEGER(R_values);
int n = Rf_length(R_values);
IntegerVector factor(n);
factor = codes_int + 1;
```
If I try the addition on the `RObject` instead, like so

```
RObject R_values = py_to_r(code_values, true);
int* codes_int = INTEGER(R_values);
int n = Rf_length(R_values);
IntegerVector factor(n);
factor = R_values + 1;
```

I instead get 

```
invalid use of incomplete type "struct sexprec"
```

skeydan: Regarding the second point, I think I'm not clear about the degree to which to perform error checking...  

In this case, the `pandas` docs say that `codes` are integers: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Categorical.codes.html#pandas.Categorical.codes

My impression was we could rely on this... Also, if we don't, I think I'd also - or even more - need to check whether the complicated (to me!) substructure of Categoricals (see the previously nested accessors to get values and levels...) has changed between `pandas` releases, which could end up in a lot of error checking (which perhaps I should do?)

src/python.cpp
skeydan: Did you mean like so:

```
      RObject R_levels = py_to_r(category_values, true);
      CharacterVector factor_levels(R_levels);
```
?

I tried it and it compiles, but the `levels` attribute has the wrong class:

```
# after roundtripping iris
> attributes(after$Species)$levels %>% class()
[1] "array"
> attributes(before$Species)$levels %>% class()
[1] "character"
```

For sure there is something one could here but I don't really know what ... (the loop was the only thing that seemed to work out)

src/python.cpp
skeydan: oh I see! Replaced by

```
if (PyObject_IsTrue(ordered)) factor.attr("ordered") = true;
```

src/python.cpp
skeydan: see above

skeydan:
Factored out Series conversion and I think I'd  do the same for Categorical.
But I'll wait to see if this ok:

```
PyObjectPtr series(PySequence_GetItem(tuple, 1));
Py_IncRef(series);
    
 // delegate to py_convert_pandas_series
PyObjectRef series_ref(series, df.convert());
    
RObject R_obj = py_convert_pandas_series(series_ref);
    
Py_DecRef(series);
```

Thanks!

src/python.cpp
kevinushey: We likely need to drop attributes on the generated vector (specifically the `dim` attribute?)

src/libpython.h
```diff
@@ -307,6 +307,10 @@ LIBPYTHON_EXTERN void (*PySys_SetArgv_v3)(int, wchar_t **);
 
 LIBPYTHON_EXTERN void (*PySys_WriteStderr)(const char *format, ...);
 
+LIBPYTHON_EXTERN PyObject* (*PyObject_CallMethod)(PyObject *o, const char *name, const char *format, ...);
+LIBPYTHON_EXTERN PyObject* (*PySequence_GetItem)(PyObject *o, Py_ssize_t i);
+LIBPYTHON_EXTERN int (*PyObject_IsTrue)(PyObject *o);
```
kevinushey: As a side note, we might need to double-check whether this is available with Python 2. (Some API functions are defined as macros rather than true functions and in that case attempts to load the associated routine would fail)

src/python.cpp
```diff
@@ -2207,3 +2207,108 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+    PyObjectPtr category_values(PyObject_GetAttrString(categories, "values"));
+    RObject R_levels = py_to_r(category_values, true);
+    
+    // get "ordered" attribute
+    PyObjectPtr ordered(PyObject_GetAttrString(dtype, "ordered"));
+    //RObject ordered = py_to_r(ordered_, true);
+    
+    // populate integer vector to hold factor values
+    int* codes_int = INTEGER(R_values);
+    int n = Rf_length(R_values);
```
kevinushey: nit: may want to use `Rf_xlength(R_values)` so that long vectors can be supported.

src/python.cpp
```diff
@@ -2207,3 +2207,108 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+  return R_obj;
+  
+}
+
+// [[Rcpp::export]]
+SEXP py_convert_pandas_df(PyObjectRef df) {
+  
+  // pd.DataFrame.items() returns an Iterator over (column name, Series) pairs
+  PyObjectPtr items(PyObject_CallMethod(df, "items", NULL));
+  if (! PyObject_HasAttrString(items, "__next__"))
```
kevinushey: I think you also need to check for `"next"`, for the Python 2 case. (Either existing should be fine.)

src/python.cpp
```diff
@@ -2207,3 +2207,108 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+    if (tuple.is_null()) {
+      if (PyErr_Occurred())
+        stop(py_fetch_error());
+      else
+        break;
+    }
+    
+    // access Series in slot 1
+    PyObjectPtr series(PySequence_GetItem(tuple, 1));
+    Py_IncRef(series);
```
kevinushey: The downside with using `Py_IncRef()` directly here is that, if a C++ exception happened to be emitted before we reached the `Py_DecRef()` call following, we'd end up leaking the object.

Do we need to call `Py_IncRef()` here? I think the object is basically protected by virtue of living within a `PyObjectPtr` already; this may be unnecessary.

src/python.cpp
```diff
@@ -2207,3 +2207,108 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+    int* codes_int = INTEGER(R_values);
+    int n = Rf_length(R_values);
+    IntegerVector factor(n);
+    // values need to start at 1
+    for (int i = 0; i < n; ++i) {
+      factor[i] = codes_int[i] + 1;
+    }
+    
+    // populate character vector to hold levels
+    n = Rf_length(R_levels);
```
kevinushey: Similar: do you think we could use `Rf_xlength()`?

src/python.cpp
```diff
@@ -2207,3 +2207,108 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+    PyObjectRef series_ref(series, df.convert());
+    
+    RObject R_obj = py_convert_pandas_series(series_ref);
+    
+    Py_DecRef(series);
+    
+    list.push_back(R_obj);
+  }
+
+  List rList(list.size());
```
kevinushey: I wonder if you can skip the for loop below and use something like `List rList(list.begin(), list.end())`? (If this doesn't work then feel free to discard; the code below is correct but I think we do allow objects to be constructed from iterators in some cases)

src/python.cpp
skeydan: Thanks, will do!

src/python.cpp
skeydan: done!

src/python.cpp
skeydan: that works great, thank you!

src/python.cpp
skeydan: I did this in response to  a segfault I got... but now I see that if I remove those statements, I still get the segfault. It does not (or not reliably) occur after a single roundtrip, but can be produced by 

```
before <- iris
p <- r_to_py(before)

microbenchmark::microbenchmark(
  py_to_r(p),
  times = 100
)
```

I'll debug this after pushing the other changes.
Just quickly noting 

```
(gdb) bt
#0  0x00007f48963cf63b in PyObject_GetAttr () from /usr/lib64/libpython3.7m.so
...
#153 0x00007f48963df167 in PyObject_GetAttrString () from /usr/lib64/libpython3.7m.so
#154 0x00007f48b04e3366 in py_convert_pandas_series (series=...) at reticulate_types.h:38

(gdb) frame 154
#154 0x00007f48b04e3366 in py_convert_pandas_series (series=...) at reticulate_types.h:38
38	    return get();
```

src/libpython.h
skeydan: I just checked, all exist in Python 2.7
(in one case, argument types changed from `char *` to `const char *`  between two Python 3 versions, but that was all)

src/python.cpp
skeydan: that works great, thanks!

skeydan:
Still don't know about the segfault (using the latest commit, https://github.com/rstudio/reticulate/pull/620/commits/5415a0cdcd5db41a2aa7b4590d1143a184002818)...

Repeated executions of the second line below

```
before <- iris
 after  <- py_to_r(r_to_py(before))
```
don't trigger it.

But I can produce it by consecutive executions of `microbenchmark` with number of executions = 1.

If I execute the following

```
before <- iris
p <- r_to_py(before)
microbenchmark::microbenchmark(
  py_to_r(p),
  times = 1
)
microbenchmark::microbenchmark(
  py_to_r(p),
  times = 1
)
```

this is what happens:

```
#0  0x00007f39a6b5ee65 in PyObject_GetAttrString (v=0x7f399a31b500, name=name@entry=0x7f39c010c4a1 "dtype") at Objects/object.c:835
835	    if (Py_TYPE(v)->tp_getattr != NULL)

#1  0x00007f39c00ff317 in py_convert_pandas_series (series=...) at reticulate_types.h:38
#2  0x00007f39c00ffee7 in py_convert_pandas_df (df=...) at /home/key/R/x86_64-redhat-linux-gnu-library/3.6/Rcpp/include/RcppCommon.h:120

```

If I recreate the converted iris object in Python every time, like so

```
test <- function() {
  before <- iris
  p <- r_to_py(before)
  py_to_r(p)
}

microbenchmark::microbenchmark(
  test(),
  times = 1
)

```

I get an assertion failure for negative refcount:

```
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f0850f50e35 in raise () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7f084c8e8ec0 (LWP 6019))]
Missing separate debuginfos, use: dnf debuginfo-install rstudio-1.3.470-1.x86_64
(gdb) bt
#0  0x00007f0850f50e35 in raise () from /lib64/libc.so.6
#1  0x00007f0850f3b895 in abort () from /lib64/libc.so.6
#2  0x00007f082dc2d768 in fatal_error (prefix=prefix@entry=0x0, msg=msg@entry=0x7f082dce36e3 "_PyObject_AssertFailed", status=status@entry=-1) at Python/pylifecycle.c:2187
#3  0x00007f082dc2ff45 in Py_FatalError (msg=msg@entry=0x7f082dce36e3 "_PyObject_AssertFailed") at Python/pylifecycle.c:2197
#4  0x00007f082db3bc47 in _PyObject_AssertFailed (obj=0x7f0821291230, expr=expr@entry=0x0, msg=msg@entry=0x7f082dce3737 "object has negative ref count", 
    file=file@entry=0x7f082dcc88ad "./Include/object.h", line=line@entry=541, function=function@entry=0x7f082dce4860 <__func__.14800> "_Py_NegativeRefcount") at Objects/object.c:2199
#5  0x00007f082db3be08 in _Py_NegativeRefcount (filename=filename@entry=0x7f082dcc88ad "./Include/object.h", lineno=lineno@entry=541, op=<optimized out>) at Objects/object.c:232
#6  0x00007f082db2236d in _Py_DECREF (op=<optimized out>, lineno=541, filename=0x7f082dcc88ad "./Include/object.h") at ./Include/object.h:473
#7  _Py_XDECREF (op=<optimized out>) at ./Include/object.h:541
#8  free_keys_object (keys=keys@entry=0x7f0821315f40) at Objects/dictobject.c:580
#9  0x00007f082db23c84 in dictkeys_decref (dk=0x7f0821315f40) at Objects/dictobject.c:324
#10 dict_dealloc (mp=0x7f0821286650) at Objects/dictobject.c:1994
#11 0x00007f082db3b534 in _Py_Dealloc (op=<optimized out>) at Objects/object.c:2214
#12 0x00007f082db23b48 in _Py_DECREF (op=<optimized out>, lineno=541, filename=0x7f082dcc88ad "./Include/object.h") at ./Include/object.h:478
#13 _Py_XDECREF (op=<optimized out>) at ./Include/object.h:541
#14 dict_dealloc (mp=0x7f08212861d0) at Objects/dictobject.c:1986
#15 0x00007f082db3b534 in _Py_Dealloc (op=<optimized out>) at Objects/object.c:2214
#16 0x00007f082db55f92 in _Py_DECREF (op=<optimized out>, lineno=1272, filename=0x7f082dce79c3 "Objects/typeobject.c") at ./Include/object.h:478
#17 subtype_dealloc (self=0x7f082127eeb0) at Objects/typeobject.c:1272
#18 0x00007f082db3b534 in _Py_Dealloc (op=<optimized out>) at Objects/object.c:2214
#19 0x00007f082db3be55 in _Py_DECREF (op=<optimized out>, lineno=541, filename=0x7f082dcc88ad "./Include/object.h") at ./Include/object.h:478
#20 _Py_XDECREF (op=<optimized out>) at ./Include/object.h:541
#21 Py_DecRef (o=<optimized out>) at Objects/object.c:247
```

Not sure at the moment what is going on...

kevinushey:
The segfault seems to indicate the crash is happening while de-allocating a dictionary object, if I'm reading it correctly?

skeydan:
No solution; just want to keep track of investigation...

1)

With the current code, the `Series` passed to `PyObject_GetAttrString` has `ob_type` NULL.
```
(gdb) p *v
$1 = {ob_refcnt = 3, ob_type = 0x0}
```
This should never happen (the object itself still has refcount = 3). No recent (3.8) bugs found...

2) 

Tried what happens when I add an additional `Py_IncRef` (which should not be needed):

```
 PyObjectPtr series(PySequence_GetItem(tuple, 1));
    Py_IncRef(series);
    // delegate to py_convert_pandas_series
    PyObjectRef series_ref(series, df.convert());
    RObject R_obj = py_convert_pandas_series(series_ref);
    Py_DecRef(series);
```

Now refcount is 8 (why not 4?) and `ob_type` has an **invalid** address:

```
#0  0x00007ff9c077ce65 in PyObject_GetAttrString (v=0x7ff99eee80a0, name=name@entry=0x7ff9c0ed54a1 "dtype") at Objects/object.c:835
835	    if (Py_TYPE(v)->tp_getattr != NULL)

(gdb) bt
#0  0x00007ff9c077ce65 in PyObject_GetAttrString (v=0x7ff99eee80a0, name=name@entry=0x7ff9c0ed54a1 "dtype") at Objects/object.c:835
#1  0x00007ff9c0ec8327 in py_convert_pandas_series (series=...) at reticulate_types.h:38
...
...
(gdb) p (PyObject*)(v)
$2 = (PyObject *) 0x7ff99eee80a0
(gdb) p (PyObject*)(v->ob_type)
$3 = (PyObject *) 0x1e30e84eb67f2d30
(gdb) p *((PyObject*)(v->ob_type))
Cannot access memory at address 0x1e30e84eb67f2d30

```

skeydan:
Not sure if this is may be related: https://bugs.python.org/issue38006
(((Some relation to weak refs cleanup? 
From the assertion failure case, at least the corresponding code is called from 

```
#22 0x00007f11b30f7961 in R_RunWeakRefFinalizer () from /usr/lib64/R/lib/libR.so
```
)))
Can try if a build from master helps...

skeydan:
New shortest reprex:

```
pd <- import("pandas")
before <- iris
after  <- py_to_r(r_to_py(before))
gc()
```

```#0  0x00007f67fe3f3e35 in raise () from /lib64/libc.so.6
#1  0x00007f67fe3de895 in abort () from /lib64/libc.so.6
#2  0x00007f67f44ab768 in fatal_error (prefix=prefix@entry=0x0, msg=msg@entry=0x7f67f45616e3 "_PyObject_AssertFailed", status=status@entry=-1) at Python/pylifecycle.c:2187
#3  0x00007f67f44adf45 in Py_FatalError (msg=msg@entry=0x7f67f45616e3 "_PyObject_AssertFailed") at Python/pylifecycle.c:2197
#4  0x00007f67f43b9c47 in _PyObject_AssertFailed (obj=0x7f67ce922f00, expr=expr@entry=0x0, msg=msg@entry=0x7f67f4561737 "object has negative ref count", 
    file=file@entry=0x7f67f45468ad "./Include/object.h", line=line@entry=541, function=function@entry=0x7f67f4562860 <__func__.14800> "_Py_NegativeRefcount") at Objects/object.c:2199
#5  0x00007f67f43b9e08 in _Py_NegativeRefcount (filename=filename@entry=0x7f67f45468ad "./Include/object.h", lineno=lineno@entry=541, op=<optimized out>) at Objects/object.c:232
#6  0x00007f67f43a036d in _Py_DECREF (op=<optimized out>, lineno=541, filename=0x7f67f45468ad "./Include/object.h") at ./Include/object.h:473
#7  _Py_XDECREF (op=<optimized out>) at ./Include/object.h:541
#8  free_keys_object (keys=keys@entry=0x7f67ce9b9f40) at Objects/dictobject.c:580
#9  0x00007f67f43a1c84 in dictkeys_decref (dk=0x7f67ce9b9f40) at Objects/dictobject.c:324
#10 dict_dealloc (mp=0x7f67ce92f470) at Objects/dictobject.c:1994
#11 0x00007f67f43b9534 in _Py_Dealloc (op=<optimized out>) at Objects/object.c:2214
#12 0x00007f67f43a1b48 in _Py_DECREF (op=<optimized out>, lineno=541, filename=0x7f67f45468ad "./Include/object.h") at ./Include/object.h:478
#13 _Py_XDECREF (op=<optimized out>) at ./Include/object.h:541
#14 dict_dealloc (mp=0x7f67ce91fa70) at Objects/dictobject.c:1986
#15 0x00007f67f43b9534 in _Py_Dealloc (op=<optimized out>) at Objects/object.c:2214
#16 0x00007f67f43d3f92 in _Py_DECREF (op=<optimized out>, lineno=1272, filename=0x7f67f45659c3 "Objects/typeobject.c") at ./Include/object.h:478
#17 subtype_dealloc (self=0x7f67ce922c80) at Objects/typeobject.c:1272
#18 0x00007f67f43b9534 in _Py_Dealloc (op=<optimized out>) at Objects/object.c:2214
#19 0x00007f67f43b9e55 in _Py_DECREF (op=<optimized out>, lineno=541, filename=0x7f67f45468ad "./Include/object.h") at ./Include/object.h:478
#20 _Py_XDECREF (op=<optimized out>) at ./Include/object.h:541
#21 Py_DecRef (o=<optimized out>) at Objects/object.c:247
#22 0x00007f67fed90961 in R_RunWeakRefFinalizer () from /usr/lib64/R/lib/libR.so
```

src/python.cpp
```diff
@@ -2207,3 +2207,98 @@ SEXP py_eval_impl(const std::string& code, bool convert = true) {
...
+      if (PyErr_Occurred())
+        stop(py_fetch_error());
+      else
+        break;
+    }
+    
+    // access Series in slot 1
+    PyObjectPtr series(PySequence_GetItem(tuple, 1));
+    // delegate to py_convert_pandas_series
+    PyObjectRef series_ref(series, df.convert());
```
kevinushey: Now that I think about it, because this PyObjectRef will now "own" the previously-created `series` object, I believe you will have to use `series.detach()` here. See e.g.

https://github.com/rstudio/reticulate/blob/86acdf16f9208cf4952b1a312bccdc91f7a25bdf/src/python.cpp#L2202-L2205

kevinushey: Hmm, but we aren't actually returning this reference back to R -- we're only using it temporarily while we convert that object back into an R object...

kevinushey: Actually yes -- I believe this is indeed the issue. PyObjectRef will also call Py_DecRef once the associated R object is cleaned up by the garbage collector; e.g. by

https://github.com/rstudio/reticulate/blob/86acdf16f9208cf4952b1a312bccdc91f7a25bdf/src/reticulate_types.h#L9-L13

So now we have two calls to `Py_DecRef()`, but only a single `Py_IncRef()` (implicitly through the construction of the `PyObjectPtr`). So `series.detach()` should indeed be the fix.

skeydan: That was it :-) :-)
thanks!!

(I hadn't really understood the `detach` so far, but now I have an impressive example... :-))

((yes and I had seen `py_ref` and checked if it did something that might make a difference in that respect, but as I saw it didn't and then it seemed like unnecessary overhead given the object was just a temporary construction)))

skeydan:
Hi Kevin, 

added `pd.Timestamp`  conversion (that hopefully will work on other platforms as well?), plus a test :-)

When a `pd.Timestamp`  has a time zone, it will be taken into account but Rcpp will return the time in UTC. 
So if we import `pandas` with `convert = FALSE` and compare the raw before/after following a roundtrip, the display will look different. However, when we call `py_to_r` on both, they will both be displayed in _local_ timezone and will be equal.

Does that look ok to you?

Regarding this:

```
DatetimeVector R_posixct(posixct.size());
    for (int i = 0; i < posixct.size(); ++i) {
      R_posixct[i] = posixct[i];
    }
```
I tried the better-looking 

```
DatetimeVector R_posixct(posixct.begin(), posixct.end())
```
resp.

```
DatetimeVector R_posixct(posixct);
```

but none of those compile...

PR #628, opened 2019-11-01
allow packages to declare specific versions of dependencies

kevinushey:
This PR makes it possible for R packages to declare specific versions of Python dependencies. R packages can hence declare e.g.

```
reticulate@R:
    list(
      packages = list(
        list(package = "scipy", version = "1.3.0")
      )
    )
```

to indicate that the package depends on `scipy [1.3.0]`.

---

- [x] Should we do any sort of (primitive) dependency resolution?
- [x] What should we do if multiple loaded R packages declare conflicting requirements (e.g. a different package requires `scipy [1.2.1]`?

### DIFF
--- a/vignettes/python_dependencies.Rmd
+++ b/vignettes/python_dependencies.Rmd
@@ -75,7 +75,7 @@ Version: 1.0.0
 Description: Provides an R interface to the Python package scipy.
 reticulate@R: list(
   packages = list(
-    c(package = "scipy")
+    list(package = "scipy")
   )
 )
 < ... other fields ...>
--- a/vignettes/python_dependencies.Rmd
+++ b/vignettes/python_dependencies.Rmd
@@ -77,7 +77,7 @@ reticulate@R: list(
   packages = list(
     list(package = "scipy")
   )
-)
+  )
 < ... other fields ...>
 ```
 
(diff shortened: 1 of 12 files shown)


vignettes/python_dependencies.Rmd
```diff
@@ -152,18 +153,22 @@ Declared Python package dependencies should have the following format:
 
 - **package**: The name of the Python package.
 
+- **version**: The version of the package that should be installed.
```
jjallaire: We should emphasize that this should only be used in cases where it's known that more recent versions have breakage. As such, conflicts here will be resolved in favor of the older version. People should just understand that in order to ensure the broadest compatibility this should work much like CRAN (where you always get the latest version and it's assumed they work / work together) and this is an exceptional escape hatch for when that isn't the case.

jjallaire:
We are purposely steering around allowing packages to make finicky version
requests because that will inevitably lead to conflicts (as R doesn't have
an npm-type system where every package can have their own set of
dependencies). This is designed to pin an older version but not necessary
provide ranges or allow specification of "only this version or newer"

terrytangyuan:
@jjallaire I see. That makes sense. As long as users are aware of this it should be fine.

kevinushey:
It looks like a dependency resolver may be coming to pip soon:

https://github.com/pypa/pip/issues/988
https://pradyunsg.me/blog/2019/06/23/oss-update-1/

When that happens (and the newer versions of `pip` are sufficiently common across commonly-used operating systems), we could allow version specifications and just defer to `pip` to figure it out for us.

NeolithEra:
Hi, maybe you can try a dependency diagnosis website http://www.watchman-pypi.com/dependency_analysis, which can help diagnoze dependency conflict problems and automatically visualize your dependency tree and those of your downstream projects.


Best,
Neolith

skeydan:
I only now saw we have a vignette already: https://github.com/rstudio/reticulate/blob/master/vignettes/python_dependencies.Rmd :-)

Here:
https://github.com/rstudio/reticulate/blob/2796ef69d1b842385be91663cf0518f18be7e080/vignettes/python_dependencies.Rmd#L160

as a reader, I would be wondering "who wins" - does the first package loaded get to keep its version, and the other package gets loaded but will have the "wrong" version?
("Warning" sounds like we don't abort)

PR #629, opened 2019-11-04
avoid compiler warning

skeydan:
hadn't seen this warning before or else this should have been part of the last Pandas-related PR :-)


kevinushey:
Thanks!

PR #630, opened 2019-11-04
WIP: Python 3.9 compatibility

skeydan:
This is just the first step reg. `Py_CompileString`, Python is successfully loaded for Python 3.9 (including 3.8 ... 3.2 - `Py_CompileStringExFlags` was added in 3.2)  as well as Python 2.7 now. 

But I haven't yet found a way to use the adequate function in `py_eval_impl`:

dummy code:

```
  if (FALSE)
    compiledCode(Py_CompileString(code.c_str(), "reticulate_eval", Py_eval_input));
  else 
    compiledCode(Py_CompileStringExFlags(code.c_str(), "reticulate_eval", Py_eval_input, NULL, 0));
```

I was searching for  a way to check if a function exists in cpython, but haven't found one yet... (tried e.g. to see if the module dict from `main` has a key with that name)

Just copying the macro from 

https://github.com/python/cpython/blob/da6ce58dd5ac109485af45878fca6bfd265b43e9/Include/pythonrun.h#L99-L103

is not an option because only one of them will be loaded, depending on Python version...

Would you have an idea what to do here Kevin? Thanks!

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2187,12 +2187,12 @@ SEXP py_run_file_impl(const std::string& file,
 SEXP py_eval_impl(const std::string& code, bool convert = true) {
 
   // compile the code
-  // TBD: find a check that works
   PyObjectPtr compiledCode;
-  if (FALSE)
-    compiledCode(Py_CompileString(code.c_str(), "reticulate_eval", Py_eval_input));
+  if (Py_CompileStringExFlags != nullptr) 
+    compiledCode.assign(Py_CompileStringExFlags(code.c_str(), "reticulate_eval", Py_eval_input, NULL, 0));
   else 
-    compiledCode(Py_CompileStringExFlags(code.c_str(), "reticulate_eval", Py_eval_input, NULL, 0));
+    compiledCode.assign(Py_CompileString(code.c_str(), "reticulate_eval", Py_eval_input));
+  
   
   if (compiledCode.is_null())
     stop(py_fetch_error());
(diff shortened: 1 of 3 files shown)


src/python.cpp
```diff
@@ -2187,10 +2187,16 @@ SEXP py_run_file_impl(const std::string& file,
 SEXP py_eval_impl(const std::string& code, bool convert = true) {
 
   // compile the code
-  PyObjectPtr compiledCode(Py_CompileString(code.c_str(), "reticulate_eval", Py_eval_input));
+  // TBD: find a check that works
+  PyObjectPtr compiledCode;
+  if (FALSE)
+    compiledCode(Py_CompileString(code.c_str(), "reticulate_eval", Py_eval_input));
```
kevinushey: You should be able to just check e.g.

```
if (Py_CompileString != nullptr) { ... }
```

skeydan:
oops, thanks, that sounds like one should have had the idea oneselves :-(
(unfortunately all tests I was thinking of resulted in an actual attempt to load the symbol ...)

so now it works with 2.7 as well as 3.9 ...

skeydan:
seems like R builds < 3.6 are built with c++ < 11 so changed `nullptr` to `NULL`...

kevinushey:
You're right -- `NULL` rather than `nullptr` is preferred for compatibility with more compilers.

kevinushey:
Thanks!

PR #635, opened 2019-11-05
[WIP] Allows creating/subclassing Python classes

dfalbel:
This PR allows creating Python classes using the following syntax:

``` r
library(reticulate)
Hi <- PyClass("Hi", list(
  name = NULL,
  `__init__` = function(self, name) {
    self$name <- name
    NULL
  },
  say_hi = function(self) {
    paste0("Hi ", self$name)
  }
))

hi <- Hi("World")
hi
#> <Hi>
hi$say_hi()
#> [1] "Hi World"
```

One possible use case for this is subclassing callbacks in Keras. The [following](https://stackoverflow.com/questions/49127214/keras-how-to-output-learning-rate-onto-tensorboard/50040233#50040233) is currently not possible in R:

```python
from keras import backend as K
from keras.callbacks import TensorBoard

class LRTensorBoard(TensorBoard):
    def __init__(self, log_dir, **kwargs):  # add other arguments to __init__ if you need
        super().__init__(log_dir=log_dir, **kwargs)

    def on_epoch_end(self, epoch, logs=None):
        logs.update({'lr': K.eval(self.model.optimizer.lr)})
        super().on_epoch_end(epoch, logs)
```

And with this PR would be implemented as following:

```r
LRTensorBoard <- PyClass(
  classname = "LRTensorBoard", 
  inherit = tensorflow::tf$keras$callbacks$TensorBoard,
  list(
    `__init__` = function(self, log_dir, ...) {
      super()$`__init__`(log_dir = log_dir, ...)
    },
    on_epoch_end = function(self, epoch, logs = NULL) {
      lr <- self$model$optimizer$lr$numpy()
      print(lr)
      logs$update(list("lr" = lr))
      super()$on_epoch_end(epoch, logs)
    }
  )
)
```


R/class.R
```diff
@@ -0,0 +1,43 @@
+inject_super <- function(fun) {
+  e <- new.env()
```
kevinushey: nit: prefer explicitly setting `parent = parent.frame()` to make it clear that this environment should inherit from the parent frame?

R/class.R
```diff
@@ -0,0 +1,43 @@
...
+  e$super <- function() {
+    bt <- reticulate::import_builtins()
+    self <- as.list(parent.frame())$self
```
kevinushey: Prefer instead using:

```
get("self", envir = parent.frame(), inherits = FALSE)
```

As this would avoid attempting to evaluate other symbols defined in the parent frame.

R/class.R
kevinushey: Or should this have `environment(fun)` as the parent environment?

R/class.R
```diff
@@ -0,0 +1,43 @@
...
+  environment(e$super) <- environment(fun)
+  environment(fun) <- e
+  fun
+}
+
+#' Create a python class
+#' 
+#' @param classname The classname
+#' @param defs The definitions 
+#' @param inherit List of python classes
```
kevinushey: Does this accept Python classes by name? Or do they need to be actual Python class objects?

R/class.R
```diff
@@ -0,0 +1,43 @@
...
+  }
+  environment(e$super) <- environment(fun)
+  environment(fun) <- e
+  fun
+}
+
+#' Create a python class
+#' 
+#' @param classname The classname
+#' @param defs The definitions
```
kevinushey: Can this be empty? (can you define an empty class?)

R/class.R
```diff
@@ -0,0 +1,43 @@
...
+    bt$super(self$`__class__`, self)
+  }
+  environment(e$super) <- environment(fun)
+  environment(fun) <- e
+  fun
+}
+
+#' Create a python class
+#' 
+#' @param classname The classname
```
kevinushey: Are there any restrictions on what class names are valid?

R/class.R
```diff
@@ -0,0 +1,43 @@
+inject_super <- function(fun) {
```
kevinushey: This function is fairly complicated: some comments would be helpful.

R/class.R
dfalbel: you are totally right! fixed

R/class.R
dfalbel: They need to be Python class objects

R/class.R
dfalbel: Yeah, passing `list()` would make an empty class. I made `defs = list()` the default.

R/class.R
dfalbel: It seems that it's controlled by this: https://docs.python.org/3/reference/lexical_analysis.html#identifiers meaning that almost any string is accepted

dfalbel:
Thanks very much for the review, Kevin!

R/class.R
```diff
@@ -0,0 +1,68 @@
...
+  
+  fun
+}
+
+#' Create a python class
+#' 
+#' @param classname Name of the class. The class name is useful for S3 method
+#'  dispatch.
+#' @param defs A named list of class definitions - functions, attributes, etc.
+#' @param inherit A list of Python class objects. Usually theese objects have
```
kevinushey: typo: `theese`

R/class.R
```diff
@@ -0,0 +1,68 @@
+inject_super <- function(fun) {
+  # for each function `fun` we need to place a `super` function in it's
```
skeydan:
```suggestion
  # for each function `fun` we need to place a `super` function in its
```

PR #643, opened 2019-11-13
use Rcpp to convert R Dates to Python

skeydan:
Hi Kevin,

is this acceptable (it adds `datetime.h`).

Execution time is about 1/4 for this: `dates <- replicate(100, Sys.Date(), simplify = FALSE)`

Old:

```
Unit: milliseconds
           expr      min       lq     mean   median       uq
 r_to_py(dates) 47.23566 51.19988 65.46687 53.35797 57.47948
      max neval
 1155.049   100
```

New:

```
Unit: milliseconds
           expr      min      lq     mean   median       uq      max
 r_to_py(dates) 11.89173 12.7865 14.36604 13.35968 16.22332 22.91708
 neval
   100
```

The code doesn't run on Python 2, but before investigating I wanted to ask if we can live with adding `datetime.h`... if yes, there may be more places we can use it ...

Thanks for reviewing! As an aside, how long are we going to continue to support Python 2?


R/conversion.R
```diff
@@ -159,20 +159,7 @@ py_to_r.datetime.datetime <- function(x) {
...
-    month <- as.integer(iso[[2]])
-    day <- as.integer(iso[[3]])
-    datetime$date(year, month, day)
-  })
-
-  if (length(items) == 1)
-    items[[1]]
-  else
-    r_to_py_impl(items, convert)
+  r_convert_date(x, convert)
```
kevinushey: This may need to handle length-one lists still (to ensure they're returned as scalars)

src/datetime.h
```diff
@@ -0,0 +1,259 @@
...
+
+/* Apply for time delta instances */
+#define PyDateTime_DELTA_GET_DAYS(o)         (((PyDateTime_Delta*)o)->days)
+#define PyDateTime_DELTA_GET_SECONDS(o)      (((PyDateTime_Delta*)o)->seconds)
+#define PyDateTime_DELTA_GET_MICROSECONDS(o)            \
+    (((PyDateTime_Delta*)o)->microseconds)
+
+
+/* Define structure for C API. */
+typedef struct {
```
kevinushey: Are we able to verify that the size / layout of this struct is stable across different versions of Python? (It seems to not be for Python 2.7, if I understand your comment on the PR correctly)

skeydan:
Hi Kevin,

thanks for reviewing! If I see correctly, the most recent changes to any `struct`s in `datetime.h` took place about 2 years ago: https://github.com/python/cpython/pull/5032...

Now I'm not really sure what to do - does this kind of volatility basically rule it out? Or should I try to test with different versions of Python 3 (would >= 3.5 suffice in that case?)

Reg. 2.7, I was wondering if one could just branch, `r_to_py.Date`, on major version (or else I'd investigate) - but the Python 2 question is really only relevant if that other point doesn't rule it out?

kevinushey:
> Now I'm not really sure what to do - does this kind of volatility basically rule it out? Or should I try to test with different versions of Python 3 (would >= 3.5 suffice in that case?)

The fact that this struct is not stable across versions does scare me, especially since it seems there's no attempt to maintain the struct layout as new versions of Python are released.

I think rather than attempting to import and use `datetime.h`, a safer approach would be to simply iterate and evaluate Python code from the C++ side. That is, we would still call back into "regular" Python evaluation, but just do so from the C++ side rather than through R (where there is more overhead). Does that sound reasonable to you?

skeydan:
Oh, yes, I think I see what you mean! Honestly I was assuming this (https://docs.python.org/3/c-api/datetime.html) would be the only way to do it using cpython ... but the alternative would be using `PyImport_ImportModule` to import `datetime`, and then go from there as usual?

kevinushey:
Yep, exactly -- I think we should be able to use `PyObject_CallMethod()`, after we get a reference to the `datetime.time` function (which would require `PyImport_ImportModule` and, I think, `PyObject_GetAttrString`)

skeydan:
Thanks! Then I'll change it like that tomorrow :-)

R/conversion.R
skeydan: Not sure exactly what you mean? I was trying to find what this
https://github.com/rstudio/reticulate/blob/64f685ab279516250ef15f2957820744c9841d32/R/conversion.R#L172
was for but didn't manage to get into the `else` branch...

skeydan:
Done! Looks like it even got a bit faster:

```
> microbenchmark(r_to_py(dates))
Unit: milliseconds
           expr      min      lq     mean   median       uq
 r_to_py(dates) 10.85329 11.3757 12.61398 11.81957 13.18172
      max neval
 19.31118   100
```

R/conversion.R
kevinushey: E.g.

```
> library(reticulate)
> r_to_py(Sys.Date())
2019-11-15
> r_to_py(c(Sys.Date(), Sys.Date()))
[datetime.date(2019, 11, 15), datetime.date(2019, 11, 15)]
```

That is, we make a Python list for R vectors of length > 1, and a 'scalar' for R vectors of length 1. Is this behavior still preserved in this PR?

src/python.cpp
```diff
@@ -2424,7 +2420,11 @@ PyObjectRef r_convert_dataframe(RObject dataframe, bool convert) {
 // [[Rcpp::export]]
 PyObjectRef r_convert_date(Date date, bool convert) {
   
-  PyObject* py_date = PyDate_FromDate(date.getYear(), date.getMonth(), date.getDay());
+  PyObject* datetime = PyImport_ImportModule("datetime");
+  PyObject* py_date = PyObject_CallMethod(
+    datetime, "date", "iii", date.getYear(), date.getMonth(), date.getDay());
```
kevinushey: Very pedantic nitpick: you may want to explicit cast `date.getYear()` and friends to `int`, e.g.

```
(int) date.getYear()
```

While the `getYear()` method does produce an `int`, in general, when using variadic functions, it's good practice to cast arguments to ensure they match the type expected / declared in the format string.

The code is correct as is, but I just want to call it out as a kind of best practice.

skeydan: done!

R/conversion.R
skeydan: Oh, got it, thanks! Does this look better now?

src/python.cpp
```diff
@@ -2415,3 +2416,45 @@ PyObjectRef r_convert_dataframe(RObject dataframe, bool convert) {
...
+    Date date = date_vector[0];
+    PyObjectPtr py_date(PyObject_CallMethod(
+        datetime, "date", "iii",
+        static_cast<int>(date.getYear()),
+        static_cast<int>(date.getMonth()),
+        static_cast<int>(date.getDay())));
+    if (py_date == NULL) {
+      stop(py_fetch_error());
+    }
+    return py_ref(py_date, convert);
```
kevinushey: Does this need `py_date.detach()`?

src/python.cpp
```diff
@@ -2415,3 +2416,45 @@ PyObjectRef r_convert_dataframe(RObject dataframe, bool convert) {
...
+      Date date = date_vector[i];
+      PyObjectPtr py_date(PyObject_CallMethod(
+          datetime, "date", "iii",
+          static_cast<int>(date.getYear()),
+          static_cast<int>(date.getMonth()),
+          static_cast<int>(date.getDay())));
+      if (py_date == NULL) {
+        stop(py_fetch_error());
+      }
+      int res = PyList_SetItem(list, i, py_date);
```
kevinushey: `PyList_SetItem` "steals" a reference, and so the associated `py_date` item should either not be wrapped in `PyObjectPtr`, or explicitly detached. See for example:

https://github.com/rstudio/reticulate/blob/64f685ab279516250ef15f2957820744c9841d32/src/python.cpp#L1309-L1313

kevinushey: (This is one of the frustrating bits of the Python API: most functions return new references, but some 'borrow' or 'steal' a reference and so have implications on how reference counting is performed)

skeydan: Oh sorry, I should have connected the dots ... (had read about e.g. `PyList_SetItem()` and also seen `PyObjectPtr*` in the piece of code I was taking as an example...)

src/python.cpp
skeydan: changed this one to `PyObject*` now too

Issue #649, opened 2019-11-19
Full paths to conda environment not supported by use_condaenv


prihoda:
This line in use_condaenv incorrectly uses `fixed=TRUE`, which means that full paths to conda environments are not supported since the regex does not match any path.

https://github.com/rstudio/reticulate/blob/565ba8f117f5fefd0a1038a522411b2cf2128e27/R/use_python.R#L73

PR #650, opened 2019-11-19
Fix conda full path regex

prihoda:
See https://github.com/rstudio/reticulate/issues/649


kevinushey:
Thanks for catching that!

PR #652, opened 2019-11-20
allow for __getitem__ access methods in weird subclasses of Dict

skeydan:


skeydan:
Some things to explain...

1) This does not eliminate the segfault if someone tries to use `$` access. The thing it allows to do is use `[`. (Instead of `py_get_item(save$mapped, "one")`, which could already be used now).

2) The other way round is not implemented (yet?) Honestly I don't think it's worth it, as it seems to be a pretty idiosyncratic thing the TF guys are doing... For example, I tried to construct a test like this

```
wrapt <- import("wrapt")
d <- dict(items = 1, apple = 42)
proxy = wrapt$ObjectProxy(d)
py_to_r(proxy)
```

but this just works like a normal `dict`...
We could wait if something like this comes up more often, and then perhaps take more action?

3) Here is the test code, not added to unit tests because of dependency on TF:

```
library(tensorflow)
save <- tf$train$Checkpoint()
save$listed <- list(tf$Variable(1))
save$listed$append(tf$Variable(2))
save$mapped <- dict(one = save$listed[0])
# this would always have worked, uses __getitem__
# reticulate::py_get_item(save$mapped, "one")
# this works with this PR
save$mapped['one']

# this works
reticulate::py_set_item(save$mapped, "two", save$listed[1])
# this would need additional action
save$mapped['two'] <- save$listed[1]
```

4) Had to use `PyObject*` here now (instead of `PyObjectPtr`)

```
PyObject* pyKey(r_to_py(key, dict.convert()));
```

but I would assume that it's ok?

src/python.cpp
```diff
@@ -1980,8 +1980,23 @@ PyObjectRef py_dict_impl(const List& keys, const List& items, bool convert) {
 
 // [[Rcpp::export]]
 SEXP py_dict_get_item(PyObjectRef dict, RObject key) {
-  PyObjectPtr pyKey(r_to_py(key, dict.convert()));
+  
+  PyObject* pyKey(r_to_py(key, dict.convert()));
   PyObject* item = PyDict_GetItem(dict, pyKey);
```
kevinushey: Shouldn't we perform `PyDict_Check(dict)` before we attempt treating the `dict` object as a builtin dictionary?

src/python.cpp
```diff
@@ -1980,8 +1980,23 @@ PyObjectRef py_dict_impl(const List& keys, const List& items, bool convert) {
 
 // [[Rcpp::export]]
 SEXP py_dict_get_item(PyObjectRef dict, RObject key) {
-  PyObjectPtr pyKey(r_to_py(key, dict.convert()));
+  
+  PyObject* pyKey(r_to_py(key, dict.convert()));
```
kevinushey: Doesn't changing this to `PyObject*` mean we risk leaking a reference to the Python object here?

src/python.cpp
kevinushey: It also looks like we might need `PyDict_CheckExact()` here. See:

https://github.com/python/cpython/blob/da6ce58dd5ac109485af45878fca6bfd265b43e9/Python/ceval.c#L2294-L2297

for example code in Python's sources.

src/python.cpp
```diff
@@ -1980,8 +1980,23 @@ PyObjectRef py_dict_impl(const List& keys, const List& items, bool convert) {
...
   PyObject* item = PyDict_GetItem(dict, pyKey);
+  
+  // some weird subclasses of Dict might only allow for __getitem__ access
+  // e.g. TensorFlow's _DictWrapper(TrackableDataStructure, wrapt.ObjectProxy)
+  if (item == NULL) {
+    if(!PyDict_Check(dict)) {
+      if (PyObject_HasAttrString(dict, "__getitem__")) {
+        item = PyObject_CallMethod(dict, "__getitem__", "O", pyKey);
```
kevinushey: You might be able to call `PyObject_GetItem()` directly here.

src/python.cpp
skeydan: The weird thing is, `PyDict_Check` is false for `OrderedDict`s ... (which it shouldn't be, right, unless `OrderedDict` is not a proper subclass?))

First, I had the logic the other way round, directly starting with `PyDict_Check` and the case where the check would return false ... but then, the `OrderedDict` test would fail (perhaps because of problems with `__getitem__` and keys that aren't strings?)

Anyway, so I reversed the logic and ended up with that slightly strange way of checking for `NULL` etc (which seemed error-prone to me though)

src/python.cpp
skeydan: no comment ;-) [--> goes hang herself]]

kevinushey:
I ended up taking a similar approach (+ tackled a number of related issues) in https://github.com/rstudio/reticulate/pull/653. Thanks!

Issue #646, opened 2019-11-14
segfault from py_dict_get_keys_as_str


skeydan:
... my talent to produce segfaults surfacing again ;-)

Code:

```
library(tensorflow)
save <- tf$train$Checkpoint()
save$listed <- list(tf$Variable(1))
save$listed$append(tf$Variable(2))
save$mapped <- dict(one = save$listed[0])
save$mapped["one"]
# is NULL
# try autocomplete to e.g. get "items"
save$ # crashes session
```

Backtrace:

```
#0  0x00007f28a0286031 in PyList_Size () from /usr/lib64/libpython3.7m.so
[Current thread is 1 (Thread 0x7f28c03ddec0 (LWP 5862))]
Missing separate debuginfos, use: dnf debuginfo-install rstudio-1.3.470-1.x86_64
(gdb) bt
#0  0x00007f28a0286031 in PyList_Size () from /usr/lib64/libpython3.7m.so
#1  0x00007f28a07093f3 in py_dict_get_keys_as_str (dict=...) at python.cpp:2020

```

I'll take a look at this next week after wrapping up a few other things I started :-)

skeydan:
The problem is that this object, even though created through `dict`, is not a `Dict`, it's a thing called `_DictWrapper`

https://github.com/tensorflow/tensorflow/blob/762b84bef3abe29e9c6a5b3dae3a22f1d44d1c4b/tensorflow/python/training/tracking/data_structures.py#L707

This is why one cannot use `PyDict_GetItem` on it. On the other hand, `__getitem__` would work fine.

```
# works fine
reticulate::py_get_item(save$mapped, "one")

# returns NULL (uses PyDict_GetItem)
save$mapped['one']
```

In Python, the equivalent

```
import tensorflow as tf
save = tf.train.Checkpoint()
save.listed = [tf.Variable(1.)]
save.listed.append(tf.Variable(2.))
save.mapped = {'one': save.listed[0]}
save.mapped['two'] = save.listed[1]
```
works fine; I guess that `[]` here just uses `__getitem__`/`__setitem__`?

One thing one could do perhaps would be to insert a check here

```
SEXP py_dict_get_item(PyObjectRef dict, RObject key) {

  if(!PyDict_Check(dict)) {
  // error?
  // or try using __getitem__ still, from here?
  }

  PyObjectPtr pyKey(r_to_py(key, dict.convert()));
  PyObject* item = PyDict_GetItem(dict, pyKey);
[...]

PR #653, opened 2019-11-20
better handle type mismatches in internal methods

kevinushey:
As per https://github.com/rstudio/reticulate/issues/646, some objects might inherit from the "base" Python types without actually remaining compatible with the constraints imposed by the Python C API.

This PR works around this issue by first attempting to use the type-specific method, and then falling back to the object-specific method (which usually involves executing the relevant user-defined method).

Closes https://github.com/rstudio/reticulate/issues/646.


skeydan:
Works great for the TF case! Autocomplete gives error, but no segfaults.

```
library(tensorflow)
save <- tf$train$Checkpoint()
save$listed <- list(tf$Variable(1))
save$listed$append(tf$Variable(2))
save$mapped <- dict(one = save$listed[0])

# get
save$mapped['one']
# correct output
# <tf.Variable 'Variable:0' shape=() dtype=float32, numpy=1.0>

# set
save$mapped['two'] <- save$listed[1]
save$mapped['two']
# correct output
# <tf.Variable 'Variable:0' shape=() dtype=float32, numpy=2.0>

# autocomplete
save$mapped$
#  Error in py_dict_get_keys_as_str(x) : 
#  SystemError: <class 'type'> returned a result with an error set

#Detailed traceback: 
#  File "/home/key/.virtualenvs/r-reticulate/lib64/python3.7/site-packages/tensorflow_core/python/training/tracking/data_structures.py", line 738, in __getattribute__
#if (hasattr(type(self), name)


```

kevinushey:
Is that error expected? Seems like TensorFlow is doing some very weird stuff now...

skeydan:
This is way deeper into TF than I normally look... but I guess we can safely assume the problem is (at least) more on their side than ours, given that it fails here...

https://github.com/tensorflow/tensorflow/blob/04493e0514c75526baf8e93a23cff35a15a893c5/tensorflow/python/training/tracking/data_structures.py#L737

```
 def __getattribute__(self, name):
    if (hasattr(type(self), name)
        and isinstance(getattr(type(self), name), property)):
      # Bypass ObjectProxy for properties. Whether this workaround is necessary
      # appears to depend on the Python version but not the wrapt version: 3.4
      # in particular seems to look up properties on the wrapped object instead
      # of the wrapper without this logic.
```

PR #655, opened 2019-11-21
Convert arguments to R types before passing to the method.

dfalbel:
This PR fixes https://github.com/rstudio/keras/issues/926#issuecomment-554489312
so the user doesn't need to manually convert arguments to R types kin order to use them in methods.

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -55,7 +55,10 @@ PyClass <- function(classname, defs = list(), inherit = NULL) {
   
   defs <- lapply(defs, function(x) {
     if (inherits(x, "function")) {
-      x <- inject_super(x)
+      f <- inject_super(x)
+      x <- function(...) {
+        do.call(f, lapply(list(...), py_to_r))
+      }
     }
     x
   })
--- a/tests/testthat/test-python-class.R
+++ b/tests/testthat/test-python-class.R
@@ -1,4 +1,4 @@
-context("classes")
+context("PyClass")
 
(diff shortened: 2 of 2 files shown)


jonathanbratt:
This works great in the vanilla test case, but appears to fail if the class inherits from `tensorflow::tf$keras$layers$Layer`.

For example, this works:

```
TestVanilla <- PyClass(
    "TestVanilla", 
    defs = list(
        `__init__` = function(self, a) {
            self$a <- a
            NULL
        },
        add_2 = function(self) {
            self$a + 2
        }
    )
)

tv <- TestVanilla(a = 2)
tv$add_2()
# [1] 4
```
But adding the `inherit` below makes it fail:

```
TestKerasLayer <- PyClass(
    "TestKerasLayer", 
    inherit = tensorflow::tf$keras$layers$Layer, # now, inherit keras layer
    defs = list(
        `__init__` = function(self, a) {
            self$a <- a
            NULL
        },
        add_2 = function(self) {
            self$a + 2
        }
    )
)

tkl <- TestKerasLayer(a = 2)
tkl$add_2()
# Error in py_call_impl(callable, dots$args, dots$keywords) : 
#     RuntimeError: Evaluation error: non-numeric argument to binary operator. 
```
(I know the above wouldn't work as a layer class for other reasons; just keeping the example compact.)

I'm afraid I don't yet understand the package well enough to know *why* it fails in the keras layer case.

PR #668, opened 2019-12-17
Add Matrix import to avoid the note in CRAN check result

terrytangyuan:
This is to resolve the following note in CRAN check result: https://cran.r-project.org/web/checks/check_results_reticulate.html

> Version: 1.14
Check: dependencies in R code
Result: NOTE
    Namespace in Imports field not imported from: ‘Matrix’
     All declared Imports should be used. 

Since we are using `methods` library to create new sparse matrices instead of using any methods explicitly from `Matrix`, I am just adding `@importFrom Matrix Matrix` as a workaround.

Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>


kevinushey:
Thanks!

PR #671, opened 2019-12-19
work in output dir when creating plots

### LINKED ISSUE #645: Plots created by Python chunk saved in wrong location during knit()
I created a sample [.Rmd file](https://github.com/fdetsch/knitr-playground/blob/master/R/plt_show.Rmd) as part of a [repository](https://github.com/fdetsch/knitr-playground) w/.Rproj. The .Rmd contains a Python code chunk that creates a plot.

As long as RStudio Preview Version 1.2.5019 > Tools > Global Options... > R Markdown > Evaluate chunks in directory: Document is set, the document compiles flawlessly (see output [.html]( http://htmlpreview.github.com/?https://github.com/fdetsch/knitr-playground/blob/master/R/plt_show.html)). However, when I change the evaluation location to 'Project' and then `knit()` the document, I get the following error: 

> File plt_show_files/figure-html/pyplot-1.png not found in resource path
> Error: pandoc document conversion failed with error 99
> Execution halted

In addition, a folder plt_show_files/figure-html with a file [pyplot-1.png](https://github.com/fdetsch/knitr-playground/blob/master/plt_show_files/figure-html/pyplot-1.png) in it is created at the root level of the project. I know that plots created by R chunks are rendered correctly despite the evaluation location being set to 'Project'; png files are in that case simply saved to a folder plt_show_files/figure-html relative to the location of the .Rmd. 

Is there any solution to get this working with Python code chunks without setting the evaluation location to 'Document'? Is this even related to **reticulate** or rather a RStudio issue?

Here's my 

```r
[master]> sessionInfo()
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18362)

Matrix products: default

locale:
[1] LC_COLLATE=German_Germany.1252  LC_CTYPE=German_Germany.1252    LC_MONETARY=German_Germany.1252
[4] LC_NUMERIC=C                    LC_TIME=German_Germany.1252    

attached base packages:

 [1] compiler_3.6.1   clisymbols_1.2.0 htmltools_0.4.0  tools_3.6.1      rstudioapi_0.10  yaml_2.2.0      
 [7] Rcpp_1.0.3       crayon_1.3.4     rmarkdown_1.16   knitr_1.26       xfun_0.11        digest_0.6.22   
[13] rlang_0.4.1      prompt_1.0.0     evaluate_0.14  
```

kevinushey:
Closes #645.

@cderv does this seem like the right fix to you?

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -210,25 +210,34 @@ eng_python_initialize <- function(options, context, envir) {
 
 eng_python_matplotlib_show <- function(plt, options) {
   
-  # work within the knitr 'base dir', to ensure that figures are
-  # generated in the same location where R figures might normally
-  # be generated
-  #
-  # https://github.com/rstudio/reticulate/issues/645
-  dir <- knitr::opts_knit$get("output.dir")
-  if (is.character(dir)) {
-    dir.create(dir, recursive = TRUE)
-    owd <- setwd(dir)
-    on.exit(setwd(owd), add = TRUE)
-  }
+  # we need to work in either base.dir or output.dir, depending
+  # on which of the two has been requested by the user. (note
+  # that output.dir should always be set)
+  dir <-
+    knitr::opts_knit$get("base.dir") %||%
+    knitr::opts_knit$get("output.dir")
+
+  # move to the requested directory
+  dir.create(dir, recursive = TRUE, showWarnings = FALSE)
+  owd <- setwd(dir)
+  on.exit(setwd(owd), add = TRUE)
   
+  # construct plot path
   plot_counter <- yoink("knitr", "plot_counter")
-  path <- knitr::fig_path(options$dev, number = plot_counter())
+  path <- knitr::fig_path(
+    suffix = options$dev,
+    options = options,
+    number = plot_counter()
+  )
+  
+  # save the current figure
   dir.create(dirname(path), recursive = TRUE, showWarnings = FALSE)
   plt$savefig(path, dpi = options$dpi)
   plt$clf()
-  knitr::include_graphics(path)
   
+  # include the requested path
+  knitr::include_graphics(path)
+
 }
 
 eng_python_initialize_matplotlib <- function(options, context, envir) {
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -209,13 +209,26 @@ eng_python_initialize <- function(options, context, envir) {
 }
 
 eng_python_matplotlib_show <- function(plt, options) {
+  
+  # work within the knitr 'base dir', to ensure that figures are
+  # generated in the same location where R figures might normally
+  # be generated
+  #
+  # https://github.com/rstudio/reticulate/issues/645
+  dir <- knitr::opts_knit$get("output.dir")
+  if (is.character(dir)) {
+    dir.create(dir, recursive = TRUE)
+    owd <- setwd(dir)
+    on.exit(setwd(owd), add = TRUE)
+  }
+  
   plot_counter <- yoink("knitr", "plot_counter")
   path <- knitr::fig_path(options$dev, number = plot_counter())
   dir.create(dirname(path), recursive = TRUE, showWarnings = FALSE)
   plt$savefig(path, dpi = options$dpi)
   plt$clf()
-  path <- normalizePath(path, winslash = "/", mustWork = FALSE)
   knitr::include_graphics(path)
+  
 }
 
 eng_python_initialize_matplotlib <- function(options, context, envir) {
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,11 @@
 
 ## reticulate 1.15 (UNDER DEVELOPMENT)
 
+- Fixed an issue where matplotlib plots would be included using absolute
+  paths, which fails in non-standalone documents rendered to HTML. (#669)
+
+- Fixed an issue where `reticulate` would attempt to flush a non-existent
+  stdout / stderr stream. (#584)
 
 ## reticulate 1.14
 


cderv:
About https://github.com/rstudio/reticulate/issues/645#issuecomment-567596524, this is by default that `knitr::opts_knit$get("base.dir")` is NULL. It happens with all engine, if not set. In fact, if NULL knitr, will put the figure directory in the _current knitr working dir_. If not null `knitr:::in_dir` will temporary set the working directory to `knitr::opts_knit$get("base.dir")`.

I would have done something like 
```r
  dir <- knitr::opts_knit$get("base.dir")
  if (is.character(dir)) {
    if (!file_test('-d', dir)) dir.create(dir, recursive = TRUE)
    owd <- setwd(dir)
    on.exit(setwd(owd), add = TRUE)
  }
```

However, it does not work like as expect. From a `debugonce` analysis, it seems the _current working directory_ of R engine and reticulate engine are different, I can't explain why yet... 
I know that with the current fix, if someone set the `base.dir` option following [knitr's options documentation](https://yihui.org/knitr/options/), it would have no effect for python plot. That is why I don't think this is the correct fix in the long run. 

I would like to look into it. I know knitr well now but not yet reticulate engine. Is it ok for you ? are you on a schedule on this one ? 

Also I wonder if some of knitr functions about directories like `in_dir` or `in_base_dir` would benefit to be exported to help knitr engine in other packages. 🤔 Will see afterwards I guess.

cderv:
@kevinushey this is clearer now. Here is my understanding

Knitr engines are executed in `knitr:::input_dir()` which is the current document directory or `root.dir` if set.  (https://github.com/yihui/knitr/blob/33d69c3acc0bb0a4f840e9ab836d0a4ec795c9fa/R/block.R#L107). The knitr `root.dir` is changed either by rstudio (what happens in #645) or by an argument in `rmarkdown::render` (what happens in rstudio/rmarkdown#1686) 

However, knitr figures are generated in `fig.path` relative to the working directory when knitting. This directory is saved in `output.dir` (https://github.com/yihui/knitr/blob/33d69c3acc0bb0a4f840e9ab836d0a4ec795c9fa/R/output.R#L144). Rmarkdown sets this directory to the input directory (https://github.com/rstudio/rmarkdown/blob/7f51e232c98b2f7db40b40cd593385fe76b3189b/R/render.R#L372)

That is why figures are saved in the input directory with R and the knitr root.dir with python. 

On this the fix you offer seems to be the correct one. I am not sure `knitr::opts_knit$get("output.dir")` needs to be tested for existence as it will always exists (directory of the Rmd file). 

However, there is also an advanced feature in knitr. It is the hability to save figures elsewhere (setting `base.dir`) and having the link of figures in markdown pointing to a specific location, (that is the `base.url` option). This is used for advanced stuff, like here in blogdown (https://bookdown.org/yihui/blogdown/hexo.html)

Currently, reticulate engine does not support that. If you want to support it, I would do 
```r
eng_python_matplotlib_show <- function(plt, options) {
  
  # knitr engines works in the knitr 'root dir' (see knitr:::input_dir) but
  # fig.path is relative to output.dir, set by rmarkdown to document directory.
  # python engine must ensure that figures are generated in the same location
  # where R figures might normally be generated
  # https://github.com/rstudio/reticulate/issues/645
  dir <- knitr::opts_knit$get("output.dir")
  owd <- setwd(dir)
  on.exit(setwd(owd), add = TRUE)
  # this allow to support advanced base.dir option from knitr
  in_base_dir <- yoink("knitr", "in_base_dir")
  in_base_dir({
    plot_counter <- yoink("knitr", "plot_counter")
    path <- knitr::fig_path(options$dev, options, number = plot_counter())
    if (!file_test("-d", dirname(path))) {
      dir.create(dirname(path), recursive = TRUE, showWarnings = FALSE)
    }
    plt$savefig(path, dpi = options$dpi)
    plt$clf()
    knitr::include_graphics(path)
  })
  
}
```

I tested it in https://github.com/cderv/reticulate/tree/feature/matplotlib-output-dir

Hope it is clear and it helps.

R/knitr-engine.R
```diff
@@ -209,13 +209,26 @@ eng_python_initialize <- function(options, context, envir) {
...
 eng_python_matplotlib_show <- function(plt, options) {
+  
+  # work within the knitr 'base dir', to ensure that figures are
+  # generated in the same location where R figures might normally
+  # be generated
+  #
+  # https://github.com/rstudio/reticulate/issues/645
+  dir <- knitr::opts_knit$get("output.dir")
+  if (is.character(dir)) {
+    dir.create(dir, recursive = TRUE)
```
cderv: this creates a warning I think because `output.dir` will always exist. 

I don't think it is necessary to test it. At least you could `&& !file_test("-d", dir)`

R/knitr-engine.R
```diff
@@ -209,13 +209,26 @@ eng_python_initialize <- function(options, context, envir) {
...
+    owd <- setwd(dir)
+    on.exit(setwd(owd), add = TRUE)
+  }
+  
   plot_counter <- yoink("knitr", "plot_counter")
   path <- knitr::fig_path(options$dev, number = plot_counter())
```
cderv: there a reason not to pass ˋoptions=options` here ?

kevinushey:
@cderv: thank you for your detailed analysis! If I understand correctly, the simplest solution here then is to use `base.dir` if it exists, and `output.dir` otherwise. (I wanted to avoid taking `in_base_dir` just because it's an internal function and I don't have explicit permission from Yihui to use it.)

I've made those changes and have updated the PR.

cderv:
> use base.dir if it exists, and output.dir otherwise

Yes this will work in most of the case. In case `base.dir` is a relative path and not an absolute one, R plot will be saved in `base.dir` relative to `output.dir` and python plot will be save in `base.dir` relative to working directory from where `render` is called. 

This is a special case I don't think will happen too often. 
* When a document is in a subdirectory from the project dir
* when `base.dir` is set to a relative path. 

`base.dir` is a an advanced option and I don't think it is recommended to be used without normalizing the path first though.

I just wanted to point you that there is still a difference in this case between R et Python chunks regarding figure path, but I think this should not happen and that your simplest solution is indeed ok.

Thanks!

PR #674, opened 2019-12-21
Implement r_to_py for external pointer

### LINKED ISSUE #660: Feature Request: convert external pointer to capsule with r_to_py
I hope to have the functionality to pass an R's external pointer object to python, possibly by converting it into a capsule object. So that

1. the pointer and the underlying data are shared between R and python.
2. R's external pointer object won't be freed while python is still pointing to it.

It would be similar to how memory are shared between numpy array and R's array.

If you think this idea is reasonable, I can try to implement it and submit a PR when I have time.

Marlin-Na:
Fixes #660

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -1159,6 +1159,21 @@ PyObject* r_to_py(RObject x, bool convert) {
   return obj;
 }
 
+// Python capsule wrapping an R's external pointer object
+static void free_r_extptr_capsule(PyObject* capsule) {
+  SEXP sexp = (SEXP)PyCapsule_GetContext(capsule);
+  ::R_ReleaseObject(sexp);
+}
+static PyObject* r_extptr_capsule(SEXP sexp) {
+  // underlying pointer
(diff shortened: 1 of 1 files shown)


Marlin-Na:
An example can be:
```
library(xptr)
library(reticulate)
xptr <- xptr::new_xptr("1122")
xptr
# <pointer: 0x462>
r_to_py(xptr)
# <capsule object NULL at 0x11938bb70>
```

kevinushey:
Thanks for the pull request!

We require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference me and this PR in your e-mail.)

Marlin-Na:
Hi @kevinushey , I have emailed the agreement.

Best,
Jialin

kevinushey:
Thank you!

PR #680, opened 2020-01-01
conda_binary didn't check conda binary location path properly

dracodoc:
- If a conda binary path is provided (xxx/bin), using `dirname` and `..` together will seek two levels up which is redundant.
- If a path didn't pass the previous test but still is a valid path, the function will report as a valid conda binary location.
   - If there do have this kind of cases, it should test the existence of conda binary, not the path itself.
   - If there is no valid case for this, it should report error. The second commit is based on this assumption, but I can change to first assumption if that's the case.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -243,11 +243,11 @@ conda_binary <- function(conda = "auto") {
   }
 
   # validate existence
-  if (!file.exists(conda))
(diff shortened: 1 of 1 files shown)


dracodoc:
Sorry, I realized the conda parameter is expecting the full path of conda binary, not the conda binary folder.

PR #682, opened 2020-01-02
don't force as this would apply to non-miniconda-envs as well

skeydan:
I am not totally sure what we want to do here... (long-term)

On the one hand, in the recent blog post we say 

> Currently, automatic Python environment configuration will only happen when using the aforementioned reticulate Miniconda installation.

On the other hand, we do have an escape hatch in place already (RETICULATE_AUTOCONFIGURE) so we could as well leave the default as is... and remove the comment 

https://github.com/rstudio/reticulate/blob/c403546661ee7f8273ff505a8cd4f4ec6e2bfa8e/R/python-packages.R#L4

instead (and update the blog post I guess)...

Sorry I didn't "really" check this out before (though I could have, as I'm in the have-to-opt-out situation for virtualenv all the time...)

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -56,7 +56,7 @@ ensure_python_initialized <- function(required_module = NULL) {
     py_inject_hooks()
     
     # install required packages
-    configure_environment()
+    if (should_configure()) configure_environment()
 
   }
 }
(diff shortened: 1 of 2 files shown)


kevinushey:
What about `force = interactive()` or something like that?

kevinushey:
FWIW the intention here is that if the user calls `configure_environment()` themselves, `force` should be unnecessary. However, if _we_ want to call `configure_environment()` ourselves (e.g. in response to Python being loaded) then it should only be done if Miniconda is being used.

Perhaps the Miniconda check needs to move outside of `configure_environment()` and into the place where we call `configure_environment()` ourselves. Does that make sense?

skeydan:
Oh, thanks for clarifying, I didn't know what was the intention behind FORCE...

I've now instead added a `should_configure` function with the (currently) 2 conditions, does this look useful to you?

R/package.R
```diff
@@ -176,4 +176,18 @@ initialize_python <- function(required_module = NULL, use_environment = NULL) {
 }
 
 
-
+should_configure <- function() {
+  
+  # allow users to opt out
+  configure_ok <- Sys.getenv("RETICULATE_AUTOCONFIGURE", unset = "TRUE")
+  configure_ok <- if (configure_ok %in% c("FALSE", "False", "0")) FALSE else TRUE
```
kevinushey: Could just use an early return false here

skeydan: right, guess it's programming not logic :-)
updated

kevinushey:
Thanks!

PR #684, opened 2020-01-06
Fix multiple inheritance

dfalbel:
This fixes a bug when creating a third level of inheritance with classes created from R. 
The problem is this line:

https://github.com/rstudio/reticulate/compare/master...dfalbel:fix-inheritance?expand=1#diff-9603c58bdf50909f0162edee456f1ef6L12

Turns out this is a common bug in python, see: https://stackoverflow.com/a/17509938/3297472

I also added a test that shows the bug if you run with the current reticulate version.

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -2,19 +2,35 @@ inject_super <- function(fun) {
   # for each function `fun` we need to place a `super` function in its
   # search path, and this `super` function must be able to access
   # the `self` argument passed to `fun`.
-  
+
   e <- new.env(parent = environment(fun))
   
   e$super <- function() {
     bt <- reticulate::import_builtins()
     # self is an argument passed to fun
-    self <- get("self", envir = parent.frame(), inherits = FALSE) 
-    bt$super(self$`__class__`, self)
+    self <- get("self", envir = parent.frame(), inherits = FALSE)
+    class_ <- get("__class__", envir = e, inherits = FALSE)
+    bt$super(class_, self)
   }
-  
+
   environment(fun) <- e # so fun can access `super`
-  
+
   fun
+}
+
+#' Enable the convertion scope so `self` fields can be accessed
+#' without the need to call `py_to_r`.
+#' 
+#' @param f a method/function of a Python class
+#' 
+enable_convert_scope <- function(f) {
+  function(...) {
+    args <- list(...)
+    # enable convertion scope for `self`
+    # the first argument is always `self`.and we don't want to convert it.
+    assign("convert", TRUE, envir = as.environment(args[[1]])) 
+    do.call(f, append(args[1], lapply(args[-1], py_to_r)))
+  }
 }
 
 #' Create a python class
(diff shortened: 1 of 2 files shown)


R/class.R
```diff
@@ -2,21 +2,37 @@ inject_super <- function(fun) {
...
   }
-  
+
   environment(fun) <- e # so fun can access `super`
-  
+
   fun
 }
 
+#' Enable the convertion scope so `self` fields can be accessed
```
kevinushey:
```suggestion
#' Enable the conversion scope so `self` fields can be accessed
```

R/class.R
```diff
@@ -56,20 +72,25 @@ PyClass <- function(classname, defs = list(), inherit = NULL) {
...
       f <- inject_super(x)
-      x <- function(...) {
-        args <- list(...)
-        # enable convertion scope for `self`
-        # the first argument is always `self`.and we don't want to convert it.
-        assign("convert", TRUE, envir = as.environment(args[[1]])) 
-        do.call(f, append(args[1], lapply(args[-1], py_to_r)))
-      }
+      x <- enable_convert_scope(f)
+      attr(x, "__env__") <- environment(f)
```
kevinushey: Do we really need to store a reference to the environment? It seems like we don't actually use it (beyond checking its existence).

R/class.R
```diff
@@ -56,20 +72,25 @@ PyClass <- function(classname, defs = list(), inherit = NULL) {
...
+  type <- bt$type(
     classname, inherit, 
     do.call(reticulate::dict, defs)
   )
+  
+  # we add a reference to the type here. so it can be acessed without needing
+  # to find the type from self.
+  lapply(defs, function(x) {
+    if(!is.null(e <- attr(x, "__env__"))) {
+      e$`__class__` <- type
```
kevinushey: Doesn't this cause some potential confusion between the real Python `__class__` attribute, and the thing we're putting in this environment?

R/class.R
```diff
@@ -56,20 +72,25 @@ PyClass <- function(classname, defs = list(), inherit = NULL) {
...
     x
   })
   
-  bt$type(
+  type <- bt$type(
     classname, inherit, 
     do.call(reticulate::dict, defs)
   )
+  
+  # we add a reference to the type here. so it can be acessed without needing
```
kevinushey:
```suggestion
  # we add a reference to the type here. so it can be accessed without needing
```

tests/testthat/test-python-class.R
```diff
@@ -214,7 +214,32 @@ test_that("self is not converted when there's a py_to_r method for it", {
   rm(py_to_r.python.builtin.Base, envir = .GlobalEnv)
 })
 
-
+test_that("can call super from an ingerited class", {
```
kevinushey:
```suggestion
test_that("can call super from an inherited class", {
```

R/class.R
```diff
@@ -2,21 +2,37 @@ inject_super <- function(fun) {
...
+
   fun
 }
 
+#' Enable the convertion scope so `self` fields can be accessed
+#' without the need to call `py_to_r`.
+#' 
+#' @param f a method/function of a Python class
+#' 
+enable_convert_scope <- function(f) {
```
kevinushey: The name of the function implies it's more generic than it really is (it's a very narrow helper used explicitly for PyClass). I'd prefer to just keep it a private helper since I doubt it would be reused in any other context.

R/class.R
dfalbel: we actually use it  to assign the `__class__` variable that is required to not depend on `self` to find the base class.  The python interpreter seems to do it automatically, but only when you create the function inside the class definition. 

See https://docs.python.org/3/library/functions.html#super

> The zero argument form only works inside a class definition, as the compiler fills in the necessary details to correctly retrieve the class being defined, as well as accessing the current instance for ordinary methods.

R/class.R
dfalbel: I used the name `__class__` following what python seems to use. For example if you run:

```
class Base:
  def __init__ ():
    print("Hello Base")
    

MyClass = type(
  "MyClass",
  (Base,),
  {
    "__init__": lambda x: super()
  }
)

a = MyClass()
```

You will get: `RuntimeError: super(): __class__ cell not found`.

R/class.R
dfalbel: There may be an easier way to do it, but I couldn't come up with. The idea is that `super` must be able to find what is the `type` without referring to `self` otherwise it will report the wrong type when there's an additional level of inheritance.

kevinushey: Ah, got it -- I think this is clear now. Thanks!

R/class.R
kevinushey: Got it. I was confused whether this was distinct from Python's own `__class__` object; but it's clear now that the intention is to ensure the Python `__class__` object is set with an appropriate object.

kevinushey:
Thanks!

PR #689, opened 2020-01-08
Fix incorrect extraction of parameter descriptions

terrytangyuan:
This should be extracting from `docs$parameters` instead of `docs$sections`.


kevinushey:
Thanks!

PR #694, opened 2020-01-11
forbid initialization of Python in .onLoad() / .onAttach()

kevinushey:
This ensures that packages do not accidentally behave in a user-hostile way; that is, by disallowing packages from initializing Python before explicitly requested by the user.


jennybc:
I played with this idea locally little and note that it does not detect the forbidden event if developer is iterating using `devtools::load_all()`. That person would discover their misstep the next time they attempt a full package build + install. Which is probably fine? But since I gathered the data, here's what the call stack looks like in this case. Note we have `run_pkg_hook()` instead of `runHook()`.

```
Browse[2]> calls
[[1]]
devtools::load_all(".")

[[2]]
pkgload::load_all(path = path, reset = reset, recompile = recompile, 
    export_all = export_all, helpers = helpers, quiet = quiet, 
    ...)

[[3]]
run_pkg_hook(package, "load")

[[4]]
nsenv[[f_name]](dirname(ns_path), package)

[[5]]
initialize_thing()

[[6]]
check_forbidden_initialization()
```

PR #700, opened 2020-01-17
add Ellipsis constant

skeydan:
Stuff like 

```
b = tf.zeros(shape = [2,2])
b[ ..., None]
```

is used pretty often from Python; from R though you'd need a workaround...

This PR adds `py_ellipsis()` so we can do

```
b <- tf$zeros(shape = c(2L,2L))
b[reticulate::py_ellipsis(), NULL]
```


kevinushey:
LGTM! Test failures just need a re-document run.

skeydan:
done, thanks!

PR #721, opened 2020-02-20
fix #706 handling of lineno with decorators

### LINKED ISSUE #706: knitr engine AST parsing needs tweak for decorator syntax, causes syntax error
If I use a python block in Rmarkdown with decorator syntax, I get an error. But pasting in IPython works (and looking closer in vim, I didn't see any weird newline characters, etc..).

Other syntax in Rmarkdown works as expected! Here's an example of erroring code.

``````
```{python}
from functools import singledispatch


@singledispatch
def repr_html(obj):
    return repr(obj)

```
``````

Error:

```
SyntaxError: multiple statements found while compiling a single statement (<string>, line 1)
```

## Screenshot

image

jsfalk:
This is my attempt to fix issue [#706](https://github.com/rstudio/reticulate/issues/706).

For functions with decorators, Python 3.8 updated `lineno` to point to the `lineno` of the `def`, rather than the `lineno` of the first decorator (see [this PR](https://github.com/python/cpython/pull/9731)).

I updated the code that extracts line numbers to use the first decorator's `lineno` if available, and otherwise the node's `lineno`. Unsure if this is the best approach, so I appreciate any feedback, but it appears to work.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -93,7 +93,11 @@ eng_python <- function(options) {
 
   # iterate over top-level nodes and extract line numbers
   lines <- vapply(parsed$body, function(node) {
-    node$lineno
+    if(py_has_attr(node, 'decorator_list') && length(node$decorator_list)) {
+      node$decorator_list[[1]]$lineno
+    } else {
+      node$lineno
+    }
   }, integer(1))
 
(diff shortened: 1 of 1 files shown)


kevinushey:
Looks good to me! Thank you for the contribution.

PR #722, opened 2020-02-20
suppress output with semicolon (fix #717)

### LINKED ISSUE #717: Avoid displaying all intermediate outputs in Rmd
The knitr engine displays all results, not just `stdout` or plots, for all lines. This is surprising because sometimes functions return values for convenience, and it's perfectly valid not to use that result if you don't need it. A real example is the sklearn `fit` function, which returns `self` just to permit easy initialization. Example (taken almost verbatim from some course materials I'm writing):

```{python}
from sklearn.feature_extraction.text import CountVectorizer
cv = CountVectorizer(analyzer="char")
docs = ["aaa", "ccc", "abc"]
cv.fit(docs)
cv.transform(docs).A
```

When knit, the chunk splits between `fit` and `transform`, interspersing a useless `CountVectorizer(...)` repr. When run as a single chunk, the two outputs appear in sequence.

I was expecting the behavior of Jupyter Notebook:

* Intermediate values are not displayed by default, but can be displayed by wrapping the result in `display` (or `print`).
* Final values are displayed by default but can be suppressed by adding a `;` at the end of the line.

Perhaps you don't want that exact behavior because R works somewhat differently, but at least it would be nice to have some way of suppressing intermediate outputs (e.g., the trailing `;`, which is the first thing I'd tried).

It looks like this might entail some changes here: https://github.com/rstudio/reticulate/blob/master/R/knitr-engine.R#L142

jsfalk:
I've implemented the functionality described in [#717](https://github.com/rstudio/reticulate/issues/717). Adding a trailing semicolon stops return values from being output.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -142,11 +142,15 @@ eng_python <- function(options) {
     # save last value
     last_value <- py_last_value()
 
+    # use trailing semicolon to suppress output of return value
+    suppress <- grepl(";\\s*$", snippet)
+    compile_mode <- if(suppress) "exec" else "single"
+
     # run code and capture output
     captured <- if (capture_errors)
-      tryCatch(py_compile_eval(snippet), error = identity)
+      tryCatch(py_compile_eval(snippet, compile_mode), error = identity)
     else
-      py_compile_eval(snippet)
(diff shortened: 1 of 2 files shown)


kevinushey:
LGTM -- thanks!

PR #735, opened 2020-03-05
Add optional pip_options argument

### LINKED ISSUE #734: Feature Request: pass pip options through `py_install`
I would really like to see `pip` options can be passed through `py_install`. More specifically, I am interested in the `--install-option` option that will be passed to `setup.py`.

```
  --install-option <options>  Extra arguments to be supplied to the setup.py install command (use
                              like --install-option="--install-scripts=/usr/local/bin"). Use
                              multiple --install-option options to pass multiple options to
                              setup.py install. If you are using an option with a directory path,
                              be sure to use absolute path.
```

It seems that it would be easy to implement an option `pip_options` for `py_install` which accepts a character vector that will be passed to `pip` (if pip is used for installation).

Marlin-Na:
Fixes #734

### DIFF
--- a/R/pip.R
+++ b/R/pip.R
@@ -14,12 +14,13 @@ pip_version <- function(python) {
 }
 
 
-pip_install <- function(python, packages, ignore_installed = FALSE) {
+pip_install <- function(python, packages, pip_options = character(), ignore_installed = FALSE) {
 
   # construct command line arguments
   args <- c("-m", "pip", "install", "--upgrade")
   if (ignore_installed)
     args <- c(args, "--ignore-installed")
+  args <- c(args, pip_options)
   args <- c(args, packages)
 
   # run it
--- a/R/conda.R
+++ b/R/conda.R
@@ -164,6 +164,9 @@ conda_remove <- function(envname, packages = NULL, conda = "auto") {
 #'   you don't want a pip install to attempt an overwrite of a conda binary
 #'   package (e.g. SciPy on Windows which is very difficult to install via pip
 #'   due to compilation requirements).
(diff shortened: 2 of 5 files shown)


R/pip.R
```diff
@@ -14,12 +14,13 @@ pip_version <- function(python) {
...
 
 
-pip_install <- function(python, packages, ignore_installed = FALSE) {
+pip_install <- function(python, packages, pip_options = character(), ignore_installed = FALSE) {
 
   # construct command line arguments
   args <- c("-m", "pip", "install", "--upgrade")
   if (ignore_installed)
     args <- c(args, "--ignore-installed")
+  args <- c(args, pip_options)
```
kevinushey: Some questions:

1. Should this be `pip_arguments` rather than `pip_options`? Since this generically allows any kind of argument to `pip`, not just `--install-option` as described in the issue.

2. Do we need to `shQuote()` these arguments? Or do you think the onus is on the caller to shell-quote arguments as necessary?

Marlin-Na: 1. I don't have opinion about the name, but I remember `devtools` has used name like `build_opts` for extra command line arguments passed to `R CMD build`. So `pip_options` possibly should be fine.

2. That's a good point. But I think if we `shQuote` here, it will force user to use `c("--option1", "--option2")` instead of `"--option1 --option2"`. Also I checked what `devtools` do with the `build_opts`, they do not `shQuote` them. Possibly we can follow that.

kevinushey: Both good points. Okay, I'm happy with this as-is. Thanks!

Issue #727, opened 2020-02-27
Error in sparse matrix conversion: "Error in validObject(.Object): invalid class "dgRMatrix" object: slot j is not increasing inside a column"


paulofelipe:
Hi, 

I'm getting an error when a sparse matrix generated in python is converted to R. Please see the code below. My session info is at the end. Thank you in advance!

``` r
library(reticulate)

sklearn <- import("sklearn")

test <- c("test reticulate", "load the package")

vec <- sklearn$feature_extraction$text$CountVectorizer()

vec$fit_transform(c("aaa", "bbb"))
#> 2 x 2 sparse Matrix of class "dgRMatrix"
#>         
#> [1,] 1 .
#> [2,] . 1

vec$fit_transform(test)
#> Error in validObject(.Object): invalid class "dgRMatrix" object: slot j is not increasing inside a column
```

<sup>Created on 2020-02-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 19.10

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.8.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.8.0

locale:
 [1] LC_CTYPE=pt_BR.UTF-8       LC_NUMERIC=C               LC_TIME=pt_BR.UTF-8       
 [4] LC_COLLATE=pt_BR.UTF-8     LC_MONETARY=pt_BR.UTF-8    LC_MESSAGES=pt_BR.UTF-8   
 [7] LC_PAPER=pt_BR.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=pt_BR.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] reticulate_1.14-9001

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.3      lattice_0.20-38 ps_1.3.2        withr_2.1.2     packrat_0.5.0  
 [6] digest_0.6.25   rappdirs_0.3.1  R6_2.4.1        grid_3.6.1      jsonlite_1.6.1 
[...]

PR #738, opened 2020-03-18
sort indices of csr sparse matrix before convert to R dgRMatrix

paulofelipe:
This pull request tries to solve the problem in the issue  #727.

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -470,6 +470,7 @@ r_to_py.dgRMatrix <- function(x, convert = FALSE) {
 #' @export
 py_to_r.scipy.sparse.csr.csr_matrix <- function(x) {
   disable_conversion_scope(x)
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks for the PR! Would you be willing to add a unit test, as well?

paulofelipe:
Hi @kevinushey!

By default, when converting a sparse matrix (dgRMatrix) from R to Python, the matrix indexes are automatically ordered in an ascending order, which makes the conversion back to R occur without any problems.

The creation of a csr sparse matrix with indexes out of order happens only when I use the CountVectorizer() function. Any tips on how to do this manually so that I can write the test?

Thanks in advance!

kevinushey:
I was hoping there was a way to create such sparse matrices "by hand" in Python or similar.

PR #752, opened 2020-04-10
Allow specifying channel in conda_create

### LINKED ISSUE #751: Allow for forge and channels arguments in conda_create
Is there a reason that `conda_create` (unlike `conda_install`) does not have the `forge` and `channels` arguments? These arguments are useful to create conda environments that include packages that live in other channels, e.g. conda-forge.

Adding these arguments should be easy. I'd be happy to make a PR.

jtilly:
This PR adds the arguments `forge` and `channel` to `conda_create` analogously to how they work in `conda_install`.

This is a tiny PR, but adding these arguments will change behavior for two reasons:
- I set the default value for `conda` to `TRUE` just like for `conda_install`. Previously, it would just use whatever the user had configured conda to do (see `conda config --show-sources`)
- I'm squeezing the `forge` and `channel` arguments in before the `conda` argument just like for `conda_install`.

I also added a tiny test for this. But to make the test meaningful I need to access the channel of installed packages somehow (ideally without activating the conda environment). So I amended `conda_list_packages` to also return a column with the `channel` just like you see when you run `conda list` in the command line.

Closes #751

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -486,6 +486,7 @@ conda_list_packages <- function(envname = NULL, conda = "auto", no_pip = TRUE) {
     package     = parsed$name,
     version     = parsed$version,
     requirement = paste(parsed$name, parsed$version, sep = "="),
+    channel     = parsed$channel,
     stringsAsFactors = FALSE
   )
   
--- a/R/conda.R
+++ b/R/conda.R
@@ -103,8 +103,18 @@ conda_list <- function(conda = "auto") {
 }
 
 
-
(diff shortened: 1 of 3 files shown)


jtilly:
Travis checks are green, see https://travis-ci.org/github/rstudio/reticulate/builds/673574540 though somehow GitHub didn't get the message. I think that the environment tests I added don't even get run on travis.

kevinushey:
LGTM -- thanks!

PR #761, opened 2020-05-04
Add python_dependencies vignette to header menu

nealrichardson:
I noticed that this existed but was not in the menu


pkgdown/_pkgdown.yml
```diff
@@ -26,6 +26,8 @@ navbar:
           href: articles/arrays.html
         - text: "Using reticulate in an R Package"
           href: articles/package.html
+        - text: "Managing an R Package’s Python Dependencies"
```
kevinushey, line 29:
```suggestion
        - text: "Managing an R Package's Python Dependencies"
```

Issue #762, opened 2020-05-07
RangeIndex._start deprecated


tseppelt:
When passing a pandas data frame from Python to R I get the following warnings:
```
sys:1: FutureWarning: RangeIndex._start is deprecated and will be removed in a future version. Use RangeIndex.start instead
sys:1: FutureWarning: RangeIndex._stop is deprecated and will be removed in a future version. Use RangeIndex.stop instead
sys:1: FutureWarning: RangeIndex._step is deprecated and will be removed in a future version. Use RangeIndex.step instead
```
I'm wondering whether this is because of ll. 316-318 in [R/conversion.R](https://github.com/rstudio/reticulate/blob/master/R/conversion.R).

galen211:
I'm also getting this error in my R `reticulate` environment while using `pd.read_feather()` in python.

galen211:
Seems like the answer is in this issue: [https://github.com/rstudio/reticulate/issues/2](https://github.com/rstudio/reticulate/issues/2)

PR #763, opened 2020-05-07
Deprecated RangeIndex #762

tseppelt:


kevinushey:
I think we need the PR to be backwards-compatible with older versions of `pandas` as well. Do older versions of Pandas also provide these members without the `_` prefix?

I wonder if we can introspect on the installed version of `pandas` in some way?

tseppelt:
As far as I understand the ``pandas`` code, the new attributes were introduced in [this pull request](https://github.com/pandas-dev/pandas/pull/26581) in June 2019. Before June 2019 these members did not exist, I think.

charlstown:
I have the same issue:
sys:1: FutureWarning: RangeIndex._* is deprecated and will be removed in a future version. Use RangeIndex.start instead
*start, stop step

The funny thing is, I didn't get the error at first but now every time I try to run py$dataframe I'm getting a fatal error.

kevinushey:
```
>>> pandas.__version__
'0.24.2'
>>> index.start
AttributeError: 'RangeIndex' object has no attribute 'start'
```

So we definitely cannot just switch from `_start` to `start` -- we need to check which of the two exists and use that.

kevinushey:
Fixed on master now.

Issue #739, opened 2020-03-18
Perform automated sys.stdout.flush() during loops


fdetsch:
Console output created during a loop is currently printed to the console only after execution finished. As I recently found out, a manual workaround for this would be to manually add `sys.stdout.flush()` to the loop body (code copied from [StackOverflow](https://stackoverflow.com/questions/60736401/reticulate-doesnt-print-to-console-in-real-time)), which then creates console output in real time.

```r
library(reticulate)

py_run_string("
import time
from sys import stdout

for i in range(5):
   print(str(i))
   time.sleep(1.5) # sleep for 1.5 sec
   stdout.flush()
")
```

Would it be feasible to integrate this approach natively in **reticulate**?

kevinushey:
Alternatively, you can call `print(str(i), flush = True)` to ensure output is flushed with the print call.

fdetsch:
Great workaround! I've added this to the linked SO post for future reference.

randy3k:
It seems that the `flush` workaround doesn't work in `repl_python()` in RStudio.

```py
import time
for i in range(10):
  print(i, flush = True)
  time.sleep(1)

```

Nothing will be printed in `repl_python()` at first until the whole block is done.

PR #774, opened 2020-06-02
do not capture output in repl mode

randy3k:
The output of the following python code in `repl_python()` will be buffered until the whole block is done.

```py
import time
for i in range(10):
  print(i, flush = True)
  time.sleep(1)
```

It is because `py_compile_eval` has captured the output. This PR adds an option to `py_compile_eval` so that the output is not captured.

Before:
before

After:
after

related to https://github.com/rstudio/reticulate/issues/739#issuecomment-637374290

### DIFF
--- a/R/repl.R
+++ b/R/repl.R
@@ -229,9 +229,7 @@ repl_python <- function(
 
       # submit previous code
       pasted <- paste(previous, collapse = "\n")
-      output <- tryCatch(py_compile_eval(pasted), error = handle_error)
-      if (is.character(output) && nzchar(output))
-        cat(output, sep = "")
+      tryCatch(py_compile_eval(pasted, capture = FALSE), error = handle_error)
 
       # now, handle the newest line of code submitted
       buffer$set(contents)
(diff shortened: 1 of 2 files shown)


randy3k:
I don't think the travis failures are relevent.

kevinushey:
LGTM -- thanks for the PR!

randy3k:
I also notice that python chunk in Rmarkdown file in RStudio has a similar issue when executed in line. Do you think the same patch is would be applied to knitr engine? Or it will break something?

randy3k:
Oh, it doesn't seem to work if I simply set `capture` to FALSE in `eng_python.`

https://github.com/rstudio/reticulate/blob/0071c9db6e8817b7bb3000145238ad73797da3c9/R/knitr-engine.R#L151-L153

Two issues

1. knitting the Rmarkdown file doesn't work, the python chunk 's output is not captured.
2. when evaluating the python chunk inline, the output goes to the console instead of inline.

PR #790, opened 2020-07-02
Prefer miniconda

### LINKED ISSUE #786: Miniconda preference
In an environment in which `conda` is installed and `conda` is in the path,  If I do:

```
install_miniconda()
conda_create("r-reticulate", packages = "python")
```

This will create the environment in the system conda installation. 
Could we prefer our `miniconda` installation in this case? 

I mean checking for `miniconda_exists` here before the conda the system installation?

https://github.com/rstudio/reticulate/blob/bf98324e575d0bd80f6d055cc60dd52240c005ea/R/conda.R#L347-L358

dfalbel:
This PR makes miniconda prefered (if installed) over system conda installation. Fix #786

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -31,8 +31,9 @@
 #' following locations:
 #' 
 #' 1. The location specified by the `reticulate.conda_binary` \R option;
-#' 2. The program `PATH`;
-#' 3. A set of pre-defined locations where Conda is typically installed.
+#' 2. The `miniconda_path()` location if it exists.
+#' 3. The program `PATH`;
+#' 4. A set of pre-defined locations where Conda is typically installed.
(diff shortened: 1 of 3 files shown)


R/conda.R
```diff
@@ -31,8 +31,9 @@
 #' following locations:
 #' 
 #' 1. The location specified by the `reticulate.conda_binary` \R option;
-#' 2. The program `PATH`;
-#' 3. A set of pre-defined locations where Conda is typically installed.
+#' 2. The `miniconda_path()` location if it exists.
```
kevinushey, line 34:
```suggestion
#' 2. The [miniconda_path()] location (if it exists);
```

PR #792, opened 2020-07-03
Check for a conda installation before prompting for miniconda

### LINKED ISSUE #789: Autoconfiguration when user already have conda installed
On Windows, when you freshly install R and the Anaconda distribution, and you run any 

```
reticulate::py_install()
```

You'll get prompted to install miniconda. If you opt-out (because you already have conda) you will get the error:

```
Error: Installing Python packages into a virtualenv is not supported on Windows
```

It happens because the discovered python is: `C:/Users/dfalbel/anaconda3/python.exe`
And the `python_info` the function we use to find the env type is returning:

```
reticulate:::python_info(reticulate::py_config()$python)
$python
[1] "C:/Users/dfalbel/anaconda3/Scripts/python.exe"

$type
[1] "virtualenv"

$root
[1] "C:/Users/dfalbel/anaconda3"
```

I could do a PR that checks for `conda_binary` around here:

https://github.com/rstudio/reticulate/blob/master/R/config.R#L172-L184

and in this case we also create the r-reticulate environment using the conda installation found:

https://github.com/rstudio/reticulate/blob/master/R/config.R#L188-L195

What do you think?

dfalbel:
This PR adds 3 fixes:

1. On Windows it won't identify the root conda path as a virtualenv by checking if  `!file.exists("condabin/conda")`.

2. It ignores the root conda environment as a possible environment in `conda_list` by checking if it's path is equal to the `root_prefix` returned by `conda info`. This environment is not correctly writable.

3. If users opts out for miniconda, they can still have a fresh conda installation that don't have any environment. (conda is installed, but no environment is available). In this case we create an `r-reticulate` environment.

It fixes #789

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -515,7 +515,7 @@ conda_list_packages <- function(envname = NULL, conda = "auto", no_pip = TRUE) {
 }
 
 conda_installed <- function() {
-  condabin <- tryCatch(conda_binary(conda = conda), error = identity)
+  condabin <- tryCatch(conda_binary(), error = identity)
   if (inherits(condabin, "error"))
     return(FALSE)
   else
--- a/R/conda.R
+++ b/R/conda.R
@@ -66,11 +66,22 @@ conda_list <- function(conda = "auto") {
   # strip out anaconda cloud prefix (not valid json)
   if (length(conda_envs) > 0 && grepl("Anaconda Cloud", conda_envs[[1]], fixed = TRUE))
     conda_envs <- conda_envs[-1]
+  
+  # parse conda info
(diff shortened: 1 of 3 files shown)


R/conda.R
```diff
@@ -506,3 +513,11 @@ conda_list_packages <- function(envname = NULL, conda = "auto", no_pip = TRUE) {
   )
   
 }
+
+conda_installed <- function() {
```
kevinushey, line 517: Should this accept a `conda` argument? In case the caller of this function wants to pass a specific `conda` path.

PR #797, opened 2020-07-15
updates for virtualenv_create

### LINKED ISSUE #796: Virtualenvs are hard-coded to be created with the --system-site-packages flag
I have an issue with a shinyapps.io application that is no longer able to pip install packages using reticulate. The error I am seeing in the logs is ```ImportError: No module named 'setuptools.build_meta'``` 

I believe it is a result of this open issue with pip: https://github.com/pypa/pip/issues/6264 (Some additional helpful information is provided here by the setuptools maintainers: https://github.com/pypa/setuptools/issues/1694#issuecomment-466010982) 

The application is running on an older version of Ubuntu (Xenial) which ships with older versions of system python packages such as setuptools: https://packages.ubuntu.com/xenial/python-setuptools

Because the reticulate virtualenv is always created with the --system-site-packages flag, the user cannot opt out of including system packages even when running in virtualized environments such as shinyapps.io where they do not have control over the system. 

https://github.com/rstudio/reticulate/blob/a8701771fe759161c76a8e4a81d4ccc506269de1/R/virtualenv.R#L79

It seems as though this flag should not be set since it breaks the isolation from the system packages or it should at least be an optional argument.

kevinushey:
Closes #796.

### DIFF
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -41,7 +41,7 @@
 #'
 #' @param system_site_packages Boolean; create new virtual environments with
 #'   the `--system-site-packages` flag, thereby allowing those virtual
-#'   environments to access the system's site packages.
+#'   environments to access the system's site packages. Defaults to `FALSE`.
 #'
 #' @param ... Optional arguments; currently ignored for future expansion.
 #'
--- a/R/virtualenv.R
+++ b/R/virtualenv.R
@@ -41,7 +41,7 @@
 #'
 #' @param system_site_packages Boolean; create new virtual environments with
 #'   the `--system-site-packages` flag, thereby allowing those virtual
-#'   environments to access the system's site packages?
+#'   environments to access the system's site packages.
 #'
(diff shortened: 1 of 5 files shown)


R/virtualenv.R
```diff
@@ -33,6 +33,16 @@
...
 #'
+#' @param packages A set of Python packages to install (via `pip install`) into
+#'   the virtual environment, after it has been created. By default, the
+#'   `"numpy"` package will be installed, and the `pip`, `setuptools` and
+#'   `wheel` packages will be updated. Set this to `FALSE` to avoid installing
+#'   any packages after the virtual environment has been created.
+#'
+#' @param system_site_packages Boolean; create new virtual environments with
+#'   the `--system-site-packages` flag, thereby allowing those virtual
+#'   environments to access the system's site packages?
```
skeydan, line 44:
```suggestion
#'   environments to access the system's site packages.
```

skeydan: should we indicate what is the default?

PR #854, opened 2020-09-30
Improve it to make it more readable  NAMESPACE to NAME_SPACE

Chirag0410:


kevinushey:
R packages require this file to be called `NAMESPACE`.

PR #864, opened 2020-10-07
[WIP] Remove wrapper.R as similar functionalities have been moved to {scaffolder}

terrytangyuan:
These functions are now being further developed in https://github.com/terrytangyuan/scaffolder instead.

Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>


kevinushey:
Since these functions are exported (and hence part of the API), I'd rather keep them but instead delegate to the associated functions in the scaffolder package. Does that sound reasonable?

terrytangyuan:
Sounds good. I'll update this PR soon.

terrytangyuan:
@kevinushey Actually scaffolder depends on reticulate and delegating would introduce circular dependencies so it does not seem feasible.

kevinushey:
Fair enough. In that case I think it's fine to just remove them since they're unlikely to be commonly used functions; if anyone is affected I'll point them towards the scaffolder package. Sorry for the trouble!

terrytangyuan:
No problem. Sounds good and feel free to ping me if anyone has encountered any issues as well.

terrytangyuan:
@kevinushey The CI for scaffolder is failing though and I just realized that it needs to use `reticulate:::help_handler`. Would it be possible for `reticulate` to export `help_handler` and mark it as internal with `#' @keywords internal`?

kevinushey:
That sounds fine to me.

kevinushey:
I've exported the function as `py_help_handler` in the development version -- let me know if that's fine.

You can also access the internals with something like:

```
reticulate <- asNamespace("reticulate")
reticulate$help_handler(...)
```

if you need something that works in older versions of `reticulate`.

terrytangyuan:
That works perfectly. Thanks!

PR #899, opened 2020-12-01
Fix formals check in `py_iterator()`

lionel-:
Fixes a typo in the validation of formals.

```r
py_iterator(function(x) x)
#> Error in formals(fn) != 0 :
#>   comparison (2) is possible only for atomic and list types
```

Becomes

```r
#> Error in reticulate::py_iterator(function(x) x) :
#>   fn must be an R function with no arguments
```


kevinushey:
Thanks!

PR #981, opened 2021-03-29
insert path to rpytools for embedded R

randy3k:
it is needed for R embedded in python because the `PYTHONPATH` trick no longer works.

related #208


kevinushey (approved):
LGTM -- thank you!

PR #1018, opened 2021-06-11
restore LD_LIBRARY_PATH on linux

t-kalinowski:
added on.exit(add=TRUE) so that this call can coexist with the
other on.exit calls in the same frame.


kevinushey:
Thanks for the PR!

Issue #1019, opened 2021-06-12
Wrong Architecture on M1 chip (R 4.1.0 ARM)


dipterix:
I installed R 4.1.0 (ARM) that supports Apple M1 chip natively. However, reticulate cannot launch python

```
 Error in py_initialize(config$python, config$libpython, config$pythonhome,  : 
  /Users/dipterix/Library/r-ravepy/ravepy-conda-env/lib/libpython3.9.dylib - dlopen(/Users/dipterix/Library/r-ravepy/ravepy-conda-env/lib/libpython3.9.dylib, 10): no suitable image found.  Did find:
	/Users/dipterix/Library/r-ravepy/ravepy-conda-env/lib/libpython3.9.dylib: mach-o, but wrong architecture
	/Users/dipterix/Library/r-ravepy/ravepy-conda-env/lib/libpython3.9.dylib: mach-o, but wrong architecture 
```

After some debugging, I found the conda installed by `reticulate` was `x86_64` arch while R is arm64:

R session information:

```r
> sessionInfo()
R version 4.1.0 (2021-05-18)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Big Sur 11.4
```

Reticulate download message:
```
* Downloading 'https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh' ...
trying URL 'https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh'
Content type 'application/x-sh' length 57112343 bytes (54.5 MB)
==================================================
downloaded 54.5 MB
```

In case you need system information:

```
> Sys.info()
                                                                                                sysname 
                                                                                               "Darwin" 
                                                                                                release 
                                                                                               "20.5.0" 
                                                                                                version 
                                                                                                "XXXXX" 
                                                                                               nodename 
                                                                                                "XXXXXX" 
                                                                                                machine 
                                                                                                "arm64" 
                                                                                                  login 
                                                                                                 "root" 
                                                                                                   user 
                                                                                             "dipterix" 
                                                                                         effective_user 
                                                                                             "dipterix" 
```

This is similar to a previous issue #1012 , but that one was on linux. This is on mac

dipterix:
Just noticed that miniconda is unavailable to osx arm. This might take a while to fix. 

For those who tracked this issue, you might need to install your own version of miniconda via https://github.com/conda-forge/miniforge

[...]

Issue #1012, opened 2021-05-28
Wrong miniconda architecture use by `reticulate::install_miniconda`


dipterix:
Linux ARM installation goes wrong by default. Possibly installing x64 version. I ended up installing miniconda by myself

OS info: (Raspberry PI 4 B)
```
R version 4.1.0 (2021-05-18)
Platform: aarch64-conda-linux-gnu (64-bit)
Running under: Debian GNU/Linux 10 (buster)
```

Lines with issues:
```r
> reticulate:::miniconda_installer_url()
[1] "https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh"
```
which should be https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-armv7l.sh instead.

Here is my R version info in case you don't have arm linux in hand. Check `arch` item

```r
> R.Version()
$platform
[1] "aarch64-conda-linux-gnu"

$arch
[1] "aarch64"

$os
[1] "linux-gnu"

$system
[1] "aarch64, linux-gnu"

$status
[1] ""

$major
[1] "4"

$minor
[1] "1.0"

$year
[1] "2021"

$month
[1] "05"

$day
[1] "18"

$`svn rev`
[1] "80317"

$language
[1] "R"

$version.string
[1] "R version 4.1.0 (2021-05-18)"

$nickname
[1] "Camp Pontanezen"


```

dipterix:
Also after installing conda, reticulate complaint `No non-system installation of Python could be found.` but python3 is available with the system

```r
> reticulate::py_config()
No non-system installation of Python could be found.
Would you like to download and install Miniconda?
Miniconda is an open source environment management system for Python.
See https://docs.conda.io/en/latest/miniconda.html for more details.

Would you like to install Miniconda? [Y/n]: n
Installation aborted.
python:         /home/pi/miniconda3/bin/python3
libpython:      /home/pi/miniconda3/lib/libpython3.9.so
pythonhome:     /home/pi/miniconda3:/home/pi/miniconda3
version:        3.9.4 | packaged by conda-forge | (default, May 10 2021, 22:03:40)  [GCC 9.3.0]
numpy:          /home/pi/miniconda3/lib/python3.9/site-packages/numpy
numpy_version:  1.20.3

```

kevinushey:
This is where we construct the Miniconda URL -- unfortunately we don't handle arm64 yet.

[...]

PR #1043, opened 2021-08-09
Allow conda to be installed from customized place

dipterix:
Added options `reticulate.miniconda.url` to allow users to specify where to download miniconda, to install conda from intranet, conda-forge or personal repos. 

This quick fix solves #1012 #1019.


kevinushey:
Thanks!

Issue #1052, opened 2021-09-15
reticulate::conda_install(pip=TRUE) broken for 1.21


andrew-christianson:
With the recently-released reticulate 1.21, `conda_install` fails if pip=TRUE e.g.

```r
install.packages('reticulate')
reticulate::conda_create("r-test-broken", packages=c('python=3.8'))
reticulate::conda_install("tqdm", pip=TRUE, envname="r-test-broken")
```
returns the error: 

```r
Error in conda_dll_paths() : conda environment r-reticulate not found
```

with traceback

```
4: stop("conda environment ", envname, " not found")
3: conda_dll_paths()
2: pip_install(python = python, packages = packages, pip_options = pip_options,
       ignore_installed = pip_ignore_installed)
1: reticulate::conda_install("tqdm", pip = TRUE, envname = "r-test-broken")
```

AFAICT, the issue is that  1.21 adds a call to [conda_dll_paths](https://github.com/rstudio/reticulate/blob/e48fb8cf09989fe7c3e7e708ece4058fdfd7cbfa/R/conda.R#L403) in [pip_install](https://github.com/rstudio/reticulate/blob/e48fb8cf09989fe7c3e7e708ece4058fdfd7cbfa/R/pip.R#L33), without passing the environment name, so it ultimately falls back to the [default](https://github.com/rstudio/reticulate/blob/4967abddb35865a1be6ad839298695f1481bcade/R/python-environments.R#L5) case of python_environment_resolve, which reports ‘r-reticulate’ unless the user has set RETICULATE_PYTHON_ENV.  That environment is not guaranteed to exist, so the following subset returns a 0-row data frame and falls back to the 'environment not found' error condition.

I'm not close enough to reticulate's development to know the right fix, but seems like, at least, pip_install should take an envname to pass to conda_dll_paths, and conda_install should provide the env set by the user? 

For completeness, session and system info, but I've replicated this on a few systems, including Databricks and the one below (a local dev machine)
sessionInfo:
```
R version 4.1.1 (2021-08-10)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 11 (bullseye)

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.13.so

locale:
[...]

PR #1053, opened 2021-09-17
inject PATHs when using pip in condaenv

t-kalinowski:
when attempting to run `tensorflow::install_tensorflow()` in RStudio on Windows, this error message can pop up:
```
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/tensorflow/
WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/tensorflow/
WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/tensorflow/
WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/tensorflow/
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/tensorflow/
Could not fetch URL https://pypi.org/simple/tensorflow/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/tensorflow/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
ERROR: Could not find a version that satisfies the requirement tensorflow==2.6.* (from versions: none)
ERROR: No matching distribution found for tensorflow==2.6.*
WARNING: pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.
Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
 Error: Error installing package(s): "\"tensorflow==2.6.*\"" 
```

This fixes the issue by prefixing the PATH variable so the correct
"openssl.exe" executable is found. e.g.: 
`"C:\Users\kalin\AppData\Local\r-miniconda\envs\r-reticulate\Library\bin\openssl.exe" `
instead of 
`"C:\rtools40\usr\bin\openssl.exe"`

related: 
https://github.com/conda/conda/issues/6064
https://github.com/conda/conda/issues/8273

https://github.com/rstudio/reticulate/commit/5339495343d277fa30380502b5ac77b2cdd41d8c
https://github.com/rstudio/reticulate/issues/1052

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -392,7 +392,7 @@ numeric_conda_version <- function(conda = "auto", version_string = conda_version
   # substitute rc|beta|alpha|whatever suffix with .
   v <- sub("[A-Za-z]+", ".", v) 
   
-  if(endsWith(v, "."))
+  if (grepl("\\.$", v))
     v <- paste0(v, "9000")
   
   numeric_version(v)
--- a/R/conda.R
+++ b/R/conda.R
@@ -589,7 +589,7 @@ conda_run <- function(cmd, args = c(), conda = "auto", envname = NULL,
   conda <- conda_binary(conda)
   envname <- condaenv_resolve(envname)
 
-  if(numeric_conda_version(conda) < "4.9")
+  if (numeric_conda_version(conda) < "4.9")
     stopf(
 "`conda_run()` requires conda version >= 4.9.
 Run `miniconda_update('%s')` to update conda.", conda)
--- a/R/conda.R
+++ b/R/conda.R
@@ -376,9 +376,25 @@ conda_version <- function(conda = "auto") {
 }
 
 
-numeric_conda_version <- function(conda = "auto") {
-  v <- conda_version(conda)
-  v <- sub("(conda) ([0-9.]+)", "\\2", v)
+numeric_conda_version <- function(conda = "auto", version_string = conda_version(conda)) {
+  # some plausible version strings: 
+  # "conda 4.6.0"                 
+  # "conda 4.6.0b0"               
+  # "conda 4.6.0rc1"              
+  # "conda 4.6.0rc1.post3+64bde06"
+  v <- version_string 
+  v <- sub("^conda ", "", v) # drop hardcoded prefix
+  
+  # https://github.com/conda/conda/blob/c1579681d1468af3d1b4af3083bed33f8391e861/conda/_vendor/auxlib/packaging.py#L142
+  # if dev version string: "{0}.post{1}+{2}".format(version, post_commit, hash)
+  v <- sub("\\.post(\\d)\\+.+$", ".\\1", v)
+
+  # substitute rc|beta|alpha|whatever suffix with .
+  v <- sub("[A-Za-z]+", ".", v) 
+  
+  if(endsWith(v, "."))
+    v <- paste0(v, "9000")
+  
   numeric_version(v)
 }
 
--- a/R/conda.R
+++ b/R/conda.R
@@ -286,28 +286,17 @@ conda_install <- function(envname = NULL,
     }
   }
 
-  
   # delegate to pip if requested
   if (pip) {
-    conda_env_PATH_additions <- normalizePath(file.path(dirname(python), c(
-      "",
-      "Library/mingw-w64/bin",
-      "Library/usr/bin",
-      "Library/bin",
-      "Scripts",
-      "bin")
-      ), mustWork = FALSE)
     
-    withr::with_path(conda_env_PATH_additions, {
-      
-      result <- pip_install(
-        python = python,
-        packages = packages,
-        pip_options = pip_options,
-        ignore_installed = pip_ignore_installed
-      )
-      
-    })
+    result <- pip_install(
+      python = python,
+      packages = packages,
+      pip_options = pip_options,
+      ignore_installed = pip_ignore_installed,
+      conda = conda,
+      envname = envname
+    )
     
     return(result)
     
@@ -385,6 +374,14 @@ conda_version <- function(conda = "auto") {
   conda_bin <- conda_binary(conda)
   system2(conda_bin, "--version", stdout = TRUE)
 }
+
+
+numeric_conda_version <- function(conda = "auto") {
+  v <- conda_version(conda)
+  v <- sub("(conda) ([0-9.]+)", "\\2", v)
+  numeric_version(v)
+}
+
 
 #' @rdname conda-tools
 #' @export
@@ -583,4 +580,25 @@ conda_installed <- function() {
     return(FALSE)
   else
     return(TRUE)
+}
+
+
+conda_run <- function(cmd, args = c(), conda = "auto", envname = NULL,
+                      run_args = c("--no-capture-output"), ...) {
+
+  conda <- conda_binary(conda)
+  envname <- condaenv_resolve(envname)
+
+  if(numeric_conda_version(conda) < "4.9")
+    stopf(
+"`conda_run()` requires conda version >= 4.9.
+Run `miniconda_update('%s')` to update conda.", conda)
+
+  if(grepl("[/\\]", envname))
+    in_env <- c("--prefix", envname)
+  else
+    in_env <- c("--name", envname)
+
+  system2(conda, c("run", in_env, run_args,
+                   shQuote(cmd), args), ...)
 }
--- a/R/conda.R
+++ b/R/conda.R
@@ -286,15 +286,28 @@ conda_install <- function(envname = NULL,
     }
   }
 
+  
   # delegate to pip if requested
   if (pip) {
+    conda_env_PATH_additions <- normalizePath(file.path(dirname(python), c(
+      "",
+      "Library/mingw-w64/bin",
+      "Library/usr/bin",
+      "Library/bin",
+      "Scripts",
+      "bin")
+      ), mustWork = FALSE)
     
-    result <- pip_install(
-      python = python,
-      packages = packages,
-      pip_options = pip_options,
-      ignore_installed = pip_ignore_installed
-    )
+    withr::with_path(conda_env_PATH_additions, {
+      
+      result <- pip_install(
+        python = python,
+        packages = packages,
+        pip_options = pip_options,
+        ignore_installed = pip_ignore_installed
+      )
+      
+    })
     
     return(result)
     
--- a/R/conda.R
+++ b/R/conda.R
@@ -595,7 +595,7 @@ conda_run <- function(cmd, args = c(), conda = "auto", envname = NULL,
 Run `miniconda_update('%s')` to update conda.", conda)
 
   if(grepl("[/\\]", envname))
-    in_env <- c("--prefix", envname)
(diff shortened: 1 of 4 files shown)


R/conda.R
```diff
@@ -286,15 +286,28 @@ conda_install <- function(envname = NULL,
     }
   }
 
+  
   # delegate to pip if requested
   if (pip) {
+    conda_env_PATH_additions <- normalizePath(file.path(dirname(python), c(
```
kevinushey, line 292: Should we remove paths that don't exist? (e.g. if things happened to change in a future version of Anaconda)

Is there any way we can ask Anaconda itself what these PATH entries are?

R/conda.R
```diff
@@ -286,15 +286,28 @@ conda_install <- function(envname = NULL,
...
+      "bin")
+      ), mustWork = FALSE)
     
-    result <- pip_install(
-      python = python,
-      packages = packages,
-      pip_options = pip_options,
-      ignore_installed = pip_ignore_installed
-    )
+    withr::with_path(conda_env_PATH_additions, {
```
kevinushey, line 301: Can we explicitly supply `action = "prefix"` here, just to make it clear that this is our intention (put those at the front of the PATH)

R/conda.R
kevinushey: Also, should we modify the PATH in the same way when initializing a copy of Python that lives in a Conda environment? We do some PATH munging here:

https://github.com/rstudio/reticulate/blob/277f4109094d36aa9865c66b8c870d567febd3ca/R/config.R#L487-L529

I suspect it would be best if we shared code in those two locations.

kevinushey (changes requested):
Can you please also update NEWS?

R/conda.R
t-kalinowski: > Should we remove paths that don't exist? (e.g. if things happened to change in a future version of Anaconda)

I don't think this is a good idea. I can imagine a scenario where conda might create one of these directories as part of whatever it's doing. Also, it's harmless to have non-existent directories on the PATH.

> Is there any way we can ask Anaconda itself what these PATH entries are?

The only reliable way I think would be to shell out with a `conda activate <envname>` command and report back the PATH, but even that get's tricky today because `conda activate` will fail with this message: 
```
CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
To initialize your shell, run

    $ conda init <SHELL_NAME>

Currently supported shells are:
  - bash
  - fish
  - tcsh
  - xonsh
  - zsh
  - powershell

See 'conda init --help' for more information and options.

IMPORTANT: You may need to close and restart your shell after running 'conda init'.
```
The absence of `cmd.exe` from that list worries me; I think that's the default shell from R on windows.

Relevant: 
Default hook that get's run on 'conda init'
https://github.com/conda/conda/blob/master/conda/shell/etc/profile.d/conda.sh
Note there is a separate path on windows, maybe this PATH injection should be confined to windows only. 
https://github.com/conda/conda/blob/0b1312ce65bf0fbb8cbea3750f07c32a2492c57a/conda/shell/etc/profile.d/conda.sh#L14

(also, my heart goes out to whoever had to write this: https://github.com/conda/conda/blob/33a142c16530fcdada6c377486f1c1a385738a96/conda/shell/condabin/_conda_activate.bat)

I'll boot back into windows and experiment a little more with the suggested changes.

Have a nice weekend!

t-kalinowski:
@kevinushey Do we have any explicit version requirements for conda? I am considering launching the `python -m pip` command through `conda run`, (kindly suggested by @msarahan). However, some of the `conda run` flags require conda [4.9.0](https://github.com/conda/conda/blob/master/CHANGELOG.md). Does that mean we can't use it?

Another approach might be to use (exec-wrappers)[https://pypi.org/project/exec-wrappers/] (also suggested by @msarahan). Or perhaps we do something similar without taking the dependency. Maybe writing a temporary .bat/.sh script that just calls `<conda_base>\Scripts\activate.bat <env_path>; <cmd>` is the better long-term solution?

t-kalinowski:
I refactored this to run `pip` through [`conda run`](https://github.com/conda/conda/blob/master/conda/cli/main_run.py), which handles some tricky platform-specific and conda-specific [details](https://github.com/conda/conda/blob/0b1312ce65bf0fbb8cbea3750f07c32a2492c57a/conda/utils.py#L323). Ready for review.

t-kalinowski:
Bump: https://stackoverflow.com/questions/69336444/installing-keras-in-r-warning-pip-is-configured-with-locations-that-require-tl

R/conda.R
```diff
@@ -373,6 +375,14 @@ conda_version <- function(conda = "auto") {
   system2(conda_bin, "--version", stdout = TRUE)
 }
 
+
+numeric_conda_version <- function(conda = "auto") {
+  v <- conda_version(conda)
+  v <- sub("(conda) ([0-9.]+)", "\\2", v)
```
kevinushey, line 381: Do the parentheses in `(conda)` need to be escaped? Or are you intentionally creating (and discarding) this capture?

R/conda.R
```diff
@@ -571,3 +581,24 @@ conda_installed <- function() {
...
 }
+
+
+conda_run <- function(cmd, args = c(), conda = "auto", envname = NULL,
+                      run_args = c("--no-capture-output"), ...) {
+
+  conda <- conda_binary(conda)
+  envname <- condaenv_resolve(envname)
+
+  if(numeric_conda_version(conda) < "4.9")
```
kevinushey, line 592:
```suggestion
  if (numeric_conda_version(conda) < "4.9")
```

R/conda.R
```diff
@@ -373,6 +375,14 @@ conda_version <- function(conda = "auto") {
   system2(conda_bin, "--version", stdout = TRUE)
 }
 
+
+numeric_conda_version <- function(conda = "auto") {
+  v <- conda_version(conda)
+  v <- sub("(conda) ([0-9.]+)", "\\2", v)
+  numeric_version(v)
```
kevinushey, line 382: Is there any possibility that `conda --version` could return something like `conda 4.2-alpha` or something similar here? Or are we safe assuming it will be a parsable numeric version?

R/pip.R
```diff
@@ -19,17 +19,34 @@ pip_version <- function(python) {
...
+
+  # quote in case of version constraints like 'Pillow<8.3'
+  # but don't double quote if already quoted
+  packages <- shQuote(gsub("[\"']", "", packages))
   args <- c(args, packages)
 
-  # run it
-  result <- system2(python, args)
+  # figure out if we should go though conda_run()
+  if(conda == "auto") {
```
kevinushey, line 38:
```suggestion
  if (conda == "auto") {
```

R/pip.R
```diff
@@ -19,17 +19,34 @@ pip_version <- function(python) {
...
+    info <- python_info(python)
+    if(info$type != "conda" || numeric_conda_version(info$conda) < "4.9")
+      conda <- NULL
+  }
+
+  result <- if(is.null(conda) || isFALSE(conda))
```
kevinushey, line 44:
```suggestion
  result <- if (is.null(conda) || isFALSE(conda))
```

kevinushey (approved):

R/conda.R
t-kalinowski: The latter

R/conda.R
t-kalinowski: I tried to chase this down in the conda code base and it seems your intuition is right. As best as I can figure, the version string is (maybe always?) generated from the last git tag. Looking through the tags it seems that "{version}" can sometimes have a suffix like 'rc0', 'b0', 'a1', 'beta2, ... there doesn't appear to be systematic approach I can discern. Also, sometimes the version can sometimes be `"{version}.post{n_commits}+{commit_hash}"`, where {version} will be the last tag.

https://github.com/conda/conda/blob/c1579681d1468af3d1b4af3083bed33f8391e861/conda/_vendor/auxlib/packaging.py#L165
https://github.com/conda/conda/tags

The `"conda "` prefix however seems hardcoded and will probably always be there. 
https://github.com/conda/conda/blob/c1579681d1468af3d1b4af3083bed33f8391e861/conda/cli/conda_argparse.py#L36

I will try to make the numeric version parser more robust.

t-kalinowski:
the new `numeric_conda_version()` gives output like:
```r
numeric_conda_version(version_string = "conda 4.6.0"                 ) # '4.6.0.1'
numeric_conda_version(version_string = "conda 4.6.0b0"               ) # '4.6.0.1.3'
numeric_conda_version(version_string = "conda 4.6.0rc1"              ) # '4.6.0'
numeric_conda_version(version_string = "conda 4.6.0rc1.post3+64bde06") # '4.6.0.0'
```
The function should always return a valid `numeric_version` object, but comparisons between beta|rc|dev conda versions won't always be correct (e,g. '4.6.0' > '4.6.0b0' will return FALSE). I don't think that matters too much here, I can't imagine there are many people who are managing dev versions of conda through reticulate.

R/conda.R
```diff
@@ -373,6 +375,30 @@ conda_version <- function(conda = "auto") {
...
+  v <- sub("^conda ", "", v) # drop hardcoded prefix
+  
+  # https://github.com/conda/conda/blob/c1579681d1468af3d1b4af3083bed33f8391e861/conda/_vendor/auxlib/packaging.py#L142
+  # if dev version string: "{0}.post{1}+{2}".format(version, post_commit, hash)
+  v <- sub("\\.post(\\d)\\+.+$", ".\\1", v)
+
+  # substitute rc|beta|alpha|whatever suffix with .
+  v <- sub("[A-Za-z]+", ".", v) 
+  
+  if(endsWith(v, "."))
```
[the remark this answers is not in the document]
kevinushey: Looks like it came with R 3.3.0: https://cran.r-project.org/doc/manuals/r-release/NEWS.3.html

I'd rather we just use our own version here (unless I'm misremembering and we already have that in reticulate)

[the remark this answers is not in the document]
t-kalinowski: Seems there is already a reticulate NEWS entry about this already:
>- Use `grepl()` rather than `endsWith()` for compatibility with R <= 3.2

t-kalinowski:
Quick update, `conda run` can sometimes fail cryptically. I filed an issue here:
https://github.com/conda/conda/issues/10972

Issue #1058, opened 2021-09-26
`import_from_path()` with `delay_load = TRUE` fails


Karel-Kroeze:
`import_from_path()` with `delay_load = TRUE` fails;

```R
modules_path <- system.file("python", package = package)
anonymize <<- reticulate::import_from_path(
    "anonymize",
    modules_path,
    delay_load = TRUE
)
```

leads to an error;
```
 Error: $ operator is invalid for atomic vectors 
3. delay_load$before_load %||% function() {} 
2. import_from_path_delayed(module, path, convert, delay_load) 
1. reticulate::import_from_path("anonymize", modules_path, delay_load = TRUE)
```

Because `reticulate:::import_from_path_delayed()` assumes `delay_load` to be a list. The documentation for the `import_` class of functions specifically states that `delay_load = TRUE` is "normally the right choice", and in fact specifically gives an example that will fail: 
```R
reticulate::import_from_path(<module>, path = path, delay_load = TRUE)
```

PR #1059, opened 2021-09-26
resolve non-list values for `delay_load` in `import_from_path_delayed`

Karel-Kroeze:
resolves the issue reported in #1058, where the (recomended) value of `TRUE` would cause errors. 

The fix is a bit ugly, as reticulate is now resolving single arguments to a list in `import_from_path_delayed`, and spreading that list back out to single arguments in `import` itself. Some refactoring might be in order, but I don't feel confident enough to make more sweeping changes. 

I aimed to cover the documented options for `delay_load` (function, list and boolean), and fail early for other inputs.

### DIFF
--- a/R/import.R
+++ b/R/import.R
@@ -175,6 +175,22 @@ import_from_path <- function(module,
 
 import_from_path_delayed <- function(module, path, convert, delay_load) {
   
+  # resolve delay_load
+  # single function, maps to on_load
+  if(is.function(delay_load)){
+    delay_load <- list(on_load = delay_load)
+  }
+  
+  # TRUE translates to an empty list of hooks
+  if(identical(delay_load, TRUE)){
+    delay_load <- list()
+  }
+  
+  # anything else not a list is an error
+  if(!is.list(delay_load)){
+    stop("delay_load should be a boolean, a single function, or a list of functions.")
+  }
+  
   .before_load <- delay_load$before_load %||% function() {}
   .on_load     <- delay_load$on_load %||% function() {}
   .on_error    <- delay_load$on_error %||% function(error) {}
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks for the pull request! We require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

R/import.R
```diff
@@ -175,6 +175,22 @@ import_from_path <- function(module,
 
 import_from_path_delayed <- function(module, path, convert, delay_load) {
   
+  # resolve delay_load
+  # single function, maps to on_load
+  if(is.function(delay_load)){
```
kevinushey, line 180:
```suggestion
  if (is.function(delay_load)) {
```

R/import.R
```diff
@@ -175,6 +175,22 @@ import_from_path <- function(module,
...
+    delay_load <- list(on_load = delay_load)
+  }
+  
+  # TRUE translates to an empty list of hooks
+  if(identical(delay_load, TRUE)){
```
kevinushey, line 185:
```suggestion
  if (identical(delay_load, TRUE)) {
```

R/import.R
```diff
@@ -175,6 +175,22 @@ import_from_path <- function(module,
...
+    delay_load <- list()
+  }
+  
+  # anything else not a list is an error
+  if(!is.list(delay_load)){
```
kevinushey, line 190:
```suggestion
  if (!is.list(delay_load)) {
```

kevinushey:
Thanks!

PR #1065, opened 2021-10-06
Add instruction for source installation

henrykironde:
Example on how to install packages from the repository


kevinushey:
If this works, it feels like a lucky accident rather than something explicitly supported. I'd recommend just calling `system("pip install <...>")` at that point.

PR #1073, opened 2021-10-23
Changes needed for branch main

henrykironde:
Moving from master to main with out breaking third party tools.

Create branch main then set it as default, and merge related changes.


kevinushey:
Thanks for the PR! I've made the necessary changes to move to the main branch here.

Issue #1006, opened 2021-05-17
Support for pipenv?


cboettig:
Would your team consider implementing more direct support for users using `pipenv` to manage the python environment?  

[Pipenv](https://github.com/pypa/pipenv) is a widely used approach maintained by the Python Packaging Authority, PyPa, for managing virtualenvs for different projects automatically by using a simple `Pipfile` in the project directory.  It is built on the virutalenv system, so we can still access pipenv environments with `reticulate` already, but it is cumbersome and lacks the syncing features we get from using pipenv directly, such as:

- You no longer need to use pip and virtualenv separately. They work together.
- Managing a requirements.txt file can be [problematic](https://kenreitz.org/essays/2016/02/25/a-better-pip-workflow)
- Hashes are used everywhere, always. Security. Automatically expose security vulnerabilities.

- pipenv also lets us install & switch between different python versions via pyenv, and a given project can be set to use a specific python version easily.

I know this also overlaps a bit with your `renv` package, and I'd be happy to transfer this issue there if you think that's better.  In general I think `pipenv` approach with a `Pipfile` is very much in the same spirit as `renv`, and is a bit cleaner way to do what `renv` tries to do with regards to python via `requirements.txt`.  Another nice feature of the `pipenv` approach is that you have consistent behavior in 'pure python' in the project, which automatically loads the Pipenv file, as you would in the R environment, instead of having to manually do something like `source renv/python/bin/activate` in the shell to make renv's python env accessible on the terminal.  Anyway I think @kevinushey would like `pipenv` and so just wanted to toss it out there!

kevinushey:
The impression I get is that [poetry](https://python-poetry.org/) is a fair bit more popular than Pipenv nowadays though.

That said, I would be open to supporting both, depending on what "supporting" means -- what exactly would that entail? What are you hoping `reticulate` and / or `renv` could do for you here?

cboettig:
Thanks @kevinushey , I'll take a closer look at `poetry`.  It doesn't seem to support a way to initialize with an arbitrary version of python though, like `pipenv --python 3.7`.  (My students and I are often switching between projects that require different versions of python, principally since we have some stuff that we want to run on the latest python and some stuff like tensorflow 1.x-based modules that don't support it!)   

[...]

PR #1074, opened 2021-10-24
Add poetry project awareness

AkhilGNair:
Hi!

Took a stab at using `poetry` as the venv provider after following the conversation in #1006. 
I have used the method shown to use Pipenv as the venv provider from [this commit](https://github.com/rstudio/reticulate/commit/3675ff51f5478554bf2b7572a4325496f3ef67c4).

Screenshot from 2021-10-24 12-21-29

The environment seems to respect the provided venv. 

To detect a poetry environment, I check whether there is a `pyproject.toml` which declares `poetry` as one of the tools, as the `pyproject.toml` is the only file `poetry init` initially creates.

```
> poetry init --no-interaction --quiet
> cat pyproject.toml 
[tool.poetry]
name = "myproject"
version = "0.1.0"
description = ""
authors = ["Akhil <...>"]

[tool.poetry.dependencies]
python = "^3.8"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

I dynamically register the `TOML` parser (as done in `blogdown`) as it doesn't seem like it should be a package dependency, but rather just warn the user if it doesn't exist and they want to use `poetry`. Hope this is ok.

Strange note - when I run `reticulate::py_config()` in a terminal R session, I do not have to unset `RETICULATE_PYTHON`, but I do in Rstudio. I wonder if this is the same with the `pipenv` implementation?

Thanks for the great package :)

### DIFF
--- a/R/config.R
+++ b/R/config.R
@@ -199,6 +199,15 @@ py_discover_config <- function(required_module = NULL, use_environment = NULL) {
     
   }
   
+  # if we're working within a project that containers a pyproject.toml, then
+  # check if that project uses poetry as a tool.
+  # if so, use the copy of Python associated with the poetry project
+  if (is_poetry_project()) {
+    python <- poetry_python()
+    config = python_config(python, required_module, forced = "poetry")
+    return(config)
+  } 
+  
   # if we're working within a project that contains a Pipfile, then
   # use the copy of Python associated with that pipenv
   pipfile <- pipenv_pipfile_path()
--- a/R/poetry.R
+++ b/R/poetry.R
@@ -1,0 +1,55 @@
+error_no_toml_parser = function(e) {
+  stop("Please install 'RcppTOML' to parse the a poetry pyproject.toml", call. = FALSE)
+}
+
+pyproject_parser = function() {
+  # From: https://github.com/rstudio/blogdown/blob/main/R/utils.R#L343
+  tryCatch(getFromNamespace('parseTOML', 'RcppTOML'), error = error_no_toml_parser)
+}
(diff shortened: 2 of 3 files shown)


kevinushey:
Thanks for the PR! I recently made some changes that I believe obviate what you put together in this PR (sorry; I did this before I saw the PR!)

For example:

```
poetry init -n
poetry run python -c 'import sys; print(sys.version)'
R -s -e 'reticulate::py_config()'
```

gives me:

```
$ R -s -e 'reticulate::py_config()'
python:         /Users/kevin/Library/Caches/pypoetry/virtualenvs/testing-J3SPl9Qp-py3.9/bin/python
libpython:      /opt/homebrew/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/config-3.9-darwin/libpython3.9.dylib
pythonhome:     /Users/kevin/Library/Caches/pypoetry/virtualenvs/testing-J3SPl9Qp-py3.9:/Users/kevin/Library/Caches/pypoetry/virtualenvs/testing-J3SPl9Qp-py3.9
virtualenv:     /Users/kevin/Library/Caches/pypoetry/virtualenvs/testing-J3SPl9Qp-py3.9/bin/activate_this.py
version:        3.9.9 (main, Nov 21 2021, 03:16:13)  [Clang 13.0.0 (clang-1300.0.29.3)]
numpy:           [NOT FOUND]

NOTE: Python version was forced by Poetry
```

PR #1080, opened 2021-11-09
Streamline ways to invoke `pip --ignore-installed` in `py_install()`

t-kalinowski:
in `py_install()`, respect both `pip_ignore_installed` and `ignore_installed`, in both conda and virtualenv python installations.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -50,6 +50,6 @@ Suggests:
     rmarkdown,
     testthat
 LinkingTo: Rcpp
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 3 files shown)


kevinushey:
Sorry for taking so long to look at this. Can you clarify the intended difference between `ignore_installed` and `pip_ignore_installed` here?

t-kalinowski:
I have this sentence in the roxygen block 
>  (`ignore_installed` is an alias for `pip_ignore_installed`, `pip_ignore_installed` takes precedence).

Maybe I should bring it forward?

kevinushey:
Gotcha. Effectively, they do the same thing; it just depends on which one you want to use (e.g. muscle memory from `conda_install()` versus `virtualenv_install()`)?

PR #1090, opened 2021-11-24
fix install_miniconda

sebffischer:
`miniconda_installer_arch` currently does not have access to the variable `info` which causes the error 
`Error: Error in miniconda_installer_arch() : object 'info' not found`


kevinushey:
Thanks for catching that!

Issue #1087, opened 2021-11-17
install_miniconda() fails on arm64 macOS


kevinushey:
With:

```
PackagesNotFoundError: The following packages are not available from current channels:

  - python=3.6

```

We should consider updating the version requirement here:

https://github.com/rstudio/reticulate/blob/dd6d6e8139dcfc4ce19e6b7150406a6f4e1af747/R/miniconda.R#L340-L343

kevinushey:
To make matters worse, after a failed install in this way one cannot directly reinstall Miniconda:

```
> reticulate:::install_miniconda()
Error: Miniconda is already installed at '/Users/kevin/Library/r-miniconda'
```

kevinushey:
Closed by https://github.com/rstudio/reticulate/pull/1094 (thanks @t-kalinowski!)

PR #1094, opened 2021-12-03
bump miniconda default python version to 3.8

t-kalinowski:
Right now, `install_miniconda()` fails at the `r-reticulate` environment creation step on an M1 Mac.
This is because the minimum (native Arm) python version available from `conda` is 3.8. 
Rather than installing different python versions on Arm Macs vs other platforms, I think it's better to bump the python version everywhere. 

There is also a case to be made for bumping to python 3.9, since that's the default python version that comes with miniconda, (meaning, smaller overall install and less to download).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@
   `pyproject.toml` file, `reticulate` will attempt to find and use the virtual
   environment managed by Poetry for that project. (#1031)
   
+- `reticulate::install_miniconda()` default python version for the default 
+  `r-reticulte` miniconda environment changes from 3.6 to 3.8.
+  
 - `reticulate::install_miniconda()` now prefers installing the latest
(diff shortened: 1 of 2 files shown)


t-kalinowski:
Just noticed there is an open issue: #1087

NEWS.md
```diff
@@ -6,6 +6,9 @@
   `pyproject.toml` file, `reticulate` will attempt to find and use the virtual
   environment managed by Poetry for that project. (#1031)
   
+- `reticulate::install_miniconda()` default python version for the default
```
kevinushey, line 9: I'd suggest something like "The default version of Python used for the `r-reticulate` Miniconda environment installed via `reticulate::install_miniconda()` has been changed from 3.6 to 3.8."

kevinushey (approved):

kevinushey:
Thanks!

Issue #1083, opened 2021-11-15
Windows-installed python is not correctly detected


cderv:
I am on a Windows 10 computer and I have the version of python installed through Windows Store. 
This is from here: https://docs.python.org/3/using/windows.html#the-microsoft-store-package 

This is what works on my system and where it is located
```powershell
❯ py.exe -0p
Installed Pythons found by C:\Windows\py.exe Launcher for Windows
 -3.9-64        C:\Users\chris\AppData\Local\Programs\Python\Python39\python.exe *
```

Having this version works in terminal and also with Quarto 
````powershell
❯ quarto check jupyter

[>] Checking Python 3 installation....OK
      Version: 3.9.5
      Path: C:/Users/chris/AppData/Local/Programs/Python/Python39/python.exe
      Jupyter: 4.7.1
      Kernels: python3

[>] Checking Jupyter engine render....OK
````

However, with reticulate, this version is not correctly found
````r
> reticulate::py_discover_config()
Error: could not find a Python environment for C:/Users/chris/AppData/Local/Programs/Python/Python39/python.exe
Run `rlang::last_error()` to see where the error occurred.
> rlang::last_error()
<error/rlang_error>
Error: could not find a Python environment for C:/Users/chris/AppData/Local/Programs/Python/Python39/python.exe
Backtrace:
 1. reticulate::py_discover_config()
 2. reticulate:::python_config(python_version, required_module, python_versions)
 3. reticulate:::python_munge_path(python)
 4. reticulate:::python_info(python)
 5. reticulate:::stopf(...)
Run `rlang::last_trace()` to see the full context.
````

It seems like `python_info()` expect an environment or else, and in any case doesn't know how to retrieve information on this one. 

Creating a virtualenv  manually and activating it solve the issue with the installation 
```r
# reticulate::virtualenv_create("r-reticulate")
> reticulate::use_virtualenv("r-reticulate")
master* > reticulate::py_discover_config()
python:         C:/Users/chris/Documents/.virtualenvs/r-reticulate/Scripts/python.exe
libpython:      C:/Users/chris/AppData/Local/Programs/Python/Python39/python39.dll
pythonhome:     C:/Users/chris/Documents/.virtualenvs/r-reticulate
version:        3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)]
Architecture:   64bit
[...]

PR #1097, opened 2021-12-06
Fix binding to system python

t-kalinowski:
Follow up to #1083

cc: @cderv.
Note, even with this fix, binding to the Python installed from the Microsoft Store fails for me, albeit for a different reason now.

Now I get this error message:
```
Error in py_initialize(config$python, config$libpython, config$pythonhome,  :
  C:/Program Files/WindowsApps/PythonSoftwareFoundation.Python.3.9_3.9.2544.0_x64__qbz5n2kfra8p0/python39.dll - Access is denied.
Calls: print ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
In addition: Warning message:
In file.info(x, extra_cols = FALSE) :
  cannot open file 'C:\Users\kalin\AppData\Local\Microsoft\WindowsApps\python.exe': The file cannot be accessed by the system
Execution halted
```

After some googeling this appears to be not easily resolvable: https://stackoverflow.com/questions/56974927/permission-denied-trying-to-run-python-on-windows-10

I tested this locally on Windows 10 with python's:
- installed via installer from www.python.org
- `r-reticulate` condaenv created by `reticulate::install_miniconda()`
- base condaenv of `r-reticulate`
- Microsoft Store Python (which errors for me)

and on WSL with `/usr/bin/python3`, `r-reticulate`, and base of `r-reticulate`.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -511,6 +511,9 @@ conda_binary <- function(conda = "auto") {
 
   conda <- normalizePath(conda, winslash = "/", mustWork = FALSE)
   
+  if(!grepl("^conda", basename(conda)))
+    stop("Supplied path is not a conda binary: ", sQuote(conda, FALSE))
+  
   # if the user has requested a conda binary in the 'condabin' folder,
   # try to find and use its sibling in the 'bin' folder instead as
   # we rely on other tools typically bundled in the 'bin' folder
--- a/R/conda.R
+++ b/R/conda.R
@@ -850,3 +850,37 @@ conda_info <- function(conda = "auto") {
   json <- system2(conda, c("info", "--json"), stdout = TRUE)
   jsonlite::parse_json(json, simplifyVector = TRUE)
 }
+
+is_conda_python <- function(python) {
+  root <- if (is_windows())
+    dirname(python)
+  else
+    dirname(dirname(python))
+  
+  file.exists(file.path(root, "conda-meta"))
+}
+
+
+get_python_conda_info <- function(python) {
+  stopifnot(is_conda_python(python))
+  
+  root <- if (is_windows()) 
+    dirname(python)
+  else
+    dirname(dirname(python))
+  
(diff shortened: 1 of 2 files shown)


kevinushey:
The Microsoft store Python is basically broken; we explicitly try not to use it.

cderv:
> The Microsoft store Python is basically broken; we explicitly try not to use it.

Ok that is news to me. I am not a heavy python users, but it works fine as of now for me. So I understand, I should revert to not using it at all ? I'll keep that in mind for next issues. 

For now, the new dev version fixes the original issues, and the version is correctly detected without error
```r
> reticulate::py_discover_config()
python:         C:/Users/chris/AppData/Local/Programs/Python/Python39/python.exe
libpython:      C:/Users/chris/AppData/Local/Programs/Python/Python39/python39.dll
pythonhome:     C:/Users/chris/AppData/Local/Programs/Python/Python39
version:        3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)]
Architecture:   64bit
numpy:          C:/Users/chris/AppData/Local/Programs/Python/Python39/Lib/site-packages/numpy
numpy_version:  1.20.3
````

t-kalinowski:
Glad it's working finally. 
I also second the recommendation to not use the Microsoft Store Python. 
I've run into a variety of different annoyances with it. Anything related to install packages via pip is made difficult through that odd indirection (good luck trying to find cli scripts that `setup.py` tries to install for python `C:/Program Files/WindowsApps/PythonSoftwareFoundation.Python.3.9_3.9.2544.0_x64__qbz5n2kfra8p0`). Also, other issues where it couldn't acquire a COM port, or run an http service. Maybe some of those have been resolved since I last tried to use it, but from prior experiences I've concluded it's a broken installation that shouldn't exist.

PR #1103, opened 2021-12-17
Feature: `repl_python()` magics

t-kalinowski:


R/repl.R
```diff
@@ -227,6 +252,37 @@ repl_python <- function(
...
+      # expand any "!!" as system commands that capture output
+      trimmed <- gsub("!!", "%system ", trimmed)
+
+      # user intends to capture output from system command in var
+      # e.g.:   x = !ls
+      if(grepl("^[[:alnum:]_.]\\s*=\\s*!", trimmed))
+        trimmed <- sub("=\\s*!", "= %system ", trimmed)
+
+      # magic
+      if(grepl("^%", trimmed)) {
```
kevinushey, line 264: Nit: please use a space after `if` (here and elsewhere).

```suggestion
      if (grepl("^%", trimmed)) {
```

R/repl.R
```diff
@@ -355,3 +411,156 @@ repl_python <- function(
...
+  command <- str_drop_prefix(command, "%")
+
+  m <- str_split1_on_first(command, "\\s+")
+  cmd <- m[1]
+  args <- m[-1]
+
+
+  if(cmd == "pwd") {
+    if(length(args))
+      stop("%pwd magic takes not arguments, received: ", command)
```
kevinushey, line 427:
```suggestion
      stop("%pwd magic takes no arguments, received: ", command)
```

kevinushey (commented):
LGTM! Only nitpick is really around style.

Do you think it'd be possible to test this? You can drive `repl_python()` non-interactively with e.g.

```
repl_python(input = "x = 1")
```

so this might be a way to test that the REPL does what we expect with some kinds of magic.

Also; should we have an option to enable / disable this just in case?

R/repl.R
```diff
@@ -37,6 +37,31 @@
...
+#'  - `%pip ...` executes pip for the active python.
+#'  - `%load`, `%loadpy`, `%run` executes a python file.
+#'  - `%system`, `!!` executes a system command and capture output
+#'  - `%env`: read current environment variables.
+#'    - `%env name`: read environment variable 'name'.
+#'    - `%env name=val`, `%env name val`: set environment variable 'name' to 'val'.
+#'       `val` elements in `{}` are interpolated using f-strings (required Python >= 3.6).
+#'  - `%cd <dir>` change working directory.
+#'     - `%cd -`: change to previous working directory.
+#'     - `%cd -3`: change to 3rd most recent working directory.
```
kevinushey, line 56: Do these operations work regardless of how the working directory was changed? Or is it only specifically for directory changes done via `%cd`? (I suspect the latter; if so, I think that's worth documenting here.)

R/repl.R
```diff
@@ -37,6 +37,31 @@
...
+#'     - `%cd -foo/bar`: change to most recent working directory matching `"foo/bar"` regex.
+#'  - `%pwd`: print current working directory.
+#'  - `%dhist`: print working directory history.
+#'
+#' Additionally, the output of magic commands can be captured in a variable, e.g.:
+#'  - `x = !ls`
```
kevinushey, line 62: Something to document: what exactly is captured here? Both stdout and stderr? Just stdout? Do we split on newlines for the user, or just capture all output as-is?

R/repl.R
t-kalinowski: Of course! Any chance we could make Cmd+shift+A do this in RStudio? It's nice when I can let the tools do the formatting... (e.g., in python I typically use black)...

t-kalinowski:
Some of the `repl_python()` tests pass fail when executed via `devtools::test()`. The tests pass when executed interactively, as a stand alone chunk (“Check" button in the IDE), or on GHA CI. Do you know what might be the cause?

kevinushey:
I do not. :( Are you able to reproduce in an interactive session with `devtools::test()`? If so, you might be able to use the R debugger to learn more.

t-kalinowski:
Thank you! That helped me track it down. The active binding `py` wasn't correctly working from the test execution environment. I modified the test to avoid using `py`.

kevinushey:
Thanks! Do you think this is ready for merge?

t-kalinowski:
Yes!

PR #1104, opened 2021-12-17
add `ipython()`

t-kalinowski:


t-kalinowski:
For the record, I don't know yet if this is a good idea. At this point it's mostly a proof of concept to experiment with and see how well it works.

kevinushey:
I'm going to leave this PR for after the next CRAN release.

t-kalinowski:
Marking this as internal until IPython can work well in RStudio console (right now the ipython repl only works well in regular terminal, not in RStudio console)

PR #1106, opened 2021-12-17
use 'required = TRUE' by default

kevinushey:
This PR makes `use_python()` and friends use `required = TRUE` by default.

For packages calling `use_*()` APIs in their `.onLoad()` hooks, we default to `FALSE` (as per the old behavior) just to reduce the chances of breaking backwards compatibility.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,10 @@
 
 - The development branch for `reticulate` has moved to the "main" branch.
 
+- `reticulate::use_python()` and friends now assume `required = TRUE` by
+  default. For backwards compatibility, when `use_python()` is called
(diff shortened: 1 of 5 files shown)


t-kalinowski:
I think this is a great idea! 
Is it common to use `use_*` python functions in package `.onLoad()` hooks? My impression is that most people use the `use_*` in scripts and projects, but not in R packages. Maybe change the default everywhere?

kevinushey:
> Is it common to use use_* python functions in package .onLoad() hooks? My impression is that most people use the use_* in scripts and projects, but not in R packages. Maybe change the default everywhere?

Unfortunately it is, as I discovered in revdep checks last time I submitted `reticulate` to CRAN...

PR #1108, opened 2021-12-20
use PyObject_Size when computing object length

kevinushey:
Closes https://github.com/rstudio/reticulate/issues/992.


kevinushey:
I wonder if some of our `length()` method definitions are redundant here...

```
$ rg -F 'length.'
R/array.R
47:"length.numpy.ndarray" <- function(x) {

R/python.R
472:length.python.builtin.tuple <- function(x) {
501:length.python.builtin.list <- function(x) {

R/conversion.R
243:length.pandas.core.series.Series <- function(x) {
402:length.pandas.core.frame.DataFrame <- function(x) {
430:length.scipy.sparse.base.spmatrix <- function(x) {

R/python-dict.R
42:length.python.builtin.dict <- function(x) {
```

I think we could use `PyList_Size()` and `PyTuple_Size()` since those methods avoid the need to try and dispatch to the associated `__len__` method, although I'm not sure how expensive that is in practice.

t-kalinowski:
LGTM! 
One thing that I wonder looking at this: When python is not initialized or the objects is a null pointer, would it be better to raise an error instead of silently returning `0`? I can imagine the latter could lead to some hard-to-find errors.

Looking at the other length S3 methods, I think it's fine to keep them. We already are using `PyTuple_Size` and `PyDict_Size` where appropriate. We could probably get rid of the one for numpy arrays, but either way the tradeoff is negligible. For pandas objects, we deviate from the meaning of `len()` in python to match R's semantics: `len(pandas_df)` returns the number of rows in python, while `length(pandas_df)` returns the number of columns in R.

t-kalinowski (approved):

kevinushey:
> One thing that I wonder looking at this: When python is not initialized or the object is a null pointer, would it be better to raise an error instead of silently returning 0? I can imagine the latter could lead to some hard-to-find errors.

I think the main reason is because `length()` gets invoked behind the scenes in a lot of places, e.g. by RStudio (which calls `str()` on objects in the global environment).

t-kalinowski:
Following up on this after a closer reading of https://docs.python.org/3/c-api/object.html?highlight=pyobject_size#c.PyObject_Size, 
should we be checking for failure by checking if `value == -1`, instead of calling out to `PyErr_Occurred`? It seems like we may inadvertently catch the wrong error here sometimes.

PR #1111, opened 2022-01-03
Add vignette: Python Primer

t-kalinowski:


vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+```
+
+-   Many numpy operations modify the array in place! This is surprising
+    to R users, who are used to the convenience and safety of R's
+    copy-on-modify semantics. Unfortunately, there is no simple scheme
+    or naming convention you can rely on to quickly determine if a
+    particular method modifies in-place or creates a new array copy. The
+    only reliable way is to consult the
+    [documentation](https://numpy.org/doc/stable/reference/index.html#reference),
+    and conduct small experiments at the `reticulate::python_repl()`.
```
kevinushey, line 781:
```suggestion
    and conduct small experiments at the `reticulate::repl_python()`.
```

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+    print("self.a_method() was called with", self)
+
+inst = MyClass()
+inst.a_method()
+
+MyClass.a_method()     # error, missing required argument `self`
+MyClass.a_method(inst) # identical to inst.a_method()
+```
+
+Other dunder's worth knowing about are:
```
skeydan, line 552: nice listing!

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+packages, and R users can access objects from R packages via `library()`
+or `::`. In Python, authors bundle code into *modules*, and users access
+modules using `import`. Consider the line:
+
+```{python}
+import numpy
+```
+
+This statement has Python go out to the file system, find an installed
+Python module named 'numpy', load it (commonly meaning: evaluate it's
```
kevinushey, line 795:
```suggestion
Python module named 'numpy', load it (commonly meaning: evaluate its
```

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+<https://docs.Python.org/3/>
+
+<https://docs.Python.org/3/library/index.html>
+
+To learn Python more fully, the built-in official tutorial is also
+excellent and comprehensive (but does require a time commitment to get
+value out of it) <https://docs.Python.org/3/tutorial/index.html>
+
+Finally, don't forget to solidify your understanding by conducting small
+experiments at the `reticulate::python_repl()`.
```
kevinushey, line 918:
```suggestion
experiments at the `reticulate::repl_python()`.
```

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+R is the common language of statistics and data analysis, but it is not
+the only language on the scene. You may find yourself occasionally
+wanting to read and understand some Python, or even port some Python to
+R. This guide is designed to enable you to do these tasks as quickly as
+possible. As you'll see, R and Python are similar enough that it's
+possible to do these tasks, without necessarily learning all of Python.
+
+### Whitespace
+
+Whitespace matters in Python. In R you groups expressions into a code
```
kevinushey, line 35:
```suggestion
Whitespace matters in Python. In R, expressions are grouped into a code
```

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+block with `{}`, in Python you do that by making the expressions share
```
kevinushey, line 36:
```suggestion
block with `{}`. In Python, that is done by making the expressions share
```

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+  cat("This is another expression. \n")
+}
+```
+
+The equivalent in Python:
+
+```{python}
+if True:
+  print("This is one expression.")
+  print("This is another expression.")
```
kevinushey, line 52: Do we need to talk about indenting with spaces versus tabs here?

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+
+### Decorators
+
+Decorators are just functions that take a function as an argument and
+then return a function. Any function can be invoked as a decorator with
+the `@` syntax, which is just sugar for this simple action:
+
+```{python}
+def my_decorator(func):
+  func.x = "a decorator modified this function by adding an attribute `x`"
```
skeydan, line 636: could this be made even simpler? Perhaps by removing the added attribute? So that it's even more clear that it's just a function call, basically...?

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+-   `@property`, which automatically calls a class method when the
+    attribute is accessed (similar to `makeActiveBinding()` in R).
+
+### Defining Generators with `yield`.
+
+Generators are special Python functions that contain one or more `yield`
+statements. As soon as `yield` is included in a code block passed to
+`def`, the semantics change substantially. You're no longer defining a
+mere function, but a generator constructor! In turn, calling a generator
+constructor creates a generator object, which is just another type of
```
skeydan, line 658: generator object == generator _class_, is that correct?

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+iter(my_generator) == my_generator == my_generator.__iter__()
+
+# step through it like any other iterator
+next(my_generator)
+my_generator.__next__() # next() is just sugar for calling the dunder
+next(my_generator)
+next(my_generator)
+```
+
+Encountering `yield` is like hitting the pause button on a functions
```
skeydan, line 689: nice analogy!

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+Python lists are typically created using bare brackets `[]`. The Python
+built-in `list()` function is more of a coercion function, closer in
+spirit to R's `as.list()`. The most important thing to know about Python
+lists is that they are modified in place. Note in the example below that
+`y` reflects the changes made to `x`, because the underlying list object
+which both symbols point to is modified in place.
+
+```{python}
+x = [1, 2, 3]
+y = x
```
kevinushey, line 75:
```suggestion
y = x   # `y` and `x` now refer to the same list!
```

kevinushey (commented):
Very nice! I added a couple of small comments; let me know if you think they're worth tweaking.

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+computing features in Python are generally provided by third party
+packages (*modules*, in Python parlance). The closest equivalent to an R
+numeric vector is a numpy array, or sometimes, a Python list of numerics
+(some Pythonistas might argue for `array.array()` here, but that's so
+rarely encountered in actual Python code we don't mention it further).
+Teaching the NumPy interface is beyond the scope of this tutorial, but
+it's worth pointing out some potential tripping hazards for users
+accustomed to R arrays:
+
+-   When indexing into multidimensional numpy arrays, trailing
```
skeydan, line 757: I was wondering if it'd make sense to even start talking about numpy here ... on the other hand, that point is totally worth making, - again one super confusing thing when seen from "outside" ...

skeydan (approved):
This is amazing!! Addresses exactly the weird/confusing things as seen from "outside", I think ...
One thing one could do is reorder a bit, so that "beginner stuff" comes before "intermediate" throughout ... and perhaps add a TOC, so that more intermediate users see that even though there is some basic content upfront, there's more "nerdy" stuff coming along soon ... :-)

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+
+```{r}
+library(reticulate)
+l <- r_to_py(list(1, 2, 3))
+it <- as_iterator(l)
+
+iter_next(it)
+iter_next(it)
+iter_next(it)
+iter_next(it, completed = "StopIteration")
```
dfalbel, line 318: Maybe also mention the `coro::loop` syntax here?

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+The equivalent in Python:
+
+```{python}
+if True:
+  print("This is one expression.")
+  print("This is another expression.")
+
+```
+
+### Container Types
```
dfalbel, line 56: Would it make sense to talk about `reticulate` casting rules in this section. Ie, what happens with R lists/named lists when they are passed to python.

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+# a dict comprehension built from x, where the key is a string.
+# Python's str() is like R's as.character()
+d = {str(element) : element + 100
+     for element in x}
+d
+```
+
+### Defining Functions with `def`
+
+Python functions are defined with the `def` statement. The syntax for
```
dfalbel, line 362: I think it's worth mentioning that Python functions require a `return` statement.

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+```{python}
+x = (1, 2, 3)
+a, *the_rest = x
+a
+the_rest
+```
+
+#### Dictionaries
+
+Dictionaries are most similar to R's named lists. They are a container
```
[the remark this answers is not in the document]
t-kalinowski: Thank you, that's the better comparison.

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+```
+
+### What about R vectors?
+
+R is a language designed for numerical computing first. Vector data
+types are baked deep into the R language, to the point that the language
+doesn't even distinguish scalars from vectors. By comparison, numerical
+computing features in Python are generally provided by third party
+packages (*modules*, in Python parlance). The closest equivalent to an R
+numeric vector is a numpy array, or sometimes, a Python list of numerics
```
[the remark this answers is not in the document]
t-kalinowski: I reordered some the sections so modules are discussed before vectors, and reworded this paragraph a little too (Sigrid also flagged this)

vignettes/python-primer.Rmd
t-kalinowski: Thanks! I added two sentences on tabs vs spaces.

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+)
+```
+
+```{r setup}
+library(reticulate)
+```
+
+## Primer on Python for R users
+
+R is the common language of statistics and data analysis, but it is not
```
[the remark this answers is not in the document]
t-kalinowski: perhaps "R is the preferred language of ..." ?

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+a, b, c = x
+a
+b
+c
+```
+
+(You can access similar unpacking behavior from R using
+`` zeallot::`%<-%` ``).
+
+Tuple unpacking can occur in multiple contexts, such as iteration:
```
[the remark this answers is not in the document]
t-kalinowski: Good suggestion! I added another example about nested unpacking

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+d2 = d
+d
+d["key1"]
+d["key3"] = 3
+d2 # modified in place!
+```
+
+Unlike R's named lists, you cannot index into a dictionary with an
+integer to get an item at a specific index position. Dictionaries are
+*unordered* containers. (However---beginning with Python 3.7,
```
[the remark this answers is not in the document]
t-kalinowski: It's guaranteed, in the language spec now!

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+dictionaries do preserve the item insertion order).
+
+```{python, error = TRUE}
+d = {"key1": 1, "key2": 2}
+d[1] # error
+```
+
+A container that closest matches the semantics of R's named list is the
```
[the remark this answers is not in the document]
t-kalinowski: It was much more common in Python 2 code, but since that change in Py 3.7, it seems to be waning in favor. I added a link to https://docs.python.org/3/library/collections.html#collections.OrderedDict, for the motivated readers :)

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+
+for key, value in d.items():
+  print(key, ":", value)
+```
+
+#### Comprehensions
+
+Comprehensions are special syntax that allow you to construct a
+container like a list or a dict, while also executing a small operation
+or single expression on each element. You can think of it as special
```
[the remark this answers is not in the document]
t-kalinowski: Interesting! using the word 'set' here feels a little dangerous, since it's not `set` type.... (Also, as I think on it, math sets can be infinite, but python comprehensions don't support `break`....)

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+my_function <- function(name = "World") {
+  cat("Hello", name, "\n")
+}
+
+my_function()
+my_function("Friend")
+```
+
+(Note for advanced R users: Python has no equivalent of R's argument
+"promises". Function argument default values are evaluated once, when
```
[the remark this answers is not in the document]
t-kalinowski: I think the convention is to typically set the default value to `None`, and then change it in the function body if it needs to be some kind of complicated or mutable object.

vignettes/python-primer.Rmd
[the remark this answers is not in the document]
skeydan: yeah, well ... for me it's still too strong, but on the other hand, this is directed to R users :-) Also, there's a "but" coming up, so why not, I guess :-)

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+symbol, `MyClass`. First note the strong naming convention, classes are
+typically `CamelCase`, and functions are typically `snake_case`. After
+defining `MyClass`, you can interact with it, and see that it has type
+`'type'`. Calling `MyClass()` creates a new object **instance** of the
+class, which has type `'MyClass'` (ignore the `__main__.` prefix for
+now). The instance prints with its memory address, which is a strong
+hint that it's common to be managing many instances of a class.
+
+In the first example, we defined an empty `class`, but when we inspect
+it we see that it already comes with a bunch of attributes (`dir()` in
```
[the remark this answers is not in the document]
t-kalinowski: All methods are accessible as attributes (said another way, a method is any attribute that was a function when the `class` was constructed, since that's when all the `self` partialing magic happens).... Not sure where to work that in exactly. Perhaps a (new) subsection on attributes?

vignettes/python-primer.Rmd
```diff
@@ -0,0 +1,920 @@
...
+print("MyClass is finished being created")
+
+instance = MyClass()
+print(instance)
+
+instance2 = MyClass()
+print(instance2)
+```
+
+A few things to note:
```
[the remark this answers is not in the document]
t-kalinowski: Interesting! I purposely took it a little slow here. I was worried it maybe a little surprising for R users that a `class` code block is executed only once. Otherwise, people might expect something similar to `def` semantics, where the code block is executed on each call. Also, once you know a class definition is executed only once, the fact that anything defined in the class body is shared by all the instances is a realization that naturally follows....

vignettes/python-primer.Rmd
t-kalinowski: Are you suggesting an identity function? I was looking for some minimal, simple-to-understand thing that demonstrates that decorators typically do *something* ... assigning an attribute seemed like the most accessible operation that came to mind. I don't want to go full bore and start talking about closures or all the machinery in `functools`...

vignettes/python-primer.Rmd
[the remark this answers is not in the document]
skeydan: that would be great!

vignettes/python-primer.Rmd
t-kalinowski: I *think* so. 
My impression is that the terminology gets a little imprecise here the way it's used in the broader world. In many places people seem to interchangeably use the bare word "generator" for both the 'generator constructor' and the 'generator iterator' (which self-identifies as a 'generator object' when printed at the REPL, and `<class 'generator'>` when queried with `type()`).

vignettes/python-primer.Rmd
t-kalinowski: Thanks! Kevin also honed in on the way Numpy is introduced. I re-ordered sections so that module are talked about first, before vectors, and I word-smithed this a little too.

vignettes/python-primer.Rmd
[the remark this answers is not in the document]
skeydan: hm, seems like I think differently about this ... I mean, a function _definition_ is also executed just once, right?
But maybe it's just that my first programming language really was Java that makes me find it obvious that a class is defined once ...

vignettes/python-primer.Rmd
skeydan: yeah, I see, terminology stuff ... here too, probably preferences are determined by how one was "socialized" ... plus, the terms you use there are pretty clear

vignettes/python-primer.Rmd
t-kalinowski: Thanks! I added a mention about coro::loop in a later "final remarks on iteration" section.

vignettes/python-primer.Rmd
skeydan: yeah... maybe i was thinking of something as simple as calling` tf_function()` ;-))) But you're right, adding an attriute makes sense.

vignettes/python-primer.Rmd
t-kalinowski: Thanks, I added a sentence in the section on dictionaries about this.

vignettes/python-primer.Rmd
t-kalinowski: Thanks! Good point.

vignettes/python-primer.Rmd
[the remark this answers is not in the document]
t-kalinowski: Re TOC, I think that pkgdown automatically generates a TOC sidebar from the headings, so people will be able to hop around. I also added a teaser sentence in the first paragraph mentioning some of the more nerdy stuff to come...

I'm afraid my ability to distinguish between what's "beginner stuff" and "intermediate" isn't so great here... I placed things in an order that makes sense for me but I'm open to suggestions for reordering.

t-kalinowski:
Thank you Kevin, Sigrid and Daniel for the excellent feedback!

kevinushey:
@t-kalinowski: do you think this is ready to be merged?

t-kalinowski:
I think so! We still need to add it to the pkgdown.yml and maybe a link in the readme, do you think that should happen on this branch, or a separate rebuild-website branch?

kevinushey:
> I think so! We still need to add it to the pkgdown.yml and maybe a link in the readme, do you think that should happen on this branch, or a separate rebuild-website branch?

I'll update `pkgdown.yml` post-merge and then we can rebuild the website following the CRAN release.

PR #1114, opened 2022-01-07
ensure length() returns a numeric value

kevinushey:
This PR ensures that `length.python.builtin.object()` always returns a numeric value, in line with what users would normally expect from the `length()` generic.

`py_len()` continues to emit an error for Python objects without a defined length.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -245,8 +245,8 @@ py_list_length <- function(x) {
     .Call(`_reticulate_py_list_length`, x)
 }
 
-py_len_impl <- function(x) {
-    .Call(`_reticulate_py_len_impl`, x)
+py_len_impl <- function(x, defaultValue) {
+    .Call(`_reticulate_py_len_impl`, x, defaultValue)
 }
 
 readline <- function(prompt) {
--- a/R/python.R
+++ b/R/python.R
@@ -493,7 +493,7 @@ py_len <- function(x) {
     return(0L)
 
   # delegate to C++
-  py_len_impl(x)
+  py_len_impl(x, NULL)
 
 }
 
@@ -504,7 +504,20 @@ length.python.builtin.list <- function(x) {
 
 #' @export
 length.python.builtin.object <- function(x) {
-  py_len(x)
+
+  # return 0 if Python not yet available
+  if (py_is_null_xptr(x) || !py_available())
+    return(0L)
+
+  # Not all Python objects have a defined length, but almost all objects in R
+  # do. To reflect this behavior, we treat Python's 'None' as a zero-length
+  # object (similar to R's NULL), and treat any other objects without a
+  # `__len__` method as though they were scalar objects.
+  if (inherits(x, "python.builtin.NoneType"))
+    return(0L)
+  else
+    py_len_impl(x, 1L)
+
 }
 
 #' Convert to Python Unicode Object
--- a/R/python.R
+++ b/R/python.R
@@ -478,22 +478,31 @@ length.python.builtin.tuple <- function(x) {
 
 #' Length of Python object
 #'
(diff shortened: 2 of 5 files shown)


kevinushey:
Some considerations to keep in mind here:

- The notion of the length of an R object, and the notion of the length of a Python object, do not really map to the same meaning. In R, `length()` almost always means the vector-length of an object, whereas in Python the concept of length is more general (and many objects do not have a defined length).

- We should ensure that existing code continues to work; e.g. a lot of code might test for "empty-ness" with something like `if (length(x)) { ... }`. IMHO, it's important that we don't break code that might've been performing tests with a Python object here.

- In R, `length()` should virtually always return a numeric value; in Python, `len()` might give an error for objects without a defined length. IMHO, given that `length()` in R usually means the "vector-length" of an object, we should treat Python objects without `__len__` as scalar objects (vector-objects of length one), and so consider `length(x) == 1L` in that case.

t-kalinowski:
Another alternative might be to return `NA_integer_` instead of `1L` when `PyObject_Size()` raises an exception because no `__len__()` method for the Python object exists.

It seems like the primary motivator for returning `1L` is to prevent introducing a breaking change to R code using the idiom `if(length(x))` as a shorthand for `if(!is.null(x) || !(is.vector(x) && identical(length(x), 0L)))`. This seems like a relatively minor and easy-to-fix breaking change, and considering the tradeoffs, I think it's worth it. (I've already implemented the required changes in {keras} and {tensorflow})

I think the proposed approach in this PR is not ideal because it leaves the meaning of `length` for python objects inconsistent: Sometimes `length(x)` will mean `` x$`__len__`() `` and sometimes it will mean `as.integer(inherits(x, "python.builtin.object"))`. This will make it hard to reason about code, and it also makes R code vulnerable to breaking changes from Python as Python objects gain `__len__()` methods.

I would prefer a consistent meaning for `length()` of Python objects. Two approaches, in order of preference:
1) `length()` is equivalent to `__len__()`, returns `NA_integer_` if the python object has no `__len__` method. 
2) `length()` always return `1L`. (there is precedence for this approach too. A close reading of `?length` has this snippet:
> ... All other objects (including functions) have length one.

kevinushey:
I don't think we can use `NA` as the return value. A lot of existing code in the RStudio IDE calls `length()` on R objects in the environment pane (or, call functions that might call `length()`), and returning `NA` can cause issues. For example:

```
x <- structure(list(), class = "foo")
registerS3method("length", "foo", function(x) NA)
```

If I set this, then every time I refresh the Environment pane, I see this in the console:

```
Warning in str.default(val) : 'str.default': 'le' is NA, so taken as 0
```

Arguably, this is an issue that should be fixed in the IDE, but the problem is that many users will run with older versions of RStudio but the latest-available versions of R packages, and so this would effectively become a change that regressed behavior in the RStudio IDE.

Packages using `reticulate` which need better guarantees about the "real" length of a Python object can still use `py_len()`.

kevinushey:
> length() always return 1L. (there is precedence for this approach too. A close reading of ?length has this snippet:

This would be consistent with the current behavior in that `length(<python.object>)` currently returns 2; that is, the length of the R environment hosting that object.

kevinushey:
> This will make it hard to reason about code, and it also makes R code vulnerable to breaking changes from Python as Python objects gain __len__() methods.

For what it's worth, we do already treat some builtin Python objects specially, e.g.

```
> library(reticulate)
> x <- py_eval("[1, 2, 3]", convert = FALSE)
> length(x)
[1] 3
```

So it seems reasonable to me to extend this to all objects implementing `__len__`.

t-kalinowski:
I think it's OK to return `1L`. It maximizes convenience and minimizes breaking changes. 👍

PR #1123, opened 2022-01-13
Feature/py bool

t-kalinowski:


R/use_python.R
```diff
@@ -147,6 +147,17 @@ use_condaenv <- function(condaenv = NULL, conda = "auto", required = NULL) {
 
   required <- required %||% use_python_required()
 
+  if (grepl("[/\\]", condaenv) &&
+      grepl("^python[0-9.]*(exe)?", basename(condaenv))) {
```
kevinushey, line 151: Can you use `python_info()` here?

R/use_python.R
```diff
@@ -147,6 +147,17 @@ use_condaenv <- function(condaenv = NULL, conda = "auto", required = NULL) {
...
+    # path to python binary provided
+    python <- condaenv
+    if (!is_conda_python(python))
+      warning("path supplied to use_condaenv() is not a conda python:\n\t",
+              path)
+    use_python(python, required = required)
```
kevinushey, line 157: I wonder if we should return the value of `use_python()` (here and elsewhere in the function)

t-kalinowski:
By the way, I only recently learned the github UI lets you hide whitespace changes!
image

PR #1136, opened 2022-01-18
Don't render vignette on CRAN

### LINKED ISSUE #1135: resolve vignette build issues
Seen on macOS and Windows.

```
checking re-building of vignette outputs ... [6s/8s] WARNING
Error(s) in re-building vignettes:
--- re-building ‘arrays.Rmd’ using rmarkdown
--- finished re-building ‘arrays.Rmd’

--- re-building ‘calling_python.Rmd’ using rmarkdown
--- finished re-building ‘calling_python.Rmd’

--- re-building ‘package.Rmd’ using rmarkdown
--- finished re-building ‘package.Rmd’

--- re-building ‘python_dependencies.Rmd’ using rmarkdown
--- finished re-building ‘python_dependencies.Rmd’

--- re-building ‘python_packages.Rmd’ using rmarkdown
--- finished re-building ‘python_packages.Rmd’

--- re-building ‘python_primer.Rmd’ using rmarkdown
Quitting from lines 155-176 (python_primer.Rmd)
Error: processing vignette 'python_primer.Rmd' failed with diagnostics:
invalid syntax (<string>, line 12)
--- failed re-building ‘python_primer.Rmd’

--- re-building ‘r_markdown.Rmd’ using rmarkdown
--- finished re-building ‘r_markdown.Rmd’

--- re-building ‘versions.Rmd’ using rmarkdown
--- finished re-building ‘versions.Rmd’

SUMMARY: processing the following file failed:
  ‘python_primer.Rmd’
```

cc: @t-kalinowski; this is why we do not run Python code on CRAN. :-)

t-kalinowski:
Closes #1135

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -16,6 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      NOT_CRAN: "true"
     steps:
       - uses: actions/checkout@v2
 
--- a/vignettes/python_primer.Rmd
+++ b/vignettes/python_primer.Rmd
@@ -20,10 +20,6 @@ knitr::opts_chunk$set(
 )
 ```
 
-```{r setup}
-library(reticulate)
-```
-
 ## Primer on Python for R users
 
 You may find yourself wanting to read and understand some Python, or
--- a/vignettes/python_primer.Rmd
+++ b/vignettes/python_primer.Rmd
@@ -14,7 +14,8 @@ editor_options:
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>",
-  eval = reticulate::py_numpy_available(initialize = TRUE)
+  eval = isTRUE(as.logical(Sys.getenv("NOT_CRAN"))) &&
+    reticulate::py_numpy_available(initialize = TRUE) 
 )
 ```
 
--- a/vignettes/python_primer.Rmd
+++ b/vignettes/python_primer.Rmd
@@ -14,8 +14,9 @@ editor_options:
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>",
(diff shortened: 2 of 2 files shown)


kevinushey:
Is pre-rendering the vignette locally an option?

kevinushey:
I might be misunderstanding. I guess vignettes are _always_ built and submitted locally, but CRAN still tries to check that it can successfully rebuild those vignettes themselves (or run their R code) as part of R CMD check?

So this change would basically stop CRAN from trying to run our R code during checks. Does that sound right?

t-kalinowski:
Just to make sure we're talking about the same thing, are you suggesting converting the vignette to a static document that doesn't evaluate code when CRAN builds it, (like the current Readme.md), or are you suggesting additional guards on the vignette render, like quarto's "freeze" parameter? I think you're suggesting the former.

You are right that with the approach in this PR, CRAN would not evaluate any of the code chunks. That's not ideal because then the output from the code chunks on the CRAN rendered vignette is also missing. Converting to a static document on the other hand is not great either, because we lose the ability to make easily make edits + and rerender in the future.

kevinushey:
> Just to make sure we're talking about the same thing, are you suggesting converting the vignette to a static document that doesn't evaluate code when CRAN builds it, (like the current Readme.md), or are you suggesting additional guards on the vignette render, like quarto's "freeze" parameter? I think you're suggesting the former.

I'm thinking of the latter -- I'm hoping we can use the same `.Rmd` source, but have the final vignette that's published to CRAN be rendered locally and submitted as part of the package tarball. In other words, have the same local vignette workflow, but without the worry that CRAN might execute the vignette code and run into issues.

kevinushey:
I also (out of some well-founded paranoia) don't like the idea that we're so obviously circumventing the CRAN check machinery here.

I'm going to prepare a separate candidate PR.

kevinushey:
Superceded by https://github.com/rstudio/reticulate/commit/40bb05babb61f86d832f2b592e6fbdf2f69176be.

PR #1138, opened 2022-01-19
check for compatible version of python when building vignette

kevinushey:
Closes https://github.com/rstudio/reticulate/issues/1135.


t-kalinowski:
I think the current version of the vignette needs Python 3.8: 
https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

PR #1142, opened 2022-01-24
Refactor Python Exception Handling

### LINKED ISSUE #1113: Show python exception message by default
Consider a simple misconfigured keras model like this:
```R
model <- keras_model_sequential(input_shape = c(28 * 28)) %>%
  layer_dense(units = 512, activation = "relu") %>%
  layer_dense(units = 10, activation = "softmax")

model %>% compile(
  optimizer = "rmsprop",
  loss = "categorical_crossentropy",
  metrics = c("accuracy"))


model %>% fit(k_random_uniform(c(4, 28*29)), 
              k_random_uniform(c(4,10)))
```
In Rstudio, this is what's visible after evaluating this chunk:
image

This is problematic because it gives a lot of output, but nothing useful to the user about what caused the error. Expanding the stacktrace helps a little, but the actual message is in a big unformatted and unevaluated jumble of a template string (which is sometimes also truncated, though not in this example) 

image

Printing the actual exception shows a different message, which I'm guessing was generated by the IDE overwriting the original python error:
image

Note the actionable information that should be shown prominently to the user is usually available 
by running `for(m in reticulate::py_last_error()) cat(m)`, which would show this as a standalone, prominent line: 
```
ValueError: Input 0 of layer "sequential" is incompatible with the layer: expected shape=(None, 784), found shape=(None, 812)
```

It might even make sense to show the full python stack track as well (e.g., from a different example of an error message):

image

t-kalinowski:
closes #1113


R/python.R
```diff
@@ -1480,3 +1480,102 @@ py_register_load_hook <- function(module, hook) {
...
+#'     -1L
+#'   })
+#' }
+#'
+#' safe_len(py_eval("object"))
+#' }
+#'
+#' @export
+py_last_error <- function(exception) {
+  if(!missing(exception)) {
```
kevinushey, line 1540:
```suggestion
  if (!missing(exception)) {
```

Please remember to be consistent with indentation + style.

R/python.R
```diff
@@ -1480,3 +1480,102 @@ py_register_load_hook <- function(module, hook) {
...
+    # set as the last exception
+    if(inherits(exception, "py_error"))
```
kevinushey, line 1542:
```suggestion
    if (inherits(exception, "py_error"))
```

R/python.R
```diff
@@ -1480,3 +1480,102 @@ py_register_load_hook <- function(module, hook) {
...
+    if (!is.null(exception) && !inherits(exception, "python.builtin.Exception"))
+      stop("`exception` must be NULL, a `py_error`, or a 'python.builtin.Exception'")
+
+    on.exit(.globals$py_last_exception <- exception)
+    return(invisible(.globals$py_last_exception))
+  }
+
+  e <- .globals$py_last_exception
+
+  if(is.null(e))
```
kevinushey, line 1554:
```suggestion
  if (is.null(e))
```

R/python.R
```diff
@@ -1480,3 +1480,102 @@ py_register_load_hook <- function(module, hook) {
...
+    return(NULL)
+
+  if(!py_available() || py_is_null_xptr(e)) {
```
kevinushey, line 1557:
```suggestion
  if (!py_available() || py_is_null_xptr(e)) {
```

R/python.R
```diff
@@ -1480,3 +1480,102 @@ py_register_load_hook <- function(module, hook) {
...
+                     collapse = "")
+  )
+  attr(out, "exception") <- e
+  class(out) <- "py_error"
+  out
+}
+
+#' @export
+print.py_error <- function(x, ...) {
+  cat(x$message)
```
kevinushey, line 1580:
```suggestion
  cat(x$message, sep = "\n")
```

So that a trailing newline is included (f one isn't already).

src/python.cpp
```diff
@@ -603,63 +505,64 @@ std::string py_fetch_error() {
...
-  }
+  for (Py_ssize_t i = 0, n = PyList_Size(formatted); i < n; i++)
+    oss << as_std_string(PyList_GetItem(formatted, i));
 
-  // get final error string
   std::string error = oss.str();
-  s_lastError.setMessage(error);
 
-  // return error
+  int max_msg_len(Rf_asInteger(Rf_GetOption1(Rf_install("warning.length"))));
```
kevinushey, line 539: Out of an abundance of caution, I would recommend PROTECTing the result of `Rf_GetOption1` before calling `Rf_asInteger()`, then UNPROTECTing after.

src/python.cpp
```diff
@@ -478,122 +478,24 @@ bool traceback_enabled() {
...
+// save it in R reticulate:::.globals$py_last_exception
+// Return a short string suitable for Rcpp::stop(),
+// Either the exception message, or a truncated version with
+// user instructions for where to see the full exception.
 std::string py_fetch_error() {
+  PyObject *excType, *excValue, *excTraceback;
+  PyErr_Fetch(&excType, &excValue, &excTraceback);  // we now own the PyObjects
+  PyErr_NormalizeException(&excType, &excValue, &excTraceback);
+  if (excTraceback != NULL) {
+    PyException_SetTraceback(excValue, excTraceback);
```
kevinushey, line 491: Can you confirm that this increases the Python reference count on excTraceback?

src/python.cpp
```diff
@@ -603,63 +505,64 @@ std::string py_fetch_error() {
...
-  if (traceback_enabled()) {
-    std::size_t n = traceback.size();
-    if (n > 0) {
-      oss << "\n\nDetailed traceback:\n";
-      for (std::size_t i = 0; i < n; i++)
-        oss << traceback[i];
-    }
-  }
+  for (Py_ssize_t i = 0, n = PyList_Size(formatted); i < n; i++)
+    oss << as_std_string(PyList_GetItem(formatted, i));
```
kevinushey, line 535: nit: good practice to document as a reminder places where we're using Python APIs which provide borrowed references

src/python.cpp
t-kalinowski: [confirmed](https://github.com/python/cpython/blob/0575551f69ba9c999835bfb176a543d468083c03/Objects/exceptions.c#L273)

R/python.R
t-kalinowski: I think exception messages returned by `traceback.format_exception` are promised to end with a "\n", but I agree an additional \n just-in-case is a good thing.

t-kalinowski:
Thanks @kevinushey

PR #1148, opened 2022-02-01
Feature/default-format-method

t-kalinowski:
- Reroute default `print()` to go through `format()` instead of `str()`.
- All Python objects gain a default `format()` method.
- `py_str()` default method no longer strips the object memory address.


kevinushey (approved):
LGTM! Do you want to update NEWS as well?

PR #1157, opened 2022-02-06
Feature/print calls py_repr

t-kalinowski:
In Python, objects auto-printed at the REPL are formatted using the builtin `repr()`.
This PR changes the print behavior in R so that, by default, the same output is observed when printing an object from the R or Python REPL.

My understanding is that the spirit of the default Python `str()` method is that it's closer to the mark when the user wants something for pretty printing, while `repr()` is intended to be less ambiguous and more informative during interactive sessions, and for simple objects `x == eval(repr(x))` should be true.

Some objects define both `__str__` and `__repr__` methods with just this type of split in purpose. For example, `tensorflow.TensorShape` objects: 
image


kevinushey (approved):
LGTM!

PR #1158, opened 2022-02-09
Fix typo in maintainer details

jeroen:
@t-kalinowski R does not seem to support comma's in the name it self.


t-kalinowski:
Thank you!

Issue #1167, opened 2022-02-28
use_python no longer works with only a directory without a file


V41U:
Our R package uses reticulate internally to wrap our Python package into R for customers. For the last 3 years this was done using:
`pythonpath = system.file("extdata/python", "python.exe", package="<our-package>", mustWork = TRUE)`
`pythondir = dirname(pythonpath)`
`use_python(pythondir)`

We are shipping an adjusted embedded Python in this 'extdata/python' directory. 

Now, a customer has reported that they cannot install our package with the following error: 
**Error in system2(command = python, args = shQuote(script), stdout = TRUE,  :
  '"C:/Users/dummy/Documents/R/win-library/4.1/<our-package>/extdata/python"' not found**

We have lookinto this and found the following:
We have always only passed the directory without the file name included. With the most recent version of reticulate this no longer works. 
This is not a big issue for new customers because they will all already download a patched version. But all existing customers who want to install our R package and haven't already installed older versions of reticulate will now encounter this issue. 

To reproduce:
Install the latest version of reticulate and call `use_python()` with the directory where a python.exe is located. 
This will only work if the it is called like `use_python('<dir>/python.exe'` instead of `use_python('<dir>')`

In version 1.20 it was working as expected (that's the version I had installed previously). When updating to 1.24 this no longer works. 

My question is: Is this an intended behavior change or will this be reverted?

t-kalinowski:
Hi, thanks for filing. 
Reading the docs for `use_python() it says:

> `python`  The path to a Python binary.

I think the fact that it previously worked with a path to the directory was more a happy accident then the intention. However, it seems like a reasonable request to restore the previous behavior and update the docs. (In version 1.20, did it look for `<dir>/bin/python` on macOS and Linux?).

Would you like to submit a PR?

PR #1168, opened 2022-03-01
Paths used in 'use_python()' are now normalized before being used to configure Python

V41U:
I have reported the issue #1167 

In previous versions of reticulate it was possible to call use_python with the directory of the python binary without including the executable. This was tested in Windows. 

In newer versions (at least 1.24) this no longer works. This is due to calling `python_config_impl` with the 'unnormalized' path. Previously, the path has been normalized before proceeding further in the code. 

With this change, the given python path within `use_python` is normalized before working with it. This should restore the previous (undocumented) behavior of `use_python("C:/Users/dummy/Documents/R/win-library/4.1//extdata/python")` which currently prompts an error like:
**Error in system2(command = python, args = shQuote(script), stdout = TRUE, :
'"C:/Users/dummy/Documents/R/win-library/4.1//extdata/python"' not found**

### DIFF
--- a/R/use_python.R
+++ b/R/use_python.R
@@ -73,6 +73,9 @@ use_python <- function(python, required = NULL) {
   if (required && !file_test("-f", python) && !file_test("-d", python))
     stop("Specified version of python '", python, "' does not exist.")
 
+  # ensure that the python path is normalized as expected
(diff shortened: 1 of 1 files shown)


t-kalinowski:
Thank you!

PR #1173, opened 2022-03-07
Fix py_to_r for scipy>=1.8.0

rcannood:
## Problem description

SciPy 1.8.0 seems to have deprecated some of its data structures. As a result, class names like `scipy.sparse.csc.csc_matrix` have been renamed to `scipy.sparse._csc.csc_matrix`. As such, `py_to_r()` doesn't work properly anymore:

```r
devtools::install_github("rstudio/reticulate")

library(reticulate)

r_mat <- Matrix::rsparsematrix(4, 4, .1)
# 3 x 3 sparse Matrix of class "dgCMatrix"
# 
# [1,] . . .   
# [2,] . . .   
# [3,] . . 0.47

py_mat <- r_to_py(r_mat)
# <4x4 sparse matrix of type '<class 'numpy.float64'>'
#         with 2 stored elements in Compressed Sparse Column format>

py_to_r(py_mat)
# <4x4 sparse matrix of type '<class 'numpy.float64'>'
#        with 2 stored elements in Compressed Sparse Column format>

class(py_mat)
# [1] "scipy.sparse._csc.csc_matrix"        "scipy.sparse._compressed._cs_matrix" "scipy.sparse._data._data_matrix"     "scipy.sparse._base.spmatrix"        
# [5] "scipy.sparse._data._minmax_mixin"    "scipy.sparse._index.IndexMixin"      "python.builtin.object"  
```

If I manually apply the correct `py_to_r()` function, it also doesn't work because `dim.scipy.sparse._csc.csc_matrix()` does not exist:
```r
reticulate:::py_to_r.scipy.sparse.csc.csc_matrix(py_mat)
# Error in validObject(.Object) : 
#   invalid class “dgCMatrix” object: Dim slot must have length 2
```

## Proposed solution
This PR implements a potential fix which works for scipy<1.8.0 and scipy>=1.8.0 alike.

Proposed changes:
- f67c6d9: Added additional S3 methods for the new class names of deprecated SciPy matrices
- 09be1e5: While toying with some test data, I noticed that a matrix like `Matrix::Matrix(.1, nrow = 1, sparse = TRUE)` couldn't be converted from R to Python. I added a few tests to detect this issue.
- 15c7dcf: Added a fix for the above-mentioned issue by making sure that the data and indices are converted to an np-array.
- 192d013: Small fix to unit tests to allow them to work for both scipy<1.8.0 and scipy>=1.8.0.

I hope this is of help to the reticulate community :)


t-kalinowski:
Thank you for putting this together!

One thought I have while looking at this: if scipy sparse matrices continues to move around across scipy versions, it might make sense to use `register_class_filter()` and give them a stable class like `"scipy.sparse_matrix"`.

PR #1185, opened 2022-04-04
Run CI against several explicit python versions

clbarnes:
Follow NEP29[1], the de facto standard of the scientific python ecosystem,
for supported versions, but more backwards compatibility could be added.

1: https://numpy.org/neps/nep-0029-deprecation_policy.html


t-kalinowski:
Thank you! I agree this is a good idea.

PR #1188, opened 2022-04-14
Feature/ensure gil

### LINKED ISSUE #208: reticulate + rpy2 sharing R process

t-kalinowski:
closes #208

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,10 @@
 # reticulate 1.25  (UNRELEASED)
 
-- Fixed an issue where reticulate would fail to bind to an Anaconda 
-  base environment on Windows. 
+- Fixed an issue where reticulate would fail if R was running embedded under rpy2.
+  reticulate now ensures the Python GIL is acquired before calling into Python.
+
+- Fixed an issue where reticulate would fail to bind to an Anaconda
+  base environment on Windows.
 
 - All commands that create, modify, or delete a Python environment now echo
   the system command about to be executed. Affected:
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -64,6 +64,10 @@ main_process_python_info <- function() {
     .Call(`_reticulate_main_process_python_info`)
 }
 
+py_clear_error <- function() {
+    invisible(.Call(`_reticulate_py_clear_error`))
+}
+
 py_initialize <- function(python, libpython, pythonhome, virtualenv_activate, python3, interactive, numpy_load_error) {
     invisible(.Call(`_reticulate_py_initialize`, python, libpython, pythonhome, virtualenv_activate, python3, interactive, numpy_load_error))
 }
@@ -248,10 +252,6 @@ readline <- function(prompt) {
 
 py_register_interrupt_handler <- function() {
     invisible(.Call(`_reticulate_py_register_interrupt_handler`))
-}
-
-py_clear_error <- function() {
-    invisible(.Call(`_reticulate_py_clear_error`))
 }
 
 py_interrupts_pending <- function(reset) {
--- a/src/RcppExports.cpp
+++ b/src/RcppExports.cpp
@@ -128,6 +128,15 @@ BEGIN_RCPP
     return rcpp_result_gen;
 END_RCPP
 }
+// py_clear_error
+void py_clear_error();
+RcppExport SEXP _reticulate_py_clear_error() {
+BEGIN_RCPP
+    Rcpp::RNGScope rcpp_rngScope_gen;
+    py_clear_error();
+    return R_NilValue;
+END_RCPP
+}
 // py_initialize
 void py_initialize(const std::string& python, const std::string& libpython, const std::string& pythonhome, const std::string& virtualenv_activate, bool python3, bool interactive, const std::string& numpy_load_error);
 RcppExport SEXP _reticulate_py_initialize(SEXP pythonSEXP, SEXP libpythonSEXP, SEXP pythonhomeSEXP, SEXP virtualenv_activateSEXP, SEXP python3SEXP, SEXP interactiveSEXP, SEXP numpy_load_errorSEXP) {
@@ -660,15 +669,6 @@ BEGIN_RCPP
     return R_NilValue;
 END_RCPP
 }
-// py_clear_error
-void py_clear_error();
-RcppExport SEXP _reticulate_py_clear_error() {
-BEGIN_RCPP
-    Rcpp::RNGScope rcpp_rngScope_gen;
-    py_clear_error();
-    return R_NilValue;
-END_RCPP
-}
 // py_interrupts_pending
 bool py_interrupts_pending(bool reset);
 RcppExport SEXP _reticulate_py_interrupts_pending(SEXP resetSEXP) {
@@ -693,6 +693,7 @@ static const R_CallMethodDef CallEntries[] = {
     {"_reticulate_r_to_py_impl", (DL_FUNC) &_reticulate_r_to_py_impl, 2},
     {"_reticulate_py_activate_virtualenv", (DL_FUNC) &_reticulate_py_activate_virtualenv, 1},
     {"_reticulate_main_process_python_info", (DL_FUNC) &_reticulate_main_process_python_info, 0},
+    {"_reticulate_py_clear_error", (DL_FUNC) &_reticulate_py_clear_error, 0},
     {"_reticulate_py_initialize", (DL_FUNC) &_reticulate_py_initialize, 7},
     {"_reticulate_py_finalize", (DL_FUNC) &_reticulate_py_finalize, 0},
     {"_reticulate_py_is_none", (DL_FUNC) &_reticulate_py_is_none, 1},
@@ -739,7 +740,6 @@ static const R_CallMethodDef CallEntries[] = {
     {"_reticulate_py_bool_impl", (DL_FUNC) &_reticulate_py_bool_impl, 1},
     {"_reticulate_readline", (DL_FUNC) &_reticulate_readline, 1},
     {"_reticulate_py_register_interrupt_handler", (DL_FUNC) &_reticulate_py_register_interrupt_handler, 0},
-    {"_reticulate_py_clear_error", (DL_FUNC) &_reticulate_py_clear_error, 0},
     {"_reticulate_py_interrupts_pending", (DL_FUNC) &_reticulate_py_interrupts_pending, 1},
     {NULL, NULL, 0}
 };
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -2011,9 +2011,11 @@ SEXP main_process_python_info_unix() {
   }
 
   // read libpython file path
-  if (strcmp(python_path.c_str(), dinfo.dli_fname) == 0) {
+  if (strcmp(dinfo.dli_fname, python_path.c_str()) == 0 ||
+      strcmp(dinfo.dli_fname, "python") == 0) {
     // if the library is the same as the executable, it's probably a PIE.
     // Any consequent dlopen on the PIE may fail, return NA to indicate this.
+    // when R is embedded by rpy2, dli_fname can be 'python'
     info["libpython"] = Rf_ScalarString(R_NaString);
   } else {
     info["libpython"] = dinfo.dli_fname;
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -51,6 +51,17 @@ void loadSymbol(void* pLib, const std::string& name, void** ppSymbol)
 #endif
 }
 
+class GILScope {
+ private:
+  PyGILState_STATE gstate;
+
+ public:
+  // aquire GIL on creation, release on destruction
+  GILScope() { gstate = PyGILState_Ensure(); }
+
+  ~GILScope() { PyGILState_Release(gstate); }
+};
+
 // track whether we have required numpy
 std::string s_numpy_load_error;
 bool haveNumPy() {
@@ -64,17 +75,17 @@ bool requireNumPy() {
 }
 
 bool isPyArray(PyObject* object) {
-  if (!haveNumPy())
-    return false;
-  else
-    return PyArray_Check(object);
+  if (!haveNumPy()) return false;
+
+  GILScope scope;
+  return PyArray_Check(object);
 }
 
 bool isPyArrayScalar(PyObject* object) {
-  if (!haveNumPy())
-    return false;
-  else
-    return PyArray_CheckScalar(object);
+  if (!haveNumPy()) return false;
+
+  GILScope scope;
+  return PyArray_CheckScalar(object);
 }
 
 // static buffers for Py_SetProgramName / Py_SetPythonHome
@@ -138,6 +149,7 @@ PyObject* py_capsule_new(SEXP object) {
 }
 
 PyObject* py_get_attr(PyObject* object, const std::string& name) {
+  GILScope scope;
 
   if (PyObject_HasAttrString(object, name.c_str()))
     return PyObject_GetAttrString(object, name.c_str());
@@ -165,8 +177,10 @@ public:
 
   virtual ~PyPtr()
   {
-    if (object_ != NULL)
+    if (object_ != NULL) {
+      GILScope scope;
       Py_DecRef((PyObject*) object_);
+    }
   }
 
   operator T*() const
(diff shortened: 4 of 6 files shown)


kevinushey:
Was using `reticulate_types.h` + redefining `BEGIN_RCPP` not an option? I'm not a fan of needing to sprinkle `GILScope` everywhere.

t-kalinowski:
I looked at it briefly, but it didn't seem safe to modify `BEGIN_RCPP` 
(please correct me if BEGIN_RCPP is meant to be overridden.). 

Also, not every rcpp exported function needs the Gil (or can even safely acquire it).

Defining our own C++ attribute and export wrapper would be a good approach too, but I opted for a scope object instead since it seemed simpler.

Happy to revise if you think it's worthwhile.

eddelbuettel:
I think you can consider that.  We (as in 'Team Rcpp') did variants on it to do / do not include RNG setters and other things.  You should then locally consistent function interfaces in the compilation unit (as `compileAttributes` will update all of them, using the macro everywhere).

t-kalinowski:
Do you know of an example of this type of approach I could use for reference?

eddelbuettel:
Not from the top of my head from _another_ package but armed with a strong coffee or tea you probably figure how the `BEGIN` and `END` macros play along for general 'resource' (in the sense of eg the RNG) handling.  Doing it quasi-mutex for the gil may work.

https://github.com/RcppCore/Rcpp/blob/master/inst/include/Rcpp/macros/macros.h

Now of course I don;t immediatly see the RNG handling in there so I guess that is handed off to attributes...

t-kalinowski:
So, trying something naive like redefining BEGIN_RCPP doesn't seem to work.

Defining this in `reticulate_types.h`:
```C
#define BEGIN_RCPP \
  BEGIN_RCPP       \
  GILScope gilscope;
```
gives:
```
  In file included from RcppExports.cpp:4:
   ./reticulate_types.h:91:9: warning: 'BEGIN_RCPP' macro redefined [-Wmacro-redefined]
   #define BEGIN_RCPP \
           ^
   /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/library/Rcpp/include/Rcpp/macros/macros.h:49:9: note: previous definition is here
   #define BEGIN_RCPP                                                                               \
           ^
   RcppExports.cpp:17:1: error: unknown type name 'BEGIN_RCPP'
   BEGIN_RCPP
   ^
   ./reticulate_types.h:92:3: note: expanded from macro 'BEGIN_RCPP'
     BEGIN_RCPP       \
     ^
   RcppExports.cpp:17:1: error: expected ';' at end of declaration
   ./reticulate_types.h:93:12: note: expanded from macro 'BEGIN_RCPP'
     GILScope gilscope;

<...truncated...>
```

Digging into Rcpp it looks like RNG scope code is emitted [here](https://github.com/RcppCore/Rcpp/blob/101ad174199330032f82518a40dc14184a503826/src/attributes.cpp#L2890), which doesn't seem to expose any entry points for adding additional resources like GILScope.

The next step I see is to mask/wrap Rcpp::export(), but it feels like that's going down the wrong path.

Any suggestions are welcome.

eddelbuettel:
For redefining a usual two-step is 

```
#ifdef foo 
#undef foo
#define foo
```

to avoid the redefinition warning.  Otherwise you seem to be on the right track but remember that these are macros and you need to get your special characters aligned.  I find that 'practising' in a one-off C(++) program (source) helps shake the cobwebs out.

t-kalinowski:
Thanks @eddelbuettel,

The warning about a redefined macro isn't the main issue, doing the `#undef` and `#define` dance still leaves me with a failed build:

```
#define RETICULATE_GIL_BEGIN_RCPP   \
  BEGIN_RCPP;         \
  GILScope gilscope ;

#undef BEGIN_RCPP
#define BEGIN_RCPP RETICULATE_GIL_BEGIN_RCPP
#undef RETICULATE_GIL_BEGIN_RCPP
```

gives:
```
─  installing *source* package ‘reticulate’ ...
   ** using staged installation
   ** libs
   clang++ -arch arm64 -std=gnu++14 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/library/Rcpp/include' -I/opt/R/arm64/include   -fPIC  -falign-functions=64 -Wall -g -O2  -Wall -pedantic -fdiagnostics-color=always -c RcppExports.cpp -o RcppExports.o
   RcppExports.cpp:17:1: error: unknown type name 'RETICULATE_GIL_BEGIN_RCPP'
   BEGIN_RCPP
   ^
   ./reticulate_types.h:98:20: note: expanded from macro 'BEGIN_RCPP'
   #define BEGIN_RCPP RETICULATE_GIL_BEGIN_RCPP
                      ^
   RcppExports.cpp:18:11: error: definition or redeclaration of 'RObject' not allowed inside a function
       Rcpp::RObject rcpp_result_gen;
       ~~~~~~^
   RcppExports.cpp:18:18: error: expected ';' at end of declaration
       Rcpp::RObject rcpp_result_gen;
                    ^
                    ;
   RcppExports.cpp:21:5: error: use of undeclared identifier 'rcpp_result_gen'
       rcpp_result_gen = Rcpp::wrap(write_stdout(text));
       ^
   RcppExports.cpp:22:12: error: use of undeclared identifier 'rcpp_result_gen'
       return rcpp_result_gen;
              ^
   RcppExports.cpp:23:1: error: expected unqualified-id
   END_RCPP
   ^
   /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/library/Rcpp/include/Rcpp/macros/macros.h:99:18: note: expanded from macro 'END_RCPP'
   #define END_RCPP VOID_END_RCPP return R_NilValue;
                    ^
   /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/library/Rcpp/include/Rcpp/macros/macros.h:62:5: note: expanded from macro 'VOID_END_RCPP'
       catch( Rcpp::internal::InterruptedException &__ex__) {                       
```

I'll try to work out the kinks in a skeleton package.

eddelbuettel:
Sounds good. Work on a local skeleton to work out kinks is what I'd do too.  Also, and @kevinushey may correct me here, I don't think we've used it this way _from another package_ but conceivably we could open a well-designed door this way.  (For example `RcppArmadillo` has used something similar for a decade to "inject" R's RNG into Armadillo's header-only library.)

t-kalinowski:
I think that what I was trying to do earlier is not possible. I wanted to redefine the `BEGIN_RCPP` macro by appending to the existing definition, but after some reading and reasoning I don't think the preprocessor lets you do that (though if you know of some delicate dance you can do to force a macro to be expanded inside a `#define` block on the first pass, please let me know).

Moving on, copy-pasting the existing `BEGIN_RCPP` definition and manually editing it gets past the preprocessor hurdle. This is what I have:

```C
#undef BEGIN_RCPP
#define BEGIN_RCPP                           \
  int rcpp_output_type = 0;                  \
  int nprot = 0;                             \
  (void)rcpp_output_type;                    \
  SEXP rcpp_output_condition = R_NilValue;   \
  (void)rcpp_output_condition;               \
  static SEXP stop_sym = Rf_install("stop"); \
  try {                                      \
    GILScope gilscope;

```

This compiles, but produces a segfault when loading. Now there needs to be some mechanism to opt-out of acquiring the GIL for the exported functions that can't safely acquire it yet (e.g., because Python is not initialized yet). 

Also, ideally it'd be nice if the `GILScope scope;` line was shown in the RcppExports.cpp adjacent to the RNG scope object, and not hidden in the `BEGIN_RCPP` macro, but `compileAttributes()` doesn't seem to have a nice built-in door for that.

t-kalinowski:
I've pushed an implementation where `BEGIN_RCPP` is redefined to acquire the GIL in all exported functions. I've added a guard in the scope object to only acquire the GIL if python is already initialized (tracked by a new global bool).

All tests pass with this new approach, and `rpy2` can also successfully call into reticulate, just like in the prior commit.

Overall, I'm not sure if this is an improvement over the commit prior. 

It's nice to not have to sprinkle the "`GILScope scope;`" line, but it:
 -  still doesn't leave a nice way for an exported function to opt out of acquiring the GIL. 
 -  feels brittle because we've hard-coded `BEGIN_RCPP`, meaning that a future update to Rcpp that changes this definition would break us.

eddelbuettel:
Six of one, half-a-dozen of another.  Sprinkling `GILscope` by hands risks forgetting it too.  These macros have probably not changed in a dozen years (and `git blame` could tell you more) so it's not clear this is a real risk.  Your call -- as a good software engineer you probably have to go to our shed now and invent a whole new preprocessor :wink:

kevinushey:
> still doesn't leave a nice way for an exported function to opt out of acquiring the GIL.

That's true. Rcpp has some built-in facilities for disabling some of the auto-generated code, e.g. for RNGScope one might use:

```
// [[Rcpp::export(rng = false)]]
double myFunction(double input) {
    // ...code that never uses the
    // R random number generation...
}
```

But nothing user-extensible. You could still hack something like:

```
GILScope scope(__func__);
```

and then use the function name to decide if `GILScope` should actually do work. Pretty hacky though.

> feels brittle because we've hard-coded BEGIN_RCPP, meaning that a future update to Rcpp that changes this definition would break us.

FWIW I'm proposing we do something like:

```
#include <Rcpp.h>

#ifdef BEGIN_RCPP
# define RETICULATE_BEGIN_CPP_PREAMBLE BEGIN_RCPP
# undef BEGIN_RCPP
#else
# define RETICULATE_BEGIN_CPP_PREAMBLE
#endif

#define BEGIN_RCPP \
  RETICULATE_BEGIN_CPP_PREAMBLE \
  GILScope scope(__func__);
```

Which doesn't seem totally insane to me.

kevinushey:
All that said, it's up to you to make the final call on what you feel will be easier to maintain / handle down the line. On the balance, I prefer the BEGIN_RCPP approach. One point in that approach's favor is that managing the GIL by default is probably the right call, and opting out could be done only on a case-by-case basis.

t-kalinowski:
Thank you for the very helpful feedback!



> ```
> define RETICULATE_BEGIN_CPP_PREAMBLE BEGIN_RCPP
> ```

I don't think the preprocessor lets you do that. When `RETICULATE_BEGIN_CPP_PREAMBLE` gets expanded to `BEGIN_RCPP` on the first pass, and then by the 2nd pass `BEGIN_RCPP` doesn't expand to the original preamble anymore. 

I've been experimenting using `gcc -E` in https://godbolt.org to troubleshoot this.


I think I'll merge this PR as-is shortly (unless there are other suggestions).

FWIW, if there are any plans to expand RCPP in this direction, the code I would *want* to write is something like this:

Define an R callback that can live somewhere like `src/rcpp-exports-callbacks.R`. The call back receives a character vector of the autogenerated function, the function, name signature and types, and *any other arguments* that were passed to `//Rcpp::export(...)`

So that I could write
```C
// [[Rcpp::export(gil = FALSE)]]
double myFunction(double input) {
    // ...code that never uses the Python GIL
}
```

Then have a callback
```R
post_process_Rcpp_export <- function(name, sig, body, ..., gil = TRUE) {
  if(gil) 
    body <- c("GILScope gilscope;", body)
  body
}
```

eddelbuettel:
We could support a hook similar to RNGScope.  A few months ago a patch added support for default arguments the signature generartor passes through to R.  But someone (you ?) has to sit down and write the patch.

kevinushey:
> I don't think the preprocessor lets you do that. RETICULATE_BEGIN_CPP_PREAMBLE gets expanded to BEGIN_RCPP on the first pass, and then by the 2nd pass BEGIN_RCPP doesn't expand to the original preamble anymore.

There should be a way to make this happen; e.g. use an intermediate function to pass along the macro (and force its expansion):

```
#define BEGIN_RCPP_IMPL(PREAMBLE) \
  PREAMBLE \
  GILScope scope(__func__);

#define BEGIN_RCPP BEGIN_RCPP_IMPL(RETICULATE_BEGIN_CPP_PREAMBLE);
```

kevinushey:
In particular, you cannot re-define a variable in terms of its previous definition; e.g. this does not work:

```
#define VAR a
#define VAR_OLD VAR
#define VAR_NEW VAR_OLD b
#undef VAR
#define VAR VAR_NEW
```

I had thought `VAR_OLD` would have been a "duplicate" definition of `VAR`, but instead they're more like "pointers" to the same content and `#undef VAR` forces both to lose their definitions.

simone-romiti:
Hi all, I'm here because I got the following message when using the `rmarkdown::render` function:
```
The R package "reticulate" only fixed recently
    an issue that caused a segfault when used with rpy2:
    https://github.com/rstudio/reticulate/pull/1188
    Make sure that you use a version of that package that includes
    the fix.
```
I have unistalled `reticulate` and installed again in an `R` session as follows:
```
library(devtools)
install_github(repo="https://github.com/rstudio/reticulate/")
```
Unfortunately I still get the segmentation fault. What should I do? Thank you in advance.

t-kalinowski:
Hi @simone-romiti Can you please open a new issue with a reprex?

PR #1191, opened 2022-04-19
only update 'envname' is python is associated with env

kevinushey:


kevinushey:
In this case, it seems like `reticulate` is able to automatically find and use its own Miniconda environment here:

```
> library(reticulate)
> library(tensorflow)
> reticulate:::python_info(py_exe())
$python
[1] "/opt/homebrew/bin/python3"

$type
[1] "system"

$root
[1] "/opt/homebrew/bin"

> install_tensorflow()
Installing: tensorflow-deps
+ '/Users/kevin/Library/r-miniconda-arm64/bin/conda' 'install' '--yes' '--name' 'r-reticulate' '-c' 'apple' '-c' 'conda-forge' 'tensorflow-deps'
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... done
< ... >
```

t-kalinowski (approved):
LGTM! 

While looking at this I wondered if it might make sense to explicitly throw an error if the user attempts to (perhaps unintentionally) install into a system python.


However, I decided against the additional guardrail since:
a) for most users, attempting to install into a system installation will throw an error about insufficient permissions.
b) configuring a minimal container with reticulate commands is a valid use-case (hence we should be able to install into a system installation).

t-kalinowski:
@Herudaio just published a new release to CRAN.

PR #1230, opened 2022-06-23
Fix documentation error for the `packages` argument

andresrcs:
Documentation for the `packages` argument shows indications to set the Python version instead of a package version.


t-kalinowski:
Thank you!

PR #1250, opened 2022-08-03
bugfix/knitr engine with html output

### LINKED ISSUE #1249: Error when testing the type for a pandas dataframe (Error in py_call_impl: TypeError: to_html() missing 1 required positional argument: 'self')
Have been using reticulate for several years in a larger shiny / knitr app and it has worked well. Hadn't used it with students for several months and am now getting the error below. Getting the type of basic python objects still works fine. However, when trying to get check the type of a pandas data frame there is a problem. 

Since I haven't changed the application and, with this exception, it still all works fine I can assume there is an issue with reticulate, knitr, or some other dependency. I'm planning to revert different package versions until it starts working again but I was hoping you might have some inspiration to share on which package this error might be most likely to originate in. Any suggestions would be much appreciated

image

t-kalinowski:
closes #1249


kevinushey:
Would it make sense to have something like py_has_method or py_has_callable for cases like this?

t-kalinowski:
I agree it's a good idea to have a `py_has_method()` function that explicitly checks for class methods with a bound `__self__`, and we could do that a hair more efficiently from C++.

Have you seen the error on the failed GHA tests before? I'm can't repro locally and I'm scratching my head about what could be different between the GHA runner and my machine (versions of OS, R, py, and packages are pretty much all the same). 

```
── Error (test-python-knitr-engine.R:18:3): An R Markdown document can be rendered using reticulate ──
[215](https://github.com/rstudio/reticulate/runs/7655765975?check_suite_focus=true#step:9:217)
<Rcpp::exception/C++Error/error/condition>
[216](https://github.com/rstudio/reticulate/runs/7655765975?check_suite_focus=true#step:9:218)
Error in `py_call_impl(callable, dots$args, dots$keywords)`: ValueError: 
[217](https://github.com/rstudio/reticulate/runs/7655765975?check_suite_focus=true#step:9:219)
Invalid named renderer(s) received: ['']
```
https://github.com/rstudio/reticulate/runs/7655765975?check_suite_focus=true

t-kalinowski:
The GHA failing test is specific to plotly in (I'm guessing) a headless Mac (https://github.com/plotly/plotly.py/blob/49dfcb1868e87627f032731ef2ac48e37891baf4/packages/python/plotly/plotly/io/_renderers.py#L221)

The value of `renderers_string` on my local machine is `'browser'` there, while it's an empty string in GHA.

t-kalinowski:
@kevinushey when you worked on the python plotly handling in the knitr engine, do you recall what the 'renderer' was supposed to be? https://plotly.com/python/renderers/. ('plotly_mimetype' perhaps?)

I ssh'd into a github runner and was able to play around a bit, I'm pretty sure we're not missing any system dependencies and this is something we have to fix in reticulate or plotly by configuring the default renderer on `plotly.io` initialization or by changing how we patch the `$show()` method.

kevinushey:
> @kevinushey when you worked on the python plotly handling in the knitr engine, do you recall what the 'renderer' was supposed to be? https://plotly.com/python/renderers/. ('plotly_mimetype' perhaps?)

Sorry, I don't recall :-/

PR #1253, opened 2022-08-04
use brew to install some python build deps if brew is available

t-kalinowski:
@kevinushey I'm admittedly not a very heavy user of brew. Do you see any pitfalls with this approach? 

(pyenv goes specifically looking for brew untapped build deps, and if this fails for any reason, then everything should still work, albeit slower since openssl@1.1 needs to be built from source.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,14 +1,17 @@
 # reticulate (development version)
 
+- Fixed an issue where reticulate would fail to bind to a conda environment on macOS or linux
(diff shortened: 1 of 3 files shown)


kevinushey:
This seems reasonable to me, but we might want an option or something similar just in case users want to opt out of this behavior?

t-kalinowski:
We can add a check for a user set `options()` when there is a feature request for that, but for now I'd rather hold off on expanding this API footprint until there is a user request.

Issue #583, opened 2019-08-22
Some errors with knitr Python engine stop render despite `error=TRUE` chunk option 


etabeta78:
While preparing some teaching material I have noticed that some Python chunks containing coding errors are not handled as I was expecting.

Namely, while chunks like this

  ```
    x=None
    x+1
  ```

print their error without stopping execution with {python error=TRUE}, other chunks e.g. with indentation errors

```
x = 0
y = 10
for ii in [1,2,3,4,5,6,7]:
    x += 1
     y -= 1
x, y
```

stop execution despite setting {python error=TRUE}. This happens with latest CRAN versions of Rmarkdown & reticulate.

Could this be handled in some way so that I can include different kinds of errors in rmarkdown:render?

p.s. might this be somehow related to issue #343 ?

kevinushey:
This is a bit challenging as we rely on the `ast` module to parse the Python code in a chunk, and so it expects each chunk to be a standalone parsable piece of code:

https://github.com/rstudio/reticulate/blob/5e0df26e404130bb689a89cabbc8c05046e25da9/R/knitr-engine.R#L84-L92

There's likely a workable alternate solution but I'm not sure what the best way to handle this is. Maybe something built on top of [InteractiveConsole](https://docs.python.org/3/library/code.html#code.InteractiveConsole)?

yihui:
I guess this should be simple enough to implement. Basically, when the chunk option `error = TRUE` and there is a parsing error, this engine can return something like this (similar to when R code chunks have syntax errors https://github.com/r-lib/evaluate/blob/6b307fa8dea4ac319a55aaa30a3699861a477b4d/R/eval.r#L44-L50):

```r
outputs <- list(
  structure(list(src = code), class = 'source'),
  parsed
)
```

kevinushey:
Thanks @yihui! You're right, I'll make a PR.

PR #1264, opened 2022-08-23
don't halt on parse error if error=TRUE

kevinushey:
Closes https://github.com/rstudio/reticulate/issues/583.


t-kalinowski (approved):
LGTM! Thanks @kevinushey!

PR #1270, opened 2022-08-30
change `py_id()` to return a string.

### LINKED ISSUE #1216: `py_id()` overflows in R-4.2 on Windows
```
library(reticulate)
py_id(py_eval("object()"))
[1] -1
```

Maybe it makes sense to change the return type of `py_id()` to be a string, rather than some flavor of an R numeric.

(originally encountered with failing keras tests, only on windows w/ R 4.2. Workaround here: https://github.com/rstudio/keras/commit/254bf6946c9a8e03e665a82574d6ea5d470bd693)

t-kalinowski:
closes #1216


src/python.cpp
```diff
@@ -3086,3 +3086,25 @@ SEXP py_has_method(PyObjectRef object, const std::string& name) {
...
+//' @return Unique identifer (as string) or `NULL`
+//'
+//' @export
+// [[Rcpp::export]]
+SEXP py_id(PyObjectRef object) {
+  if (py_is_null_xptr(object))
+    return R_NilValue;
+
+  auto id((uintptr_t) object.get());
+  return CharacterVector({std::to_string(id)});
```
kevinushey, line 3109: Minor nit, but `std::to_string` requires C++11 -- do we need to declare that? Or, we could use a `std::stringstream`?

t-kalinowski: Not minor at all, thanks for pointing this out! I did a quick search before writing this and noticed [`to_string()`](https://github.com/rstudio/reticulate/blob/d096cb782f6d8d1ceb8cfbf1b6c28b6cda2b3984/src/python.cpp#L95-L102), and at a cursory glance assumed it was fine to use... but on closer look maybe not. Perhaps `sprintf()`?

kevinushey: Thus far we've supported the now-ancient C++98 standard, but now that R itself defaults to C++14 (IIRC) we could probably start using C++11 facilities in reticulate.

t-kalinowski: I'll aim to keep the current compatibility this release; we can bump C++ std version the same time we drop python 2 support.

PR #1289, opened 2022-10-18
resolve module names in import hook

kevinushey:
Fixes an issue where module names (especially in recursive imports) might not be properly resolved.


t-kalinowski:
LGTM! Thank you for the comments :)

Do you think it's worth adding a unit test for resolving a module name in a recursive import?

t-kalinowski (approved):

kevinushey:
> Do you think it's worth adding a unit test for resolving a module name in a recursive import?

Good idea! I'll add something.

PR #1300, opened 2022-12-05
More explicitly remove problematic PYTHONPATH entries

CreRecombinase:
sys.path[0] is often an empty string, but that isn't always the case. This fixes a problem I had using a non-standard python interpreter ([spack python](https://spack.readthedocs.io/en/latest/developer_guide.html#cmd-spack-python))


t-kalinowski:
Thank you!

PR #1309, opened 2022-12-28
add `miniconda_path` to `use_miniconda`, fix #1308

### LINKED ISSUE #1308: `use_miniconda` doesn't work if miniconda wasn't installed in the default location
The documentation of `use_miniconda` states that:

https://github.com/rstudio/reticulate/blob/a1d7f7f573f652212bc2c72c39317340e6d8b511/man/use_python.Rd#L34-L36

But it won't work

```r
reticulate::install_miniconda("/tmp/Rtmp6Z1dcm/gfx", update = TRUE, force = TRUE)
### the absolute prefix path 
reticulate::use_miniconda("/tmp/Rtmp6Z1dcm/gfx/envs/r-reticulate")
### the absolute path to the python binary
reticulate::use_miniconda("/tmp/Rtmp6Z1dcm/gfx/envs/r-reticulate/bin/python")
```

The reason for it is that only the default path is checked

https://github.com/rstudio/reticulate/blob/a1d7f7f573f652212bc2c72c39317340e6d8b511/R/use_python.R#L238

chainsawriot:
This is a proposed fix to #1308 . The problem is that it introduces a new param `miniconda_path` and default to `miniconda_path()`. For most users who use the default location (or use the envvar `"RETICULATE_MINICONDA_PATH"`), it doesn't matter.

For some cases, it can be useful. (for example, if I want to have another envvar that store the path of another miniconda installation and I don't want to mask `"RETICULATE_MINICONDA_PATH"`).

### DIFF
--- a/R/use_python.R
+++ b/R/use_python.R
@@ -58,6 +58,9 @@
 #'   The path to a `conda` executable. By default, `reticulate` will check the
 #'   `PATH`, as well as other standard locations for Anaconda installations.
 #'
+#' @param miniconda_path
+#'   The path to miniconda. By default, it is [miniconda_path()].
+#' 
 #' @param required
 #'   Is the requested copy of Python required? If `TRUE`, an error will be
 #'   emitted if the requested copy of Python does not exist. Otherwise, the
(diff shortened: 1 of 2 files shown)


t-kalinowski:
Thank you for putting together the PR!

I think the functionality you're proposing to add to `use_miniconda()` is 
already covered by `use_condaenv()`. I updated the docs to make that clearer.

Issue #1328, opened 2023-01-25
`source_python()` unexpected input "errors"


chainsawriot:
I encountered this problem with `source_python` in several Ubuntu platforms (including the ones from Github Actions). It raises errors (enough to stop `testthat`) but it actually won't stop `source_python`. In the following example, the function `one` is actually available.

Digging into it with the debugger shows that the "errors" are raised in this line

https://github.com/rstudio/reticulate/blob/3bbc5b3a80088a812315f980c9f640b7eda226f7/R/source.R#L42

when `name` is "r" and `main_dict[[name]]` is unpacking.

I am creating another reprex with Github Actions.

``` r
test_py <- "
import os
def one():
\treturn(1)
"

temppy <- tempfile(fileext = ".py")
writeLines(test_py, temppy)
reticulate::source_python(temppy)
#> Error in parse(text = as_r_value(code)): <text>:1:2: unexpected input
#> 1: __
#>      ^
#> Error in parse(text = as_r_value(code)): <text>:1:2: unexpected input
#> 1: __
#>      ^
"one" %in% ls()
#> [1] TRUE
sessionInfo()
#> R version 4.2.2 Patched (2022-11-10 r83330)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Ubuntu 22.04.1 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.10.0
#> LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.10.0
#> 
#> locale:
#>  [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
#>  [3] LC_TIME=de_DE.UTF-8        LC_COLLATE=en_US.UTF-8    
#>  [5] LC_MONETARY=de_DE.UTF-8    LC_MESSAGES=en_US.UTF-8   
#>  [7] LC_PAPER=de_DE.UTF-8       LC_NAME=C                 
#>  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
#> [11] LC_MEASUREMENT=de_DE.UTF-8 LC_IDENTIFICATION=C       
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
[...]

PR #1329, opened 2023-01-27
invoke R functions from Python with `tryCatch()`

t-kalinowski:
Rcpp release 1.0.10 turned on use of `R_UnwindProtect()` by default. Previous versions of Rcpp would eval R code with a `tryCatch()`, silently catching errors instead of propagating them back up to the user session. Reticulate was implicitly depending on this behavior in various locations.

With this patch, reticulate explicitly places a `tryCatch()` on the stack when Python is calling back into R, and then reticulate will raise a python Exception if an R error was caught.

closes https://github.com/rstudio/reticulate/issues/1328

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,13 @@
 # reticulate (development version)
 
+- Fixed issue where `source_python()` (and likely many other entrypoints)
+  would error if reticulate was built with Rcpp 1.0.10. Exception and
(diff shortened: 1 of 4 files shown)


Enchufa2:
Great assessment, nice solution and cleaner code! Apologies for the trouble, but in the end I think that this led to an improvement in both packages, so a win-win. :)

PR #1343, opened 2023-02-27
Exception handling improvements

t-kalinowski:
This PR enables reticulate to capture and preserve an R error when it is signaled from an R function called from Python.

A secondary goal of this PR is to enable raising of arbitrary Python Exceptions (not just `RuntimeError()`) from R code.

To accomplish this:

-   Introduce a new custom C++ exception type, `PythonException()`.
    `PythonException()` can be initialized with an SEXP, which can be a R condition object or anything suitable for passing to `base::stop()`

-   Redefine the `RCPP_END` macro to include a catch for PythonException that passes through the R condition SEXP directly to `base::stop()`

-   Enable `base::stop()` to accept py_refs to Python Exception objects by:

    -   exporting `conditionCall()` and `conditionMessage()` methods for `python.builtin.BaseException`
    -   ensuring that py_refs to py objects that are instances of BaseException also inherit from S3 classes "error" and "condition"

-   For R functions invoked by the Python runtime,

    -   register calling handlers and restarts that capture the full R traceback before the R call stack is unwound.
    -   add an `r_to_py()` method for R errors, to ensure that a signaled R error is raised as an exception in the Python runtime.

-   Add some basic enhancements to `py_last_error()` and print methods, to present the additional captured information (not pretty yet).

-   To allow for preserving of R calls, introduce a `py_capsule()` function that creates a capsule from an SEXP, with no conversion.

-   Add `py_capsule()` as a fallback for `r_to_py()`.
    Objects that would previously error now return a py_capsule.

-   Since creation and deletion of capsules are now a hotter code path, replace usage of `R_PreserveObject()` and `R_ReleaseObject()` for capsules with `Rcpp_precious_preserve()` and `Rcpp_precious_remove()`


kevinushey (approved):
LGTM; nice work!

Issue #1226, opened 2022-06-14
`py_to_r()` and python subclasses.


t-kalinowski:
We ran into an interesting situation in Tensorflow.

Assigning a list of keras layers to `self` in a custom tensorflow model produces something that is meant to transparently behave like a python list to python users, but in actuality is a more complex object. This is what the R `class` attribute produced by reticulate looks like:
```
Browse[1]> class(self$dense_layers)
 [1] "tensorflow.python.training.tracking.data_structures.ListWrapper"           
 [2] "tensorflow.python.training.tracking.data_structures.List"                  
 [3] "tensorflow.python.training.tracking.data_structures.TrackableDataStructure"
 [4] "tensorflow.python.training.tracking.base.Trackable"                        
 [5] "collections.abc.MutableSequence"                                           
 [6] "collections.abc.Sequence"                                                  
 [7] "collections.abc.Reversible"                                                
 [8] "collections.abc.Collection"                                                
 [9] "collections.abc.Sized"                                                     
[10] "collections.abc.Iterable"                                                  
[11] "collections.abc.Container"                                                 
[12] "python.builtin.list"                                                       
[13] "python.builtin.object"   
```

This is the issue: If in an `__init__` method you build up a list of layers like this:
```r
    self$dense_layers <- lapply(hidden_units, function(x) layer_dense(units = x))
```
and then attempt to iterate over them in, e.g, `__call__`, it fails:
```
    for (layer in self$dense_layers) {
      x <- layer(x)
Error in for (l in self$dense_layers) layer(x) : 
  invalid for() loop sequence
```

Similarly, indexing into the "list" fails, because indexing is 0-based, not 1-based as an R user will expect:
```
    for (i in seq_along(self$dense_layers)) {
      x <- self$dense_layers[[i]](x)
Error in py_get_item_impl(x, key, FALSE) : 
  IndexError: list index out of range
```

In both cases, the issue is that `py_to_r.python.builtin.list` doesn't convert to an R list because it's not a "pure" python list. It seems to me like it should, especially since [`ListWrapper` explicitly subclasses the builtin `list`](https://github.com/tensorflow/tensorflow/blob/6db48475b45ba5a60e4f472076e962f1456017f4/tensorflow/python/trackable/data_structures.py#L459), but there are probably edge cases that I'm not thinking through. Making `py_to_r` succeed in situations like this may be a breaking change for some users.

[...]

PR #1352, opened 2023-03-13
Fix/convert subclassed unicode str

t-kalinowski:
This redefines the `PyUnicode_Check()` macro in reticulate to return true for subclassed Python `str()` objects, matching the behavior of `PyUnicode_Check()` defined in the official Python API (e.g., from `"Python.h"`)

Closes #1348 

This might also be a model for how to resolve https://github.com/rstudio/reticulate/issues/1226, using the other fields in `PyTypeObject`, like [`Py_TPFLAGS_LIST_SUBCLASS`](https://docs.python.org/3/c-api/typeobj.html#Py_TPFLAGS_LIST_SUBCLASS)

Issue #1348, opened 2023-03-07
fastai model$summary()


turgut090:
Hello. ```model$summary()``` from fastai throws:

```
model$summary()
Error: TypeError: expected bytes, PrettyString found 
```
But in python summary method works fine.
Is it an issue of reticulate?

Here is minimal python code:

```
from fastai.tabular.all import *

path = untar_data(URLs.ADULT_SAMPLE)
df = pd.read_csv(path/'adult.csv')
df_main,df_test = df.iloc[:10000].copy(),df.iloc[10000:].copy()


cat_names = ['workclass', 'education', 'marital-status', 'occupation', 'relationship', 'race']
cont_names = ['age', 'fnlwgt', 'education-num']
procs = [Categorify, FillMissing, Normalize]
splits = RandomSplitter()(range_of(df_main))


to = TabularPandas(df_main, procs, cat_names, cont_names, y_names="salary", splits=splits)

dls = to.dataloaders()

learn = tabular_learner(dls, layers=[200,100], metrics=accuracy)
learn.summary()
```
However, summary method fails from R

t-kalinowski:
Thanks for reporting. I just gave it a try but couldn't reproduce the error. Can you give a little more info about what OS/python-verion/package-versions you're using? 
![image](https://user-images.githubusercontent.com/8462255/223445093-71e4a7fa-3848-4b01-a644-efae6cfd5334.png)

turgut090:
Thanks for you reply. If you run the same code from R with reticulate. It fails and throws:
```
model$summary()
Error: TypeError: expected bytes, PrettyString found 
```

R code:

```

ft=reticulate::import('fastai.tabular.all')

# download
#URLs_ADULT_SAMPLE()

# read data
df = data.table::fread('/Users/turgutabd/.fastai/data/adult_sample/adult.csv')

dep_var = 'salary'
cat_names = c('workclass', 'education', 'marital-status', 'occupation', 'relationship', 'race')
cont_names = c('age', 'fnlwgt', 'education-num')

procs = reticulate::r_to_py(list(ft$FillMissing,ft$Categorify,ft$Normalize))

dls = ft$TabularPandas(df, procs,cat_names =  cat_names,cont_names =  cont_names,
                       y_names = dep_var, splits = reticulate::r_to_py(list(c(1L:32000L),c(32000L:32560L))))

to = dls$dataloaders()

model = ft$tabular_learner(to,layers=c(200L,100L), metrics=ft$accuracy)

[...]

PR #1354, opened 2023-03-14
update `Py*_Check()` macros to accept subtypes

t-kalinowski:
Also add `Py*_CheckExact()` macros,  with the previous implementation.

This change makes the type check macros match
the behavior of the same named macros in the Python API.

closes: #1226
related: #1352, #1348


kevinushey (approved):
Very nice, LGTM!

PR #1361, opened 2023-03-29
R wrapper functions with matching Py callable signature

t-kalinowski:
To check that this doesn't have a major impact on conversion times, I did some quick benchmarks by timing how long it takes to convert all callables in the tensorflow module, going down 3 levels of recursion (~22k callables)

```r
Sys.setenv("CUDA_VISIBLE_DEVICES" = "")

library(tensorflow)
library(reticulate)

py_is_callable <- import_builtins()$callable
stack <- fastmap::faststack(22500)

fetch_functions <- function(x, depth = 0, path = character()) {
  nms <- py_list_attributes(x)
  nms <- nms[!startsWith(nms, "_")]
  for(nm in nms) {
    if (depth > 3) return()
    obj <- x[[nm]]

    if (py_is_callable(obj)) {
      obj_ <- list(obj)
      names(obj_) <- paste0(c(path, nm), collapse = ".")
      stack$push(obj_)
    }

    if (inherits(obj, c("python.builtin.module", "python.builtin.type"))) {
      # recurse
      fetch_functions(obj, depth = depth + 1, c(path, nm))
    }
  }
  invisible()
}


invisible(as_tensor(1)) # initialize tensorflow
print(system.time(fetch_functions(tf, path = "tf")))



## with this py_formals branch:
# cold start terminal
#   user  system elapsed
# 11.441   0.031  11.371

# with dev main tensorflow/reticulate
#   user  system elapsed
# 10.131   0.029  10.160

# with cran tensorflow/reticulate
#  user  system elapsed
# 8.142   0.016   8.163
```

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -1263,13 +1263,14 @@ SEXP NewList(void) {
 }
 
 /* Add named element to the end of a stretchy list */
-void GrowList(SEXP args, SEXP tag, SEXP dflt) {
-  SEXP tmp;
-  tmp = Rf_cons(dflt, R_NilValue);
+void GrowList(SEXP args_list, SEXP tag, SEXP dflt) {
+  PROTECT(dflt);
+  SEXP tmp = PROTECT(Rf_cons(dflt, R_NilValue));
   SET_TAG(tmp, tag);
 
-  SETCDR(CAR(args), tmp); // set cdr on the last cons-cell
-  SETCAR(args, tmp);      // update pointer to last cons cell
+  SETCDR(CAR(args_list), tmp); // set cdr on the last cons-cell
+  SETCAR(args_list, tmp);      // update pointer to last cons cell
+  UNPROTECT(2);
 }
 
 // [[Rcpp::export]]
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -1251,110 +1251,139 @@ SEXP py_to_r(PyObject* x, bool convert) {
 
 }
 
+/* stretchy list, modified from R sources
+   CAR of the list points to the last cons-cell
+   CDR points to the first.
+*/
+
+SEXP NewList(void) {
+  SEXP s = Rf_cons(R_NilValue, R_NilValue);
+  SETCAR(s, s);
+  return s;
+}
+
+/* Add named element to the end of a stretchy list */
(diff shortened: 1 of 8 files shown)


src/python.cpp
```diff
@@ -1251,110 +1251,140 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
-      // Here we could convert a subset of python objects to R defaults.
-      // Plain values (numeric, character, NULL, ...) are stored as is,
-      // variables, calls, ... are stored as `symbol` or `language`.
-      formals << Named(as_utf8_r_string(param_name.get()), R_NilValue);
+    else
+    {
+      arg_default = py_to_r(param_default, true); // callable.convert());
     }
+
+    GrowList(r_args, Rf_install(PyUnicode_AsUTF8(name)), arg_default);
```
kevinushey, line 1381: Nit: arguments to `Rf_install()` should be in the native encoding, which may not be UTF-8. You might want to use `translateChar()` or some similar R API to confirm you're getting a string in the native encoding.

src/python.cpp
```diff
@@ -1251,110 +1251,140 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
+*/
+
+SEXP NewList(void) {
+  SEXP s = Rf_cons(R_NilValue, R_NilValue);
+  SETCAR(s, s);
+  return s;
+}
+
+/* Add named element to the end of a stretchy list */
+void GrowList(SEXP args, SEXP tag, SEXP dflt) {
```
kevinushey, line 1266: Do we need to worry about how these arguments are protected here?

kevinushey (commented):
Approach LGTM overall; nice work!

src/python.cpp
t-kalinowski: Thanks, I didn't know that about `Rf_install()`! I've updated the snippet to 
```cpp
const char *name_char = PyUnicode_AsUTF8(name);
SEXP name_sym = Rf_installTrChar(Rf_mkCharCE(name_char, CE_UTF8));
```

src/python.cpp
t-kalinowski: Good catch! I added a `PROTECT(dflt)` before the `Rf_cons()` call.

PR #1370, opened 2023-05-10
new default: don't simplify Python lists

### LINKED ISSUE #1023: Simplification of python lists
Hello. I'm working on improving the experience around subclassing python classes from R. Subclassing is a design pattern heavily encouraged in the python keras/tensorflow api, and I think it's important to enable similar usage on the R side.

One of the issues I'm encountering is due to automatic simplification of python lists.
User supplied R lists convert to python lists, but then they don't convert back to R lists.

### Where it comes up

In python it's common to do:
```python
class MyLayer(keras.layers.Layer):
  def __init__(self, **kwargs):
    super().__init__(**kwargs)
    
my_layer = MyLayer(input_shape = [32])
```

In R, I want users to be able to do the equivalent:

(for the moment, please ignore the R syntax for constructing and instantiating
classes and resolving `super`, I intend to open a separate discussion/PR for that soon)

```r
MyLayer <- r_to_py(R6Class("MyLayer",
  inherit = keras$layers$Layer,
  public = list(
    `__init__` = function(self, ...) {
      super$`__init__`(...)
    }
  )
))

my_layer <- MyLayer(input_shape = shape(32))
```

In the above R example, the issue I'm encountering is that arguments supplied to
`...` are converted between python and R multiple times, Concretely, they :

- start as R objects, 
- convert to python objects (for `MyLayer.__new__()`), 
- convert back to R objects (for `MyLayer.__init__()`), 
- convert back again to python objects (for `super().__init__()`). 

In the process, what the caller intended to be a list gets simplified to a scalar:
(R) `shape(32)` -> (py) `[32]` -> (R) `c(32)` -> (py) `32`, resulting in an exception from `keras.layers.Layer.__init__`.

Note, this doesn't just come up when passing R `...` from one python function to another. Because `self` is a python object, the potential for unintended simplification is present anytime `self` has an attribute set and then accessed. For example, another common pattern in keras custom layers is to defer building a layer's weights by providing a `build` method:

```r
MyLayer <- r_to_py(R6Class("MyLayer",
  inherit = keras$layers$Layer,
  public = list(
    `__init__` = function(self, weight_shape, ...) {
      super$`__init__`(...)
      self$weight_shape <- weight_shape
    },
    build = function(self, input_shape) {
      self$custom_weight <- super$add_weight(
        shape = self$weight_shape
      )
      super$build(input_shape)
    }
  )
))

my_layer <- MyLayer()
input <- layer_input(shape(10))
my_layer(input) 
```

This also throws an exception, because the value supplied to `weight_shape` is converted to a python list on assignment to `self$input_shape`, simplified to an R numeric when it's accessed in `build`, and then converted to a python scalar when passed on to `super$add_weight()`.

I can try to "fix-up" specific keywords I know about in advance like this in keras, but I wonder if there is perhaps a more sustainable solution that allows for lossless conversion for more objects crossing the R <--> py boundary multiple times.

### Some ideas for how to resolve this (in order of most-to-least disruptive)

1) Disable auto-simplification of python lists to R vectors; python lists always get converted to R lists. This is probably too disruptive a change at this point, so maybe it's off the table, but IMO this is the cleanest resolution. Just to go through the thought exercise, this is what the new conversion behavior could be: 
```
R          ->       py      ->      R            lossless
---------------------------------------------------------
1                    1              1               yes
c(1,2)            [1,2]         list(1,2)           no
list(1)             [1]         list(1)             yes
list(1,2)         [1,2]         list(1,2)           yes
array(c(1))    np.array([1])    array(c(1))         yes
array(c(1,2))  np.array([1,2])  array(c(1,2))       yes
```

In slack @dfalbel made an excellent suggestion: to assist with the transition we could provide a context switch for users with code that depends on the automatic simplification `with_old_conversion_behavior({})`.

2) Add a dummy python class: 
  ```python
  class Rlist(list): pass
  ```
  Objects that originate as R lists get converted to python's `Rlist`. The `Rlist` behaves exactly like a builtin python list, but on the way back to R we do the equivalent of `py_to_r.rpytools.rtypes.Rlist <- function(x) as.list(NextMethod())`.

3) Introduce a breaking change by changing `tensorflow::shape` to return a `tf$TensorShape()` object. This solves the roundtrip issue, but breaks any code that manipulates the shape where an R list is expected. We could help lessen the impact by providing `c`, `[`, `[[`, `[<-`, and `[[<-` S3 methods for `tf.TensorShape` objects.

4) Do the equivalent of (2), but confined to just the R lists returned by `tensorflow::shape`. This would probably mean making the return object from `shape` an S3 class with `c`, `[`, `[[`, `[<-`, and `[[<-`, and `r_to_py` methods, and then a companion `py_to_r` method.

5) Leave `tensorflow::shape()` as-is. Do the equivalent of (3) but in new function `tf_shape`. Update all docs/tutorials/unittests to swap `shape` for `tf_shape()`.

Objects round-tripping between python-land and R-land is going to be pretty unavoidable, and requiring the user to "fix-up" those objects every time with `as.list()` seems cumbersome, especially if it means objects have to be broken out of `...` to do so.

Here is a minimal example demonstrating simplification:
```r
py_reflect <- reticulate::py_run_string(
"def reflect(x):
  print('python fn received: ', x)
  return x
")$reflect
list()    |> py_reflect() |> py_reflect()   # ok     
list(1,2) |> py_reflect() |> py_reflect()   # not ideal, but not my main issue  
list(1)   |> py_reflect() |> py_reflect()   # my main issue
```
giving output:
```r
> list()    |> py_reflect() |> py_reflect()   # ok     
list()
python fn received:  []
python fn received:  []
> list(1,2) |> py_reflect() |> py_reflect()   # not ideal, but not my main issue  
[1] 1 2
python fn received:  [1.0, 2.0]
python fn received:  [1.0, 2.0]
> list(1)   |> py_reflect() |> py_reflect()   # my main issue
[1] 1
python fn received:  [1.0]
python fn received:  1.0
```

After writing this out, I'm on the fence between options (1) or (5), perhaps with a small preference for (1). I would be curious to hear your thoughts.
@kevinushey @dfalbel @skeydan

t-kalinowski:
closes #1023

### DIFF
--- a/R/python.R
+++ b/R/python.R
@@ -1724,7 +1724,7 @@ py_last_error <- function(exception) {
   if(is.null(etb))
     formatted_traceback <- NULL
   else
-    formatted_traceback <- traceback$format_tb(etb)
+    formatted_traceback <- as.character(traceback$format_tb(etb))
 
   out <- list(
     type = py_get_attr_impl(etype, "__name__", TRUE),
@@ -1865,3 +1865,5 @@ format_py_exception_traceback_with_clickable_filepaths <- function(etb) {
 
   cli::col_silver(paste("Run", py_last_error, "for details."))
 }
+
+
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -831,6 +831,12 @@ int r_scalar_type(PyObject* x) {
 // check whether the PyObject is a list of a single R scalar type
 int scalar_list_type(PyObject* x) {
 
+  // check R option, default to false
+  SEXP simplify_s = Rf_GetOption1(Rf_install("reticulate.simplify_lists"));
+  bool simplify = simplify_s == R_NilValue ? false : Rf_asLogical(simplify_s);
+  if(!simplify)
+    return NILSXP;
+
   Py_ssize_t len = PyList_Size(x);
   if (len == 0)
     return NILSXP;
--- a/tests/testthat/resources/venv-activate.R
+++ b/tests/testthat/resources/venv-activate.R
@@ -7,4 +7,4 @@ Sys.unsetenv("RETICULATE_PYTHON_ENV")
 
 reticulate::use_virtualenv(venv, required = TRUE)
 sys <- reticulate::import("sys")
-writeLines(sys$path)
+writeLines(as.character(sys$path))
--- a/tests/testthat/test-python-base-r-generics.R
+++ b/tests/testthat/test-python-base-r-generics.R
@@ -92,6 +92,7 @@ test_that("Ops group generics dispatch correctly when only one argument is a Pyt
 })
 
 
+if(getRversion() >= "4.3.0")
 test_that("matrixOps group generics dispatch", {
 
   r_obj1 <- array(1:9, c(3, 3))
--- a/tests/testthat/test-python-complex.R
+++ b/tests/testthat/test-python-complex.R
@@ -10,7 +10,11 @@ test_that("Complex scalars are converted correctly", {
 test_that("Complex vectors are converted correctly", {
   skip_if_no_python()
   z <- complex(real = stats::rnorm(100), imaginary = stats::rnorm(100))
-  expect_equal(z, test$reflect(z))
+  expect_equal(as.list(z), test$reflect(z))
+  withr::with_options(c(reticulate.simplify_lists = TRUE), {
+    expect_equal(z, test$reflect(z))
+  })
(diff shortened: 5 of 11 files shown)


t-kalinowski:
With the distance of time, I no longer think the benefit is worth the breaking change.

PR #1389, opened 2023-06-07
restore R 3.5 compatability 

### LINKED ISSUE #1388: reticulate-1.29 compilation error with R-3.5; missing Rf_installTrChar
reticulate-1.29 fails to compile from source with R-3.5.

Here is a `Dockerfile` to recreate the problem:

```dockerfile
FROM rstudio/r-base:3.5-focal

RUN export DEBIAN_FRONTEND=noninteractive \
    && apt-get update \
    && apt-get install -y libpng-dev \
    && rm -rf /var/lib/apt/lists/*

RUN R -e 'install.packages("reticulate", repos = c(CRAN="https://cran.rstudio.com/"))'
```

```bash
docker build --progress=plain .
```

Output from compilation:

```
#6 70.49 python.cpp: In function ‘SEXPREC* py_get_formals(PyObjectRef)’:
#6 70.49 python.cpp:1470:21: error: ‘Rf_installTrChar’ was not declared in this scope; did you mean ‘Rf_installNoTrChar’?
#6 70.49  1470 |     SEXP name_sym = Rf_installTrChar(Rf_mkCharCE(name_char, CE_UTF8));
#6 70.49       |                     ^~~~~~~~~~~~~~~~
#6 70.49       |                     Rf_installNoTrChar
#6 70.89 make: *** [/opt/R/3.5.3/lib/R/etc/Makeconf:172: python.o] Error 1
#6 70.89 ERROR: compilation failed for package ‘reticulate’
```

t-kalinowski:
- replace usage of `Rf_installTrChar()`
- update tests and DESCRIPTION

closes #1388


src/python.cpp
```diff
@@ -1467,7 +1467,7 @@ SEXP py_get_formals(PyObjectRef callable)
     const char *name_char = PyUnicode_AsUTF8(name);
     if (name_char == NULL) throw PythonException(py_fetch_error());
 
-    SEXP name_sym = Rf_installTrChar(Rf_mkCharCE(name_char, CE_UTF8));
+    SEXP name_sym = Rf_installChar(Rf_mkCharCE(name_char, CE_UTF8));
```
kevinushey, line 1470:
```suggestion
    SEXP name_sym = Rf_installChar(Rf_mkCharCE(name_char, CE_NATIVE));
```

I believe symbols need to be in the native encoding, not UTF-8, for older versions of R?

kevinushey, line 1470: I _think_ you need to explicitly translate the CHARSXP here from UTF-8 to the native encoding, at least for older versions of R?

kevinushey: Yeah, `installChar()` in older versions of R assumes that the CHARSXP is already in the native encoding, so I think you should translate first.

t-kalinowski: I think `Rf_installChar()` just calls `Rf_installTrChar()`, at least in R 3.5 and 4.3 ( 1st and 3rd link in https://github.com/rstudio/reticulate/issues/1388#issuecomment-1581245580)

t-kalinowski: i.e., `installChar()` and `installTrChar()` are functionally identical.

kevinushey: This is the definition I see in the R 3.4.4 sources:

```
/* This function is equivalent to install(CHAR(charSXP)), but faster.
   Like the equivalent code pattern, it discards the encoding information,
   hence in almost all cases installTrChar should be used, instead. */
SEXP installChar(SEXP charSXP)
{
    SEXP sym;
    int i, hashcode;

    if( !HASHASH(charSXP) ) {
	hashcode = R_Newhashpjw(CHAR(charSXP));
	SET_HASHVALUE(charSXP, hashcode);
	SET_HASHASH(charSXP, 1);
    } else {
	hashcode = HASHVALUE(charSXP);
    }
    i = hashcode % HSIZE;
    /* Check to see if the symbol is already present;  if it is, return it. */
    for (sym = R_SymbolTable[i]; sym != R_NilValue; sym = CDR(sym))
	if (strcmp(CHAR(charSXP), CHAR(PRINTNAME(CAR(sym)))) == 0) return (CAR(sym));
    /* Create a new symbol node and link it into the table. */
    int len = LENGTH(charSXP);
    if (len == 0)
	error(_("attempt to use zero-length variable name"));
    if (len > MAXIDSIZE)
	error(_("variable names are limited to %d bytes"), MAXIDSIZE);
    if (IS_ASCII(charSXP) || (IS_UTF8(charSXP) && utf8locale) ||
					(IS_LATIN1(charSXP) && latin1locale) )
	sym = mkSYMSXP(charSXP, R_UnboundValue);
    else {
	/* This branch is to match behaviour of install (which is older):
	   symbol C-string names are always interpreted as if
	   in the native locale, even when they are not in the native locale */
	PROTECT(charSXP);
	sym = mkSYMSXP(mkChar(CHAR(charSXP)), R_UnboundValue);
	SET_HASHVALUE(PRINTNAME(sym), hashcode);
	SET_HASHASH(PRINTNAME(sym), 1);
	UNPROTECT(1);
    }

    R_SymbolTable[i] = CONS(sym, R_SymbolTable[i]);
    return (sym);
}
```

If we want to support R 3.4.x, then I think we do need to translate first.

DESCRIPTION
```diff
@@ -34,7 +34,7 @@ BugReports: https://github.com/rstudio/reticulate/issues
 SystemRequirements: Python (>= 2.7.0)
 Encoding: UTF-8
 Depends:
-    R (>= 3.0)
+    R (>= 3.4)
```
kevinushey, line 37: Or we could just use >= 3.5?

t-kalinowski: I think this is the best option. Otherwise, we'll have to manually install older versions of rlang (and testthat!). 
Trying to run the current CI with R 3.4 I see this:
```
Error: 
  ! error in pak subprocess
  Caused by error: 
  ! Could not solve package dependencies:
  * deps::.: Can't install dependency rlang
  * rlang: Needs R >= 3.5.0
  * rcmdcheck: Can't install dependency webfakes
  * webfakes: Needs R >= 3.5.0
  * remotes: Can't install dependency webfakes
  * any::sessioninfo: dependency conflict
```

kevinushey: Oof. Let's just bump to 3.5 then :-)

Issue #1391, opened 2023-06-13
Matplotlib bar plots need `plt.show()`, line plots do not


matthew-brett:
I have the following file `test.Rmd`:

~~~rmarkdown
# A notebook file

```{python}
import matplotlib.pyplot as plt
```

```{python}
plt.bar([1, 2, 3], [2, 3, 1])
```

```{python}
plt.plot(range(10))
```

```{python}
plt.plot(range(0, -10, -1))
```
~~~

When I render this file with:

```r
rmarkdown::render("test.Rmd", output_format = "pdf_document")
```

I get the following file:
[test.pdf](https://github.com/rstudio/reticulate/files/11736943/test.pdf)

Oddly - the `plt.plot` commands each generate their own figure, without the need for `plt.show()`, as I was expecting.  But the `plt.bar()` command does not generate a figure without `plt.show()`, and instead gets written to an as-yet hidden figure, then appended to by the subsequent `plt.plot` command.

This difference in behavior between `plt.bar` (also `plt.hist`) and `plt.plot` was surprising to me - and doesn't correspond to any behavior I know if in Matplotlib itself.  Is it expected?  Is there a way to work around it?

I'm using:

```
$ R --version
R version 4.3.0 (2023-04-21) -- "Already Tomorrow"
Copyright (C) 2023 The R Foundation for Statistical Computing
Platform: x86_64-apple-darwin22.4.0 (64-bit)
```

with just-built `rmarkdown` and `reticulate`.

t-kalinowski:
Thanks for reporting! This should be fixed on main now. 
You can install the development version with:
```r
remotes::install_github("rstudio/reticulate")
```

matthew-brett:
Ah - sorry - I see that the same problem still applies to `plt.hist()`:

~~~rmarkdown
# A notebook file                                                                               
                                                                                                
```{python}                                                                                     
import numpy as np                                                                              
import matplotlib.pyplot as plt                                                                 
```                                                                                             
                                                                                                
```{python}                                                                                     
plt.hist(np.random.normal(size=1000))                                                           
```                                                                                             
                                                                                                
```{python}                                                                                     
plt.plot(range(10))                                                                             
```                                                                                             
                                                                                                
```{python}                                                                                     
plt.plot(range(0, -10, -1))                                                                     
``` 
~~~

I see this is because you are checking the return types from the plot commands - `plt.hist` returns an array and a BarContainer object.

[...]

PR #1394, opened 2023-06-20
MRG: First draft of jupyter_compat mode

matthew-brett:
This is a first attempt to replicate the default Jupyter behavior with
Reticulate.

That is:

* Always display plots
* Do not display results of code expressions unless at last code line.
* Display result of final expression unless suppressed with ;

See: gh-1391

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,21 +1,25 @@
 # reticulate (development version)
 
 - New optional feature: Reticulate now accepts a new option `jupyter_compat`
-  set to `FALSE` by default, that changes the default display behavior of
-  Reticulate chunks, to better match the behavior of Jupyter.  In the
-  Reticulate default, each standalone code expression in the code chunk that
-  does not end in a semi-colon, generates display of the expression output. For
-  Matplotlib plots, an explicit `plt.show()` generates the figure, but
-  otherwise, the chunk (by default) only generates a figure if the last line in
-  the chunk is an expression returning a recognized Matplotlib object, and
-  without a semi-colon.  With `jupyter_compat=TRUE`, a standalone expression
-  returning a Matplotlib object, anywhere in the chunk, will cause the plot to
-  be displayed automatically after the chunk.  No expression in the chunk will
+  set to `FALSE` by default, that changes the default expression output display
+  behavior of Reticulate chunks, to better match the behavior of Jupyter.  In
+  the Reticulate default, each standalone code expression in the code chunk
+  that does not end in a semi-colon, generates display of the expression
+  output. With the `jupyter_compat` option set, no expression in the chunk will
   generate output, except if there is a standalone expression as the last code
-  statement in the chunk, and that expression does not have a semicolon
-  — a semicolon suppresses the output, as it does for the default behavior.
-  See [PR](https://github.com/rstudio/reticulate/pull/1394) and [original
-  issue](https://github.com/rstudio/reticulate/issues/1391).
+  statement in the chunk, and that expression does not have a semicolon.
+  A semicolon always suppresses the expression output, for the default and
+  `jupyter_compat` case. See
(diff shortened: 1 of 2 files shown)


matthew-brett:
I've added a `NEWS` entry - look ok?

I had two worries:

* There are currently two ways of preventing code output - the `suppress` option, that changes the `compile_mode`, used now for the standard mode, and my post-hoc deletion of the return value for `jupyter_compat`.   Do we need both?
* I guess one problem will be, that we can only detect Matplotlib plotting when the code uses standalone expressions, rather than statements.  Is that right?   If so, it would be easy to miss with code like `arrs, bins = plt.hist(some_array)`.

t-kalinowski:
Thanks
- In the news entry, can you please include a link to this pr and the original issue?
- I agree, consolidating the code paths makes sense. I'd be OK always parsing in `"single"` mode, and then handling or discarding the output afterwards, in `eng_python_autoprint()` or just after it. Additionally, with this change, it might make sense to completely supplant the 'inspect-the-last-value-and-call-`plt.show()`-if-its-a-matplotlib-object' logic with the new logic based around `matplotlib_pending_show`. Seems like that's a cleaner solution over-all. 
- We might not even need a `jupyter_compat` option, and my preference is not to expand the API if it's not needed. If I'm understanding correctly:
	- the `repr` of a matplotlib plot is by default always suppressed by the ending
	- with this new approach, a `plt.show()` is always called at the end of a chunk if a matplotlib plot object was returned by any of the individual expressions.
	- The presence or absence of a trailing `;` on the last expression of a chunk returning a matplotlib object is then, effectively, a no-op---it makes no difference since the `plt.show()` will be called regardless, and the `repr()` is suppressed regardless.

matthew-brett:
I've added the links to the NEWS entry.

For the logic as currently implemented with `jupyter_compat`

* Only the final expression gets output displayed.
* Semi-colon will suppress that final output, if present, but otherwise, yes, it's a no-op.
* I defered to `eng_python_autoprint` for the repr for the Matplotlib-returning expressions, so yes, these are currently suppressed regardless.   I guess for full Jupyter compatibility, we should not suppress reprs for the final expression (without a semi-colon).

Did you look at the Matplotlib figure manager?   I have a vague memory that is where you can look for any new but not-yet-displayed figures.

t-kalinowski:
Thank you!

> Did you look at the Matplotlib figure manager? I have a vague memory that is where you can look for any new but not-yet-displayed figures.

I have not. I remember having a brief, unsuccessful, look for something like a "pending undisplayed matplotlib plots queue". I'm open to other suggestions for how to to better detect if there is a pending plot, though I think this new approach embodied in this PR seems good.

t-kalinowski:
>  I guess for full Jupyter compatibility, we should not suppress reprs for the final expression (without a semi-colon).

Is this necessary? The `repr` seems to me output that nobody will want in their polished and rendered report, perhaps excepting someone writing a matplotlib tutorial.

I'm inclined to keep the current behavior.

matthew-brett:
> > 
> 
> Is this necessary? The `repr` seems to me output that nobody will want in their polished and rendered report, perhaps excepting someone writing a matplotlib tutorial.

Fair enough - it's certainly niche - although I can imagine a student being surprised they get the output in the Jupyter notebook, and don't see it in the report.

matthew-brett:
I asked for advice on detecting new plots [over on the Matplotlib discourse](https://discourse.matplotlib.org/t/best-way-to-detect-new-figure-in-jupyter-code-cell/23889) ....

matthew-brett:
Also tried asking about plot detection over at [Jupyter Discourse](https://discourse.jupyter.org/t/how-does-jupyter-detect-matplotlib-plots-in-a-cell/20047).

t-kalinowski (approved):
LGTM! Thank you!

PR #1401, opened 2023-06-24
MRG: use Matplotlib to detect new Matplotlib figures

matthew-brett:
This uses what [appears to be the standard
idiom](https://github.com/matplotlib/matplotlib/issues/26172#issuecomment-1604216078)
to detect pending Matplotlib plots.

With this PR, we _do_ detect plots generated by code such as:

`res = plt.plot(range(10))`

whereas previously we would have missed this as a plot generator, because
it was not a standalone expression.


matthew-brett:
I see that tests are failing, and detecting output differences, but I'm having some trouble replicating locally.  Can I ask for hints as where to look to debug the test failures?

t-kalinowski:
Nice, this looks like a much better approach!

Do you know if we have to cache the result of `get_fignums()` between chunks? It might be good to add a test for this edge case, to make sure we don't call `plt.show()` too many times.

The relevant test file is: https://github.com/rstudio/reticulate/blob/main/tests/testthat/test-python-knitr-engine.R

To run locally and interactively, from the package directory you can evaluate the R expressions:
```r
devtools::load_all()
testthat::test_file("tests/testthat/test-python-knitr-engine.R")
```

The test snapshots and resources are available under the parent folder, named "testthat/resources" and "testthat/_snaps", respectively.

Thank you!

matthew-brett:
OK - thanks - with those hints, I tracked down the failures.  It turns out that, at least from document to document, there can be stray empty Matplotlib figures, resulting in multiple empty figures being generated, and then inserted into the output documents.

For now, I've just checked if pending figures are empty, and closed those found to be empty.

I did think about what to do if there were figures pending from before the chunk starts - but it seemed to me that a) I wasn't sure how those would come about, given the current mode, of always displaying figures if they arose in a chunk, and b) if they did come about - it seemed to me that the Jupyter way would be to display them - but I don't think there is a way to have panding figures from a Jupyter chunk.

matthew-brett:
Am I right in thinking that the current test failures are unrelated?

t-kalinowski:
I made some commits to the branch, but they diverged with the most recent force-push. 
I pushed them to a separate branch here: https://github.com/rstudio/reticulate/tree/mpl-figure-check

Re the `import("matplotlib")` question, that call should always succeed, since it's only called from a hook that is registered to run only after the first time the `matplotlib` module is imported successfully.

t-kalinowski:
The failing tests are unrelated, they're due to a new scipy release. I'll fix on a different branch.

matthew-brett:
No problem, thanks for guiding it through….

PR #1405, opened 2023-07-05
Add `prefetch` arg to `py_iterator()`

t-kalinowski:
`py_iterator()` is used to create a python iterator from an R function. This PR adds a `prefetch` argument to `py_iterator()`. 

TensorFlow/Keras can accept user defined iterators for producing data (e.g., batches of training data). TensorFlow will in some circumstances call `iterator.__next__()` from a Python background thread which has the GIL acquired, and with the main thread fully blocked in a C++ call until the `__next__()` call has completed. 

With an iterator which calls an R function, this results in either:
- A `stackOverflowError` R error (or segfault, or corrupted data), due to the R API being called from the non-main thread. 
- A deadlock, where the main thread is blocked waiting on the R function to run, and the R function call is waiting to run on the main thread.

One such location where this happens is `tf.data.Dataset.from_generator()`.
 
Keras version 2.13 refactored internals to consolidate code paths in `fit()`, `predict()`, etc.  to use `tf.data.Dataset`, with user-supplied iterators being converted to TF Datasets via `tf.data.Dataset.from_generator()`. Hence, the deadlocks/error observed in various API entrypoints in TensorFlow 2.13 when R iterators are supplied.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -27,8 +27,8 @@ jobs:
 
           - { os: ubuntu-latest, r: 'oldrel-1', python: '3.9' }
           - { os: ubuntu-latest, r: 'oldrel-2', python: '3.9' }
-          - { os: ubuntu-latest, r: '3.6',      python: '3.8' }
-          - { os: ubuntu-latest, r: '3.5',      python: '3.6' }
+          - { os: ubuntu-20.04,  r: '3.6',      python: '3.8' }
+          - { os: ubuntu-20.04,  r: '3.5',      python: '3.6' }
           # - { os: ubuntu-latest, r: 'devel'   , python: '3.8', http-user-agent: 'release' }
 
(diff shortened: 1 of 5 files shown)


t-kalinowski:
An alternate approach explored consisted of allowing `call_r_function()` to succeed when called from a background thread.

This consisted of:
- Implementing an "RIL" (R interpreter lock), similar to the Python GIL, to prevent concurrent R API access.
- Disabling the R stack checking mechanism by (temporarily) setting `R_CStackStart = -1`, which erroneously signals an R error when R is called from a non-main thread.

This approach is viable, and may be revisited in the future. It was not chosen at this time for following reasons:
- CRAN forbids R packages from changing the value of `R_CStackStart` (though there do appear to be exceptions made for some [exceptional cases](https://github.com/search?q=org%3Acran%20R_CStackLimit&type=code).)
- patching `R_CStackStart` is insufficient if R is built with `--enable-strict-barrier`. See the usage of the `R_CHECK_THREAD` macro in the R sources for more.

PR #1406, opened 2023-07-06
Ensure `py_capsule` finalizer runs on main thread

t-kalinowski:
While working on the latest threading issues in TF 2.13 (#1405), I observed that `py_capsule()` finalizer may access the R API from not the main thread, if the last `Py_DecRef()` on the object is called from a background thread. This PR fixes that, ensuring that the finalizer only accesses the R API from the main thread.


kevinushey:
Nice catch!

PR #1407, opened 2023-07-07
flush python stdout/stderr from event loop.

t-kalinowski:
Presently, long running Python calls in `repl_python()` that stream to stdout will not show any output in the R console until the Python call has fully returned. This PR add periodic calls to flush the stdout and stderr buffers from the event loop, ensuring that streaming python output is visible in the R console.

Related: https://github.com/rstudio/reticulate/issues/739
cc: @falbukrek

### DIFF
--- a/src/event_loop.cpp
+++ b/src/event_loop.cpp
@@ -38,6 +38,7 @@ namespace {
 
 // Tracks whether we've requested Python to poll on the main thread.
 volatile sig_atomic_t s_pollingRequested;
+bool s_flush_std_buffers = true;
 
 // Forward declarations
 int pollForEvents(void*);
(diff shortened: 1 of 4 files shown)


src/event_loop.cpp
```diff
@@ -87,6 +87,11 @@ int pollForEvents(void*) {
   // can be throttled)
   s_pollingRequested = 0;
 
+  // Periodically flush stdout/stderr buffers to ensure that any output from
+  // long-running Python calls is visible in the R console.
+  if (flush_std_buffers() != 0)
```
kevinushey, line 92: If this fails, I think we should either:

1. Stop attempting to flush buffers;
2. Only log the warning once.

My main worry is that this background thread might fail to flush buffers repeatedly and so overload the console (if something got messed up with Python's stdout / stderr streams).

kevinushey (approved):

PR #1412, opened 2023-07-12
Display python stdout when using knitr

### LINKED ISSUE #331: reticulate output in R chunks not captured in R Markdown
To reproduce, try rendering the following document (repro'ed with macOS 10.13.6 + R 3.5.1):

`````
---
title: "Reticulate"
output: md_document
---
  
```{r}
reticulate::py_run_string("print('hello')")
```
`````

You'll see that the printed output (`hello`) is not captured and reflected in the generated Markdown document.

I think the issue here is that `knitr` attempts to capture R output with a `sink()`; however, because we don't always install redirects for Python output back to R, that output is lost. See e.g.:

https://github.com/rstudio/reticulate/blob/82b2eb822469f3cd0dc0fe2ca34d55c65b998474/R/output.R#L3-L11

The fix here would likely be to ensure that the output streams are remapped when we're running `reticulate` within `knitr` -- or perhaps we should just always unconditionally remap the output streams.

See also: https://github.com/rstudio/rstudio/issues/3271

### LINKED ISSUE #1387: knitr engine doesn't capture python stdout
Rendering a quarto doc with two chunks like this produces no output:
````
```{python}
def print_something():
    print("Hello")
```

```{r}
py$print_something()
```
````

### PR DESCRIPTION
**Pull request by dfalbel**
Fixes #1387 and fix #331

The approach used in this PR is to register a [knitr hook](https://yihui.org/knitr/hooks/) that runs at the begining and end of each document chunk. In the begining of each chunk we remap Python's std streams to the same as R's (so knitr can capture them), at the end of the chunk we reset the streams to the default reticulate behavior (which is configurable with the `RETICULATE_REMAP_OUTPUT_STREAMS` env var).

Another approach would be to always set a [output remap class](https://github.com/rstudio/reticulate/blob/e80fa4385ffc15121632e39aafcb83345567ebe0/inst/python/rpytools/output.py#L120-L143) in Python that would check at runtime if knitting is in progress by checking `getOption('knitr.in.progress'))`, but this felt problematic because of possible concurrent calls, and the need to call R from the main thread.

This PR changes the default behavior of reticulate when rendering RMarkdown files, but users can easily go back to the previous behavior by setting `include=FALSE` in the chunk options. We also honor the `RETICULATE_REMAP_OUTPUT_STREAMS` if it's set:

- if it's set to "0" we never set-up the hook and there's no output redirection. 
- If it's set to "1" we also don't set the hook as redirections are already in place.

### CODE REVIEW

File: R/package.R [hunk 1]
```diff
@@ -66,6 +66,7 @@ ensure_python_initialized <- function(required_module = NULL) {
 
   # remap output streams to R output handlers
   remap_output_streams()
+  set_knitr_python_stdout_hook()
```
t-kalinowski, line 69: Should this hook be confined to `eng_python()`? {knitr} is only in Suggests, and to register this hook globally, we would need to move knitr to Depends.

inst/python/rpytools/output.py
```diff
@@ -143,15 +143,16 @@ def flush(self):
...
-
+def remap_output_streams(r_stdout, r_stderr, tty):
+  sys.stdout = OutputRemap(sys.stdout, r_stdout, tty)
+  sys.stderr = OutputRemap(sys.stderr, r_stderr, tty)
 
+def reset_output_streams():
+  # https://stackoverflow.com/a/51340381/3297472 suggests that the initial stdout
+  # is always available in `sys.__stdout__` so we just need to reassign to revert
+  # it.
+  sys.stdout = sys.__stdout__
```
t-kalinowski, line 154: I think it'd be safer to cache the previous streams, and then restore them. (So we don't incidentally undo any patching done by user or framework code). A context manager seems appropriate here.

tests/testthat/test-python-knitr-engine.R
```diff
@@ -63,3 +63,28 @@ test_that("knitr 'warning=FALSE' option", {
...
+  setwd(owd)
+
+  rendered <- test_path("resources", "knitr-print.md")
+  expect_snapshot_file(rendered)
+
+  # if remaping is set by default we have no way to check that the options
+  # is correctly reset
+  skip_if(!is.na(Sys.getenv("RETICULATE_REMAP_OUTPUT_STREAMS", unset = NA)))
+
+  py <- reticulate::import_builtins()
```
t-kalinowski, line 84: Small nit, we already have a `py` symbol in the package namespace. Can we call this `bt`?

t-kalinowski (changes requested):
This is great @dfalbel. I left some comments in line. 
Thank you!

R/package.R
dfalbel: `eng_python` is only used when there are Python chunks in the document, but in theory this problem could hapen even in R only chunks that call reticulate. Thus we need a global knitr hook.

inst/python/rpytools/output.py
dfalbel: Done. I used a context manager, even though we have to call `__enter__` and `__exit__` manually.

t-kalinowski (approved):

R/output.R
```diff
@@ -1,20 +1,70 @@
...
+
+  # we don't want to to force a knitr load namespace, so if it's already loaded
+  # we set the knitr hook, otherwise we schedule an onLoad hook.
+  if (isNamespaceLoaded("knitr")) {
+    set_knitr_hook()
+
+    # if knitr is already in progress here, this means that python was initialized
+    # during a chunk execution. We have to force an instant remap as the hook won't
+    # have a chance to run for that chunk.
+    if (isTRUE(getOption('knitr.in.progress'))) set_output_streams(tty = FALSE)
```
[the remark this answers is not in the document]
dfalbel: It's possible that a code in an R engine tries to print to stdout from python (eg using reticulate), in such case the hook is called with `before = TRUE` and at this point reticulate might have not loaded python yet - or even reticulate might not yet been namespace attached yet.

PR #1415, opened 2023-07-18
accept `requirements` in `virtualenv_{install,create}`

### LINKED ISSUE #1321: allow py_install / vitualenv_install to take a requirements.txt file?
Apologies if there is already a way to do this but I haven't found it. 

It would be great if there were simply a way to do the equivalent of `pip install -r requirements.txt` from reticulate, e.g. `virtualenv_install(requirements = "requirements.txt")` or such.  (I know renv already supports something like this, and I know reticulate has support for pipenv.  But many python projects simply have a requirements.txt floating around in their repo, and it would be nice to just clone a repo into RStudio and install these without having to switch over to the bash shell)

t-kalinowski:
Closes #1321

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -29,6 +29,9 @@
   python binary for creating a virtual environmnent. This is now the default
   method for finding the python binary when calling `virtualenv_create(version = <version>)`.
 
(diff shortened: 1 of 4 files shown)


cboettig:
:pray: thank you!

PR #1416, opened 2023-07-19
Correctly handles Fortran ordering in numpy character arrays

### LINKED ISSUE #1409: py_to_r returns array with wrong orders, where the array is numpy.array with character type
Dear reticulate team,
Thank you so much for writing this wonderful package! I face the array transformation inconsistence issue today:

In a python script
```
import pickle
import numpy as np
s1 = np.array([1,2,3])
s2 = np.array([3,4,5])
s = np.stack([s1,s2], axis = 1)
with open("s.pl", 'wb') as f:
    pickle.dump(s, f)

c1 = np.array(["1","2","3"])
c2 = np.array(["3","4","5"])
c = np.stack([c1,c2], axis = 1)
with open("c.pl", 'wb') as f:
    pickle.dump(c, f)
```

Then in R script under the same directory
```
library(reticulate)
# I use python_path to set a python version in my local env,
# which I ignore here.

py <- import_builtins()
pickle <- import("pickle")

with(py$open("s.pl", 'rb') %as% f, {
  s <- pickle$load(f)
})

with(py$open(".c.pl", 'rb') %as% f, {
  c <- pickle$load(f)
})
```
Then, in R console, if I print s and c
```
> s
     [,1] [,2]
[1,]    1    3
[2,]    2    4
[3,]    3    5
> c
     [,1] [,2]
[1,] "1"  "4" 
[2,] "3"  "3" 
[3,] "2"  "5" 

```
`s` is the right order, but `c` is not.

My R session is:
```
sessionInfo()
R version 4.3.0 (2023-04-21)
Platform: x86_64-apple-darwin22.4.0 (64-bit)
Running under: macOS Ventura 13.4.1

Matrix products: default
BLAS:   /usr/local/Cellar/openblas/0.3.23/lib/libopenblasp-r0.3.23.dylib 
LAPACK: /usr/local/Cellar/r/4.3.0_1/lib/R/lib/libRlapack.dylib;  LAPACK version 3.11.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/Los_Angeles
tzcode source: internal

attached base packages:

other attached packages:
[1] reticulate_1.30

 [1] compiler_4.3.0  here_1.0.1      Matrix_1.5-4    rprojroot_2.0.3 cli_3.6.1       withr_2.5.0     Rcpp_1.0.10    
 [8] grid_4.3.0      jsonlite_1.8.7  rlang_1.1.1     png_0.1-8       lattice_0.21-8 

```

Would you mind to double check this or if I need to pay specific attention on it? 
Thanks!
Songpeng

dfalbel:
Fixes #1409, also refer to linked issue for more details on the approach.

### DIFF
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -930,8 +930,10 @@ bool py_is_callable(PyObjectRef x) {
 
 // caches np.nditer function so we don't need to obtain it everytime we want to
 // cast numpy string arrays into R objects.
-PyObject* np_nditer;
 PyObject* get_np_nditer () {
+
+  static PyObject* np_nditer;
+
   if (np_nditer) {
     return np_nditer;
   }
@@ -1165,21 +1167,13 @@ SEXP py_to_r(PyObject* x, bool convert) {
         rArray = Rf_allocArray(STRSXP, dimsVector);
         RObject protectArray(rArray);
 
-        // empty tuple, casts to scalar value
-        PyObjectPtr itemArg(PyTuple_New(0));
+
         for (int i=0; i<len; i++) {
           PyObjectPtr el(PyIter_Next(iter)); // returns an scalar array.
-
-          PyObjectPtr itemFunc(PyObject_GetAttrString(el, "item"));
-          if (itemFunc.is_null()) {
-            throw PythonException(py_fetch_error());
-          }
-
-          PyObjectPtr pyStr(PyObject_Call(itemFunc, itemArg, NULL));
+          PyObjectPtr pyStr(PyObject_CallMethod(el, "item", NULL));
           if (pyStr.is_null()) {
             throw PythonException(py_fetch_error());
           }
-
           set_string_element(rArray, i, pyStr);
         }
         break;
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -928,6 +928,28 @@ bool py_is_callable(PyObjectRef x) {
     return py_is_callable(x.get());
 }
 
+// caches np.nditer function so we don't need to obtain it everytime we want to
+// cast numpy string arrays into R objects.
+PyObject* np_nditer;
+PyObject* get_np_nditer () {
+  if (np_nditer) {
+    return np_nditer;
+  }
+
+  requireNumPy();
+
+  PyObjectPtr numpy(PyImport_ImportModule("numpy"));
+  if (numpy.is_null()) {
+    throw PythonException(py_fetch_error());
+  }
+
+  np_nditer = PyObject_GetAttrString(numpy, "nditer");
+  if (np_nditer == NULL) {
+    throw PythonException(py_fetch_error());
+  }
+
+  return np_nditer;
+}
 
 // convert a python object to an R object
 SEXP py_to_r(PyObject* x, bool convert) {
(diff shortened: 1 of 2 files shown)


src/python.cpp
```diff
@@ -1126,19 +1148,38 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
-        if (itemFunc.is_null())
+
+        PyObjectPtr nditerArgs(PyTuple_New(1));
+        // PyTuple_SetItem steals reference the array, but it's already wraped
+        // into PyObjectPtr earlier (so it gets deleted after the scope of this function)
+        // To avoid trying to delete it twice, we need to increase its ref count here.
+        PyTuple_SetItem(nditerArgs, 0, (PyObject*)array);
+        Py_IncRef((PyObject*)array);
+
+        PyObjectPtr iter(PyObject_Call(get_np_nditer(), nditerArgs, NULL));
```
t-kalinowski, line 1159: Does it make sense to pass any flags to nditer, like `"readonly"`.
Do we need to pass something like `order = 'F'`, to ensure we get the elements in the correct order, regardless of how the underlying array buffer is arranged?

src/python.cpp
```diff
@@ -1126,19 +1148,38 @@ SEXP py_to_r(PyObject* x, bool convert) {
...
+        // empty tuple, casts to scalar value
+        PyObjectPtr itemArg(PyTuple_New(0));
         for (int i=0; i<len; i++) {
-          PyObjectPtr pyArgs(PyTuple_New(1));
-          // PyTuple_SetItem steals reference to object created by PyInt_FromLong
-          PyTuple_SetItem(pyArgs, 0, PyInt_FromLong(i));
-          PyObjectPtr pyStr(PyObject_Call(itemFunc, pyArgs, NULL));
+          PyObjectPtr el(PyIter_Next(iter)); // returns an scalar array.
+
+          PyObjectPtr itemFunc(PyObject_GetAttrString(el, "item"));
```
t-kalinowski, line 1173: Would it make sense to resolve the `item` method once from `np.ndarray.item`, before the loop, then call `item(el)` inside the loop?

src/python.cpp
```diff
@@ -928,6 +928,28 @@ bool py_is_callable(PyObjectRef x) {
     return py_is_callable(x.get());
 }
 
+// caches np.nditer function so we don't need to obtain it everytime we want to
+// cast numpy string arrays into R objects.
+PyObject* np_nditer;
```
t-kalinowski, line 933: Style nit, can this please be static defined inside the function rather than a global?

t-kalinowski (changes requested):
Looks great @dfalbel! Minor comments inline.

src/python.cpp
dfalbel: `'readonly'` is the default for nditer. We could force an ordering, but it would be redundant as we already create a Fortran ordered array earlier.

https://github.com/rstudio/reticulate/blob/70a23d693736e7f82b1bbb484579f2849ce41ddd/src/python.cpp#L1077-L1082

src/python.cpp
dfalbel: I figured I could use `PyObject_CallMethod` instead to avoid having to manually extract the class method.

PR #1418, opened 2023-07-20
Fix knitr output remapping from a clean new session

dfalbel:
Follow up for https://github.com/rstudio/reticulate/pull/1412#issuecomment-1644001337


dfalbel:
This should fix: https://github.com/tidymodels/parsnip/issues/990

PR #1421, opened 2023-07-20
Render markdown quietly to avoid flooding tests

dfalbel:
Should make testthat output less verbose.

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -625,7 +625,9 @@ eng_python_autoprint <- function(captured, options) {
 
   } else if (isHtml && py_has_method(value, "_repr_html_")) {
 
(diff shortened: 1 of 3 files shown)


R/knitr-engine.R
```diff
@@ -625,7 +625,9 @@ eng_python_autoprint <- function(captured, options) {
 
   } else if (isHtml && py_has_method(value, "_repr_html_")) {
 
-    data <- as_r_value(value$`_repr_html_`())
+    py_capture_output({
```
dfalbel, line 628: For some reason `_repr_html_` on the GH machine is also printing to stdout (with this we capture and discard the stdout output).

t-kalinowski:
Many thanks!!

PR #1423, opened 2023-07-25
Still include the stdout output captured before the autoprint mechanism.

dfalbel:
Should fix: https://github.com/rstudio/reticulate/issues/1387#issuecomment-1645724902
@matthew-brett Can you confirm this is the expected behavior?


dfalbel:
Thanks @matthew-brett , also added your example to the test suite :)

matthew-brett:
Excellent - thank you for adding that.

PR #1431, opened 2023-08-03
add `@` as py attr accessor

### LINKED ISSUE #251: should '$' and '[' behave differently for pandas DataFrames?

### LINKED ISSUE #1429: Missing method, even when not converting to R objects.
Hello - I'm trying to replicate the code snippet that's located in the top box of this HuggingFace NLP course: https://huggingface.co/learn/nlp-course/chapter3/4?fw=pt.

For reference, here's the Python code that I'd like to transcribe into reticulate:
`from datasets import load_dataset
from transformers import AutoTokenizer, DataCollatorWithPadding

raw_datasets = load_dataset("glue", "mrpc")
checkpoint = "bert-base-uncased"
tokenizer = AutoTokenizer.from_pretrained(checkpoint)

def tokenize_function(example):
    return tokenizer(example["sentence1"], example["sentence2"], truncation=True)

tokenized_datasets = raw_datasets.map(tokenize_function, batched=True)
data_collator = DataCollatorWithPadding(tokenizer=tokenizer)`

Here's the R code I write to import the datasets module, and pull the raw data:

`library(reticulate)
virtualenv_create("test")
py_install(c("datasets", "transformers"), envname = "test")
datasets <- import("datasets", convert = FALSE)
raw_datasets <- datasets$load_dataset("glue", "mrpc")`

You'll notice that, in Python, the raw_datasets object has a method called map.  When I start typing raw_datasets$..., I don't see the map method come up -- all I see are the 3 dictionary values ("Train," "Validation," "Test").  I do see it if I start typing raw_datasets$train$...

However, I see map as an attribute on the object when I pull it into R:
`py_list_attributes(raw_datasets)
`

t-kalinowski:
Closes #1429
Closes #251

This enables unambiguously accessing attrs in contexts where `$` mingles python attrs and items (e.g., with python dictionaries, pyspark tables). This also enables autocompletion in R outside the terminal. However, autocompletion in the IDE won't work until https://github.com/rstudio/rstudio/issues/13451 is resolved.


dfalbel:
Maybe we can add docs in this PR explaining the differences between `$`, `[`, `[[` and `@` so we can also fix https://github.com/rstudio/reticulate/issues/263.

t-kalinowski:
I don't think merging this at this time is a good idea. 

`@` is valuable new syntax, which doesn't come around very often. Using it for python attribute access at this time doesn't strike me as a good return on value--it wouldn't add any new capability over what `$` does already in practice. I don't have any better ideas for how to use `@` in reticulate at this time, but I'd rather reserve the syntax for a better idea, when it does come along.

Inside a package that uses reticulate, (e.g., pysparklyr), I think it's reasonable for package authors to opt-in to a "safer" attribute access method that doesn't fallback to `__getitem__` by defining an internal S3 method, e.g.,
```r
`@.python.buildin.object` <- function(x, name) py_to_r(reticulate::py_get_attr(x, name))
```

PR #1435, opened 2023-08-08
Add documentation for most most Python engine options.

dfalbel:
I think this covers all knitr options that the Python engine deals explicitly. There might be other options that are just used by knitr.


t-kalinowski:
Thank you!

Issue #207, opened 2018-03-31
pandas DataFrame with np.nan are not converted correctly


lepennec:
Dear all,
I've stumbled with an issue with the way NA are encoded in Python: it seems that the clasical way is to use np.nan whatever the data type as described in https://pandas.pydata.org/pandas-docs/stable/missing_data.html. This is the convention used by pd.read_csv for instance. When the data type is a "string" this leads to an issue: reticulate fails to convert such a DataFrame as shown in the following example
```
library(reticulate)

py_run_string('import pandas as pd')
py_run_string('import numpy as np')
py_run_string('date = ["2017/01/12", np.nan]')
py_run_string('data = {"date" : date}')
py_run_string('df = pd.DataFrame(data)')

py$df
py$dfcol
```
It seems that the column is converted as a list and not to a vector with missing data which seems to be the expected behavior from a Python user perspective.

Yours;

Erwan

jjallaire:
@kevinushey Could you take a look at this one?

kevinushey:
Perhaps rather than deserializing `np.nan` as `NaN` (which only exists in R for numeric types) we should use `NA` (which R defines for all vector types)? Or perhaps in this specific case we need to use `NA` instead of `NaN`.

That said, I would _highly_ recommend using NumPy `datetime64` objects rather than Python `datetime` objects within Pandas DataFrames.

kevinushey:
Now fixed up on master, but note that deserializing this object gives you list-columns, e.g.

```
> py$df
        date
1 2017/01/12
2        NaN
> str(py$df)
'data.frame':	2 obs. of  1 variable:
 $ date:List of 2
  ..$ : chr "2017/01/12"
  ..$ : num NaN
```

andyfarmerboy:
Note: a related conversion failure occurs where missing values in Pandas DataFrames are coded as "None".  Ideally these would be converted to NA in R at least in character columns.  The only way around the issue I've found so far is using .fillna method and inserting an arbitrary character string as a placeholder.

terrytangyuan:
Related issue but for NAs in general: https://github.com/rstudio/reticulate/issues/197

alexenge:
Has there been any update on this? It would be super useful if `NaN`s inside a string column of a pandas DataFrames would automatically be converted to R's `NA`, so that R could treat the column as a character vector instead of a list.

sushantd195:
[...]

Issue #197, opened 2018-03-26
What should `NA` be converted to in Python?


terrytangyuan:
Currently `NA` is converted to Python as `True` which doesn't make too much sense. This is because `NA` is logical in R.
```
> class(NA)
[1] "logical"
> r_to_py(NA)
True
```

Perhaps we should treat it as `np.nan` (similar to R's `NaN`) since that's the representation Python users use for missing data, e.g. [missing data in pandas objects](https://pandas.pydata.org/pandas-docs/stable/missing_data.html)? 
```
> r_to_py(NaN)
nan
```

"`NaN` is the the default missing value marker for reasons of computational speed and convenience. However, if you want to consider `inf` and `-inf` to be "NA" in computations, you can set `pandas.options.mode.use_inf_as_na = True`" ([source](https://pandas.pydata.org/pandas-docs/stable/missing_data.html#values-considered-missing)). So this might be something we need to keep in mind when deciding how we approach this.

Jiang-Li:
What I just observed are: the missing values in a character column was converted to the character of NA, and the missing values in an integer column was converted to -2147483648

```{r}
set.seed(1)
N <- 1000
df <- tibble(
dimension1 = sample(c("I", "II", "III"), N, replace = T),
dimension2 = sample(c("A", "B", "C"), N, replace = T),
measure1 = sample(1:10, N, replace = T),
measure2 = sample(1:10, N, replace = T)
)

df <- as_tibble(lapply(df, function(r) r[sample(c(TRUE, NA), prob = c(0.85, 0.15), size = length(r), replace = TRUE) ]))

```

```{python}
print(r.df)
```

hilaryparker:
Just confirming that missing values in integer columns are also being translated to `-2147483648` in my context.

kevinushey:
Some more context:

https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html
https://stackoverflow.com/questions/12708807/numpy-integer-nan

From what I understand, NumPy integer arrays don't have a 'missing value' number in the same way R integer vectors do, so we would have to either:

1. Convert from an R integer vector into a NumPy "float64" array;
2. Convert from an R integer vector into the Pandas "Int64" extension integer type.

I'm not sure if either of these is preferable, or even if this is something reticulate would want to do automatically. It seems instead like users should be able to request what kind of conversions should happen when going between the R and Python worlds, but I'm not sure what the interface for that should look like yet.

[...]

PR #1439, opened 2023-08-10
Preserve NAs when casting R data.frames to pandas.

dfalbel:
Related #197 and #207

This PR adds an optional behavior for handling of R data.frame atomic columns when converting into pandas data.frames. The main motivation for this PR is to better handle R data frames that contain columns with `NA`s.

If the `reticulate.pandas_use_nullable_dtypes` is set, instead of converting into raw numpy data types, we use (newish) pandas [nullable data types](https://pandas.pydata.org/docs/user_guide/integer_na.html), that have built-in support for NA`s .

With the new behavior, doing:

```
df <- data.frame(
  int = c(NA, 1:4),
  num = c(NA, rnorm(4)),
  lgl = c(NA, rep(c(TRUE, FALSE), 2)),
  string = c(NA, letters[1:4])
)
options(reticulate.pandas_use_nullable_dtypes = TRUE)
r_to_py(df)
```

Results in:

```
    int       num    lgl string
0  <NA>      <NA>   <NA>   <NA>
1     1  0.491903   True      a
2     2   1.54789  False      b
3     3  0.470248   True      c
4     4  0.125063  False      d
```

While with old behavior we get:

```
          int       num    lgl string
0 -2147483648       NaN   True   None
1           1  0.282169   True      a
2           2 -0.001555  False      b
3           3 -0.190735   True      c
4           4  0.678618  False      d
```

Preserving NA`s helps making round trip casts more reliable.

Benchmark shows twice as slow for a 100k row df, which is somewhat expected as we need to build a boolean mask for each column.

```
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory              time       gc      
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>              <list>     <list>  
1 old           3.2ms   3.49ms      280.    54.8KB     2.95    95     1      339ms <NULL> <Rprofmem [22 × 3]> <bench_tm> <tibble>
2 new          6.72ms   6.97ms      143.    54.8KB     0       72     0      504ms <NULL> <Rprofmem [22 × 3]> <bench_tm> <tibble>
```

<details>
<summary>Code</summary>

```
df <- data.frame(
  int = c(NA, 1:4),
  num = c(NA, rnorm(4)),
  lgl = c(NA, rep(c(TRUE, FALSE), 2)),
  string = c(NA, letters[1:4])
)


library(dplyr)
devtools::load_all()

set.seed(100)
x <- df %>% sample_n(100000, replace = TRUE)

bench::mark(check = FALSE,
  old = withr::with_options(c(reticulate.pandas_force_numpy = TRUE), r_to_py(x)),
  new = withr::with_options(c(reticulate.pandas_force_numpy = FALSE), r_to_py(x))
)
```

</details>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # reticulate (development version)
 
+- reticulate now supports casting R data.frames to Pandas data.frames using nullable
+  data types allowing users to preserve NA's from R atomic vectors. This feature is
+  opt-in and can be enabled by setting the R option `reticulate.pandas_use_nullable_dtypes`
+  to `TRUE`. (#1439)
+
 # reticulate 1.31
 
 ## Python Installation Management
--- a/src/libpython.h
+++ b/src/libpython.h
@@ -552,6 +552,9 @@ LIBPYTHON_EXTERN void **PyArray_API;
           (*(PyObject * (*)(PyTypeObject *, int, npy_intp *, int, npy_intp *, void *, int, int, PyObject *)) \
              PyArray_API[93])
 
+#define PyArray_SimpleNew(nd, dims, typenum) \
+          PyArray_New(&PyArray_Type, nd, dims, typenum, NULL, NULL, 0, 0, NULL)
+
 inline void* PyArray_DATA(PyArrayObject *arr) {
   return ((PyArrayObject_fields *)arr)->data;
 }
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -3476,7 +3476,7 @@ PyObject* r_to_py_pandas_nullable_series (const RObject& column, const bool conv
     // the minimum pandas version.
     // we show a warning and force the numpy construction.
     Rcpp::warning(
-      "Nullable data types require pandas version >= 1.0.0. "
+      "Nullable data types require pandas version >= 1.2.0. "
       "Forcing numpy cast. Use `options(reticulate.pandas_force_numpy = TRUE)` "
       "to disable this warning."
     );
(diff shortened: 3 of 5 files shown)


dfalbel:
This PR, will also cast `NA_character_` into `None` when converting R data.frames into Pandas data.frames. The old behavior would convert into the string `'NA'` possibly causing data loss.

With CRAN reticulate

``` r
library(reticulate)
df <- data.frame(
  string = c(NA, letters[1:4])
)
p <- r_to_py(df)
p
#>   string
#> 0     NA
#> 1      a
#> 2      b
#> 3      c
#> 4      d
p$string[0]
#> 'NA'
```

With this PR:

``` r
library(reticulate)
options(reticulate.pandas_force_numpy = TRUE)
df <- data.frame(
  string = c(NA, letters[1:4])
)
p <- r_to_py(df)
p
#>   string
#> 0   None
#> 1      a
#> 2      b
#> 3      c
#> 4      d
p$string[0]
#> None
```

t-kalinowski (approved):
Great work @dfalbel!

Can you please add a NEWS entry?

Seeing this makes me wonder if we should similarly take this opportunity to change `NA_character` converting to Python `None` outside of a data.frame as well. E.g., `r_to_py(c(NA, letters[1:3]))` (or if pandas is not installed).

dfalbel:
Yeah, I think we can do this in a future PR. We have to think if we want to be consistent with other data types, eg, what should we do with NA's in numeric, integers and logicals when outside a pandas data frame. And what's the desired behavior when getting back from R - should we simplify the vectors?

PR #1444, opened 2023-08-14
Make sure we initialize python before calling `r_to_py_impl`

### LINKED ISSUE #1442: Segfault with `np_array()`
reprex:
```r
reticulate::np_array(array(1:12, c(3, 4)))
```

Looks like we are missing an `ensure_python_initialized()` and/or `ensure_numpy_available()` call somewhere. It seems we dispatch straight to c++ code and attempt to call `PyArray_New()` before the symbol is loaded.

dfalbel:
Fixes #1442


t-kalinowski:
Thank you!

PR #1459, opened 2023-08-23
CRAN guardrails

### LINKED ISSUE #1441: Reticulate can initialize Python on CRAN
With reticulate release 1.31, an issue was encountered with CRAN, where a few reverse dependencies of reticulate packages began failing tests on the Fedora r-devel testing machine. 

After some investigation, it became apparent that on that testing machine, there existed a venv at `~/.virtualenvs/r-reticulate`. How that venv came into existence is an open question, but it must have came into existence some time ago, because it was derived from `/usr/bin/python`, which at the time was Python 3.9. At some later time, the machine was upgraded to Fedora 36, and `/usr/bin/python/` upgraded to Python 3.10 with it, making the `~/.virtualenvs/r-reticulate` venv unusable (at the very least, because symlinks to versioned targets like `/usr/bin/python3.9` were broken.)

With the updated "Order of Discovery", the invalid "r-reticulate" venv was now being selected by reticulate when initializing Python, resulting in the failing tests. 

The broken venv has now been removed, tests are all back to passing, and an urgent new release of reticulate is not necessary at this time. There are A few action items from this exercise that we should handle before the next release:

- [x] Improved error message when reticulate attempts to load a broken venv like this. At the least, it should include the path to the python. We'll need to wrap the `system2()` call in `python_config_impl()` with a `tryCatch()` to augment the error message. 
- [ ] A guardrail that deters people from creating or finding venvs on CRAN. When reticulate detects its running on CRAN, we should set `WORKON_HOME` to a private location, perhaps `tempfile()`.
- [ ] In `initialize_python()`, issue a warning if we're on a CRAN machine, and then in a subsequent release, turn it into an error. Packages should not be initializing Python on CRAN. (See `reticulate:::check_forbidden_initialization()` for a start on this)
- [ ] Update guidance in vignettes about how to avoid initializing Python on CRAN (guard examples and tests with `if(py_available()) {...}`, and/or `if(py_module_available('foo'))`
- [x] Reticulate attempting to load a broken python installation should be a warning, not an error. Reticulate should emit a warning and move on the the next python in the 'Order of Discovery'.
- [x] Sprinkle a few more `reticulate:::check_forbidden_install()` calls across the venv installation management functions.

t-kalinowski:
Closes #1441 

- [x] Improved error message when reticulate attempts to load a broken venv like this. At the least, it should include the path to the python. We'll need to wrap the `system2()` call in `python_config_impl()` with a `tryCatch()` to augment the error message. 
- [ ] A guardrail that deters people from creating or finding venvs on CRAN. When reticulate detects its running on CRAN, we should set `WORKON_HOME` to a private location, perhaps `tempfile()`.
- [ ] In `initialize_python()`, issue a warning if we're on a CRAN machine, and then in a subsequent release, turn it into an error. Packages should not be initializing Python on CRAN. (See `reticulate:::check_forbidden_initialization()` for a start on this)
- [ ] Update guidance in vignettes about how to avoid initializing Python on CRAN (guard examples and tests with `if(py_available()) {...}`, and/or `if(py_module_available('foo'))`
- [x] Reticulate attempting to load a broken python installation should be a warning, not an error. Reticulate should emit a warning and move on the the next python in the 'Order of Discovery'.
- [x] Sprinkle a few more `reticulate:::check_forbidden_install()` calls across the venv installation management functions.


t-kalinowski:
Revisiting this, I don't see a reliable way to detect if we're running on a CRAN machine. The two tasks that would modify reticulate behavior on CRAN test machines I will leave unimplemented.

Additionally, the guidance of using `py_module_available()` is already discussed here: https://rstudio.github.io/reticulate/articles/package.html#checking-and-testing-on-cran

While `py_module_available()` can cause reticulate to initialize python, I'm now thinking this is an acceptable approach, even on CRAN machines. If the need for stringent guarding against python initialization reappears, we can revisit and maybe export `reticulate:::python_module_available()`, which probes the python installation via a `system2()` call rather than attempting to import in the current session. But I'm inclined to not expand/clutter the exported package API at this moment, for the (hopefully one-off) scenario where a forbidden broken python installation is present on a CRAN testing machine.

PR #1489, opened 2023-10-06
avoid sys.version in config.py

johnxhobbs:
Previous logic breaks when Python 3.10+ is used. Replace with proper version check, as per [documentation](https://docs.python.org/3/library/sys.html#sys.version)


t-kalinowski:
Thank you!

PR #1507, opened 2023-11-22
Save plot images in different formats with `opts_chunks$dev`

Rumengol:
The [knitr documentation](https://yihui.org/knitr/options/#plots) indicates that it is possible to pass a vector of target devices to save the same plot in different formats. Reticulate uses `dir.create()` to create the destination directory if it doesn't exist, however this function can only accept one path and will fail with multiple target devices, as `knitr::fig_path()` will return a vector of 2 paths.

This PR solves the issue by considering the return type of `knitr::fig_path()` as a vector of strings rather than a single string and calling `lapply` on functions that relies on a single element.


t-kalinowski:
Thank you very much!

PR #1519, opened 2024-01-09
Unwrapping extptr stored in capsules

### LINKED ISSUE #1511: `py_to_r` does not unwrap external pointer
Currently the `py_to_r` function does not unwrap an external pointer from a capsule. Namely `py_to_r(r_to_py(x))` is not `x` if `x` is a R external pointer.

Code to reproduce:
```r
library(reticulate)
library(Rcpp)

cppFunction("
SEXP fun() {
  static int k = 17;
  Rcpp::XPtr< int > ret(&k);
  return ret;
} 
")

x = fun()

identical( py_to_r(r_to_py(x)), x )
```

I'm open to writing a pull request with the correct the behaviour of `py_to_r()` function in `src/python.cpp`.

llaniewski:
R external pointers were wrapped as a python Capsule (in `r_extptr_capsule(...)`) but never properly unwrapped by `py_to_r`. As the `r_extptr_capsule(...)` function stores the pointer address in the Capsule, and the corresponding R object in the Capsule's "context" field, the original R external pointer can be easily retrieved. This preserves the external pointer's finaliser.

Example of usage:
```R
library(Rcpp)
library(reticulate)

# Mock class for testing
Rcpp::sourceCpp(code='
#include <Rcpp.h>
using namespace Rcpp;

class AClass {
public:
  AClass() {
    printf("AClass created\\n");
  }
  ~AClass() {
    printf("AClass destroyed\\n");
  }
};

// [[Rcpp::export]]
SEXP getA() {
  AClass* ptr = new AClass;
  return Rcpp::XPtr< AClass >(ptr);
}
')

x = getA() #<--- AClass created
xpy = r_to_py(x)
x_ = py_to_r(xpy)

identical(x,x_) #<--- TRUE

# Removing objects one by one with gc called to check when the C++ object is destroyed.
rm(x); for (i in 1:3) gc(full = TRUE);
rm(xpy); for (i in 1:3) gc(full = TRUE);
rm(x_); for (i in 1:3) gc(full = TRUE); #<--- AClass destroyed
```

This PR should not break any dependencies, as the capsule visible in python still stores the same pointer. I only changed its "name" field from `NULL` to `"r_extptr"` to easily identify it.
This PR fixes #1511

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # reticulate (development version)
 
+- R external pointers (EXTPTRSXP objects) now round-trip through 
+  `py_to_r(r_to_py(x))` successfully.
+  (reported in #1511, fixed in #1519, contributed by @llaniewski).
+
 - Fixed issue where `virtualenv_create()` would error on Ubuntu 22.04 when
   using the system python as a base. (#1495, fixed in #1496).
 
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -107,6 +107,7 @@ SEXP py_fetch_error(bool maybe_reuse_cached_r_trace = false);
 
 
 const char *r_object_string = "r_object";
+const char *r_extptr_string = "r_extptr";
 
 // wrap an R object in a longer-lived python object "capsule"
 SEXP py_capsule_read(PyObject* capsule) {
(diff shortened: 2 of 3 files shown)


t-kalinowski:
Very nice, Thank you!

PR #1524, opened 2024-01-24
Fix issue with unsorted indices in csc_matrix

rcannood:
This PR solves an issue with converting csc matrices with unsorted indices, similar to #727.

* Add test to check whether reticulate can convert csc_matrix, csr_matrix and coo_matrix with unsorted indices
* Sort indices of csc_matrix before converting to dgCMatrix

In https://github.com/rcannood/reticulate/pull/1, the CI checks fails for the first commit (in which I added a test to detect this problem), but the CI checks succeed for the second commit (in which I added the fix).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
 - Fixed issue where `virtualenv_create()` would error on Ubuntu 22.04 when
   using the system python as a base. (#1495, fixed in #1496).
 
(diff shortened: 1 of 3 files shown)


t-kalinowski:
@rcannood Thank you for the PR, this looks great. Would you like to add a NEWS entry?

rcannood:
Done!

t-kalinowski:
Thank you!

Issue #1535, opened 2024-02-05
Regex bug in python_info_condaenv_find


cameronbriar:
The regular expression [here](https://github.com/rstudio/reticulate/blob/v1.35.0/R/python-tools.R#L189) added in this [change set](https://github.com/rstudio/reticulate/pull/1154/files) appears incorrect, resulting in failures to resolve the correct Conda environment.

To reproduce (inside a Conda-env enabled workspace):

```
> library(reticulate)
> reticulate::use_condaenv('my-env')
Error in normalizePath(conda, winslash = "/", mustWork = TRUE) : 
  path[1]="# cmd: /opt/conda/bin/mamba create": No such file or directory
```

The code linked above attempts to process `/opt/conda/envs/my-env/conda-meta/history` and, if I'm not mistaken, use a regex `^#\\s+cmd: (.+)\\s+(create|rename)\\s+.*` to pull the script path from a line like `# cmd: /opt/conda/bin/mamba create `. Note the trailing space in this example -- which the regex requires at least one of. **In my environment, there's no trailing space and the regex fails, resulting in the error shown above.**

Without refactoring how this is done, I'd propose the following fix to the regex: `^#\\s+cmd: (.+)\\s+(create|rename).*` (or similar)

I'm aware of two workarounds to this issue:

1. (Preferred) In my environment, I can `sed -i 's/mamba create/mamba create /g' /opt/conda/envs/my-env/conda-meta/history` before running the `reticulate::use_condaenv` method OR
2. Use the method described [here](https://github.com/rstudio/reticulate/issues/1460#issuecomment-1803762448) prior to running the `reticulate` method

```
assignInNamespace("is_conda_python", function(x){ return(FALSE) }, ns="reticulate")
```

I'll seek to submit a pull request when I can -- but wanted to create the issue in case it helps the next person in the meantime.

t-kalinowski:
Can you share the contents of the conda history file and how it was generated?

paulimer:
I have the same issue as @jacp1. When you create a environment not with `conda create -n my-env` but with `conda env create -f my-env.yaml`, `use_condaenv()` fails because of the regex. Instead of returning the path to the conda command it returns `/home/username/miniconda3/bin/conda env`, the regex matching `env` as well.

I am also surprised that this fails, given that creating a environment from a yaml file seems common.

I have written an alternative regex that accounts for creating environments with conda env create:
```r
# note perl = TRUE
# line 189 of python-tools.R
[...]

PR #1543, opened 2024-03-04
fix regex to support environments created with conda env create

colinvwood:
Looks like the previous regex only supported environments created with the `conda create` or `conda rename` commands, this adds support for `conda env create`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,9 @@
 - Fixed an issue where `py_discover_config()` would discover `python` (v2) on the PATH
   in preference of `python3` on the PATH. (#1547)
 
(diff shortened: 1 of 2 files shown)


t-kalinowski:
Thanks for opening the PR. Note, there is some active discussion about this in: https://github.com/rstudio/reticulate/issues/1535#issuecomment-1976597254 

Should the first capture group be marked non-greedy?

colinvwood:
That might be better! I guess I was assuming that because the the author went through the trouble of matching create/rename that there were other options being implicitly excluded, but I only really read enough of the code to get things working on my end.

t-kalinowski:
Thanks for the PR!

Issue #1024, opened 2021-07-02
Can't call some python callables


t-kalinowski:
Hi,

Should R functions converted to python still be callable from R?

``` r
library(reticulate)
fn <- r_to_py(function(x) x)
try(fn(1)) # should this succeed?
#> Error in fn(1) : could not find function "fn"
```

Calling `py_to_r` enables calling again, but doesn’t change the class or printed output.

``` r
fn
#> <function make_python_function.<locals>.python_function at 0x7f243344aef0>
class(fn)
#> [1] "python.builtin.function" "python.builtin.object"
fn <- py_to_r(fn)
fn
#> <function make_python_function.<locals>.python_function at 0x7f243344aef0>
class(fn)
#> [1] "python.builtin.function" "python.builtin.object"
fn(1)
#> 1.0
```

(Changing the value of `convert` seems to make no difference.)

[This comment](https://github.com/rstudio/reticulate/blob/4967abddb35865a1be6ad839298695f1481bcade/R/python.R#L220) and the extensive use of `as.environment` suggests that maybe the intent is for all python callables to be callable from R and this is a bug?

It would be very convenient if all python callables are always directly callable from R without needing to go though `py_call`.

<sup>Created on 2021-07-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

kevinushey:
I don't think so, because Python objects (as represented in `reticulate`) are just R environments with a special class, and so cannot be plainly called.

Any callable Python function is really just a wrapper R function over the underlying Python object; e.g.

```
library(reticulate)

builtins <- import_builtins(convert = TRUE)
abs <- builtins$abs
typeof(abs)
unclass(abs)
```

and you should see something like:

```
> abs <- builtins$abs
> class(abs)
[1] "python.builtin.builtin_function_or_method"
[2] "python.builtin.object"                    
> typeof(abs)
[1] "closure"
> typeof(abs)
[1] "closure"
> unclass(abs)
function(...) {
    
    dots <- py_resolve_dots(list(...))
    result <- py_call_impl(callable, dots$args, dots$keywords)
    
    if (convert)
      result <- py_to_r(result)
      
    if (is.null(result))
      invisible(result)
    else
      result
    
  }
<bytecode: 0x62500cc631c8>
<environment: 0x62500ba6c620>
attr(,"py_object")
<built-in function abs>
```

[...]

PR #1552, opened 2024-03-19
`py_to_r` refactor (c++)

t-kalinowski:
This PR diff is unfortunately larger than I'd like. It started with trying to fix one bug, but with the way the `py_to_r()` conversion routines were factored, I realized it made sense to implement the fix "upstream", in `PyObjectRef`, rather than at each cpp call site of `py_to_r()` and `py_ref()`. That led me to need to touch code in lots of places (and load it all in my head). As I refactored, I discovered more bugs and, with the code fresh in my mind, I also implemented some straightforward fixes and optimizations. The size of the changes snowballed, and at this point, dividing them into smaller PRs doesn't seem worth the effort.

User facing changes:
 - Callable python objects created with `convert = FALSE` now *always* get wrapped in a function. 
   (i.e., `typeof(py_eval("lambda x: x", convert = FALSE)) == "function")`). This now works:
   ```r
   fn <- py_eval("lambda x: x + 1", convert = FALSE)
   fn(2) # error in current version, required using py_call()
         # now returns a py ref to '3.0' with this PR
   ```
 - `py_to_r()` S3 methods now are called on python objects supplied as args to R functions being called from Python if the R function was converted with `convert = TRUE`. This used to error, and now works:
   ```r
   myfoo <- reticulate::py_eval("type('MyFoo', (), {})")()
   py_docall <- reticulate::py_eval("lambda fn, *args: fn(*args)")
   py_to_r.__main__.MyFoo <- function(x) 42
   registerS3method("py_to_r", "__main__.MyFoo", 
                    py_to_r.__main__.MyFoo, 
                    asNamespace("reticulate"))
   py_docall(function(x) {
     if (x != 42) stop("custom py_to_r method not found")
   }, myfoo)    
   ```
 
 - `py_to_r(x)` no longer signals an error if `x` is not a python object. In that case, `x` is returned unmodified. 
   (This is to avoid introducing new errors in user code, if users wrote previous workarounds for S3 generics that *should* have been getting called, but weren't.)
 - `attr(x, "tzone")` attributes are (better) preserved when converting POSIXt to Python.
   POSIXt types with a non-empty `tzone` attr will always convert to a datetime.datetime, 
   otherwise they will convert to numpy datetime64[ns] arrays.
 - Fixed an issue where calling `py_set_item()` on a subclassed dict would not invoke a custom `__setitem__` method.
 - `py_del_attr(x, name)` now returns `x` invisibly
 - `source_python()` no longer exports assigns the `r` symbol 
   (the "R Interface object" that is used by python code get a reference to the R globalenv)

 
 Developer facing changes:
 
 
- in c++ functions, `r_to_py(RObject, convert)` is now all you need to go from SEXP to PyObject*
- in c++ functions, `py_to_r(PyObject*, convert)` is now all you need to go from PyObject* to SEXP
- `py_to_r(PyObject*, convert = false)` will now *always* return a PyObjectRef.
- `py_to_r(PyObject*, convert = true)` will now invoke `py_to_r()` S3 methods, if the object is not a "simple" python object.
- `r_to_py(RObject, convert)` now returns the underlying PyObject* if it is a `PyObjectRef` already.
- The SEXP underlying PyObjectRef can now be either an R environment, or an R closure (if PyObject* is callable). All the common logic for building up a PyObjectRef for presentation in R is now done within the PyObjectRef constructor. This includes building up the closure if the PyObject is callable, and building up the S3 `class` attribute.
- `py_to_r_wrapper()` S3 generic will be called on all callable PyObjectRefs 
  (both with convert = TRUE and convert = FALSE)
 
- in R code there is now generally no need to call 
     `if(py_is_module_proxy(x)) py_resolve_module_proxy(x)`
   or
    `ensure_python_initialized()`
   This is handled upstream in `PyObjectRef` or the individual c++ methods
   

Optimizations:
 - logical R arrays get converted to numpy bool arrays that are strided views of the R array (avoiding a full copy)
 - removed (now) unnecessary S3 methods/code that are now handled in c++
 - misc small speedups (e.g., repeated calls in for loops to INTEGER() and similar have been moved out)
  
 In some ad-hoc benchmarks simulating some series of operations with python objects, I see ~2x speedup. E.g., this is ~2x faster:
 ```r
{
  x <- np_array(array(runif(1000)))
  y <- np_array(array(runif(1000)))
  z <- x + y
  z <- z[x > .5 & y < .5]
  py_to_r(z)
}
```
 
Though, in some minimal benchmarks of a callable created with `convert = FALSE`, there is now a small slowdown due the extra work of building up an R closure (including S3 dispatch of `py_to_r_wrapper()`). I.e. `py_eval("lambda x: x", convert = FALSE)`

### CODE REVIEW

File: src/python.cpp [hunk 1]
```diff
@@ -489,58 +629,57 @@ std::vector<std::string> py_class_names(PyObject* object) {
...
 
-  // wrap
-  PyObjectRef ref(object, convert);
+  // if it's an iterator, include python.builtin.iterator, before python.builtin.object
+  if(PyIter_Check(object))
+    classNames.insert(classNames.end() - 1, "python.builtin.iterator");
 
-  // class attribute
-  std::vector<std::string> attrClass;
+  return eval_call(r_func_py_filter_classes, Rcpp::wrap(classNames));
```
kevinushey, line 652: `Rcpp::wrap()` returns a raw `SEXP` which needs to be protected.

src/python.cpp
```diff
@@ -992,8 +1115,120 @@ PyObject* get_np_nditer () {
...
+
+SEXP py_to_r_wrapper(SEXP x) {
+  SEXP x2 = eval_call_in_userenv(r_func_py_to_r_wrapper, x);
+  if(x == x2) // no method, py_to_r_wrapper.default() reflects
+    return(x);
+
+  // copy over all attributes ("class" and "py_object", typically)
+  PROTECT(x2);
+  SEXP a = ATTRIB(x);
+  while (CAR(a) != R_NilValue) {
```
kevinushey, line 1142: I think this wants to check `a != R_NilValue`; check how R iterates over the attributes pairlist in https://github.com/wch/r-source/blob/25ed2f1ac6256ece6b2f355f158c21196127acb5/src/main/attrib.c#L152-L162

kevinushey: It's also worth checking if `Rf_copyMostAttrib()` does what you want, e.g. https://github.com/wch/r-source/blob/25ed2f1ac6256ece6b2f355f158c21196127acb5/src/main/attrib.c#L280-L299 -- that is, do you want to copy names / dimensions too?

src/python.cpp
```diff
@@ -1084,7 +1319,12 @@ SEXP py_to_r(PyObject* x, bool convert) {
   }
 
   // dict
-  else if (PyDict_CheckExact(x)) {
+  if (PyDict_CheckExact(x)) {
+    // if you tempted to change this to PyDict_Check() to allow subclasses, don't.
```
kevinushey, line 1323: 😆

src/python.cpp
```diff
@@ -1760,42 +1935,37 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
...
-  if (x.hasAttribute("dim") && requireNumPy()) {
+  if (Rf_getAttrib(sexp, R_DimSymbol) != R_NilValue &&
+      requireNumPy()) {
     return r_to_py_numpy(x, convert);
   }
 
   // integer (pass length 1 vectors as scalars, otherwise pass list)
   if (type == INTSXP) {
 
+    R_xlen_t len = LENGTH(sexp);
```
kevinushey, line 1957: XLENGTH?

src/python.cpp
```diff
@@ -1807,17 +1977,16 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
   if (type == REALSXP) {
 
     // handle scalars
-    if (LENGTH(sexp) == 1) {
-      double value = REAL(sexp)[0];
-      return PyFloat_FromDouble(value);
-    }
+    R_xlen_t len = LENGTH(sexp);
```
kevinushey, line 1980: XLENGTH?

src/python.cpp
```diff
@@ -1828,18 +1997,18 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
   // complex (pass length 1 vectors as scalars, otherwise pass list)
   if (type == CPLXSXP) {
 
-    // handle scalars
-    if (LENGTH(sexp) == 1) {
-      Rcomplex cplx = COMPLEX(sexp)[0];
+    R_xlen_t len = LENGTH(sexp);
```
kevinushey, line 2000: XLENGTH?

src/python.cpp
```diff
@@ -1850,18 +2019,17 @@ PyObject* r_to_py_cpp(RObject x, bool convert) {
   // logical (pass length 1 vectors as scalars, otherwise pass list)
   if (type == LGLSXP) {
 
-    // handle scalars
-    if (LENGTH(sexp) == 1) {
-      int value = LOGICAL(sexp)[0];
-      return PyBool_FromLong(value);
-    }
+    R_xlen_t len = LENGTH(sexp);
```
kevinushey, line 2022: XLENGTH? (check other usages of LENGTH below as well)

kevinushey (commented):
Overall LGTM, but it's a large PR so it's challenging to review in depth. Are there more tests we could add to validate that we haven't changed any existing behavior unexpectedly?

kevinushey:
> Callable python objects created with convert = FALSE now always get wrapped in a function.
(i.e., typeof(py_eval("lambda x: x", convert = FALSE)) == "function")).

Can you elaborate on the motivation? What if a user wants to create a Python function from R, that is later passed back into Python? A dummy example:

```
py_run_string("def apply(f, x): return f(x)")
fn <- py_eval("lambda x: x + 1")
main <- import_main()
main$apply(fn, 2)
```

src/python.cpp
t-kalinowski: Will do. Just for my own understanding, does `Rcpp::wrap()` keep the object protected until `eval_call()` returns? (btw, eval_call() also protects args that come in).

src/python.cpp
```diff
@@ -432,22 +465,129 @@ PyObject* py_import(const std::string& module) {
...
+  }
+
+  // Rf_PrintValue(get_r_trace(false, false));
+
+  return Rf_eval(call, R_BaseEnv);
+}
+
+SEXP eval_call(SEXP r_func, SEXP arg) {
+  SEXP cl = Rf_lang2(r_func, arg);
+  RObject cl_(cl); // protect
```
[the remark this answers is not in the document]
t-kalinowski: I think I'll switch this to a regular `PROTECT` call, to avoid confusion.

t-kalinowski:
The example works just fine - and should be unchanged from before:
``` r
library(reticulate)

py_run_string("def apply(f, x): return f(x)")
fn <- py_eval("lambda x: x + 1")
main <- import_main()
main$apply(fn, 2)
#> [1] 3
```
What is different is what happens when `convert = FALSE`
```r
fn <- py_eval("lambda x: x + 1", convert = FALSE)
fn(2) # error in current version, returns a py ref to '3.0' with this PR
```

t-kalinowski:
The specific bug that I was initially trying to fix was in tfdatasets:

```r
library(tfdatasets)
make_csv_dataset("hearts.csv") |>
  dataset_map(\(x) { ... })
```

Where I was expecting `x` to come in as a named R list, but instead was coming in as an unconverted OrderedDict. Pulling at the thread, I noticed that we already define the S3 method, but that S3 methods are not called by the C++ version of `py_to_r()`.

t-kalinowski:
Closes #1024

src/python.cpp
```diff
@@ -432,22 +465,129 @@ PyObject* py_import(const std::string& module) {
...
+}
+
+SEXP eval_call(SEXP r_func, SEXP arg1, bool arg2) {
+  return eval_call(r_func, arg1, Rf_ScalarLogical(arg2));
+}
+
+SEXP eval_call_in_userenv(SEXP r_func, SEXP arg) {
+  SEXP cl = Rf_lang2(r_func, arg);
+  RObject cl_(cl); // protect
+  return Rf_eval(cl, current_env());
```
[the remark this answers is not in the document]
t-kalinowski: `current_env()` should always return something that's already protected. (because it's on the R call stack)
It also isn't quite right for enabling `UseMethod("r_to_py")` to find methods in the calling env, since it resolves a reticualte internal frame instead, we'll need a new function for that anyway.

t-kalinowski:
> Are there more tests we could add to validate that we haven't changed any existing behavior unexpectedly?

Do you have any specific edge-cases in mind? I already expanded some tests related to conversion, and will still add a few more related to logical numpy array conversion. Anything else?

t-kalinowski:
A quick benchmark:
This snippet across the current main branch and this PR branch:
``` r
library(reticulate)
bench::mark(np_array_work = {
  x <- np_array(array(runif(1000)))
  y <- np_array(array(runif(1000)))
  z <- x + y
  z <- z[x > .5 & y < .5]
  py_to_r(z)
})
```
```r
#> # A tibble: 2 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 main_now     1.15ms   1.21ms      801.     190KB     4.20   381     2
#> 2 refactored 484.98µs 505.18µs     1893.     395KB    15.6    851     7
#> # ℹ 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#> #   time <list>, gc <list>
```

src/python.cpp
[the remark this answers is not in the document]
t-kalinowski: Actually, it's probably better to switch to using `Rcpp_fast_eval()` - slightly slower but safer (since fast_eval uses unwind protect by default).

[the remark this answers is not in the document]
kevinushey: Good idea!

src/python.cpp
kevinushey: I don't think `Rcpp::wrap()` makes any guarantees about whether or not the object it returns is protected; in general, if you're receiving an object from `Rcpp::wrap()`, you should protect it to be safe.

kevinushey:
> Do you have any specific edge-cases in mind? I already expanded some tests related to conversion, and will still add a few more related to logical numpy array conversion. Anything else?

Nothing specifically, just general trepidation resulting from seeing a large PR :-)

Do we already have test coverage for confirming that the removal of these S3 methods doesn't affect anything here? https://github.com/rstudio/reticulate/pull/1552/files#diff-b6b6854a02ae177f8860f49654ff250c1554d021ae434b6efdbe99d00bdc6055L7

src/python.cpp
t-kalinowski: Good catch! I opted to not use `Rf_copyMostAttrib()` for backcompat. The previous R implementaiton did this (which would copy `dimnames` `dim` `names` too):
```r
attributes(x2) <- attributes(x)
```

PR #1556, opened 2024-03-23
prefer 'agg' for non-interactive R

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13840.

@cderv does this look like the right fix to you?

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -510,6 +510,10 @@ eng_python_matplotlib_backend <- function() {
     return("agg")
 
   # prefer using the agg backend in non-interactive environments
+  # (matplotlib might prefer using the tkAgg backend, but we've seen
+  # issues when trying to use in some environments, e.g. Windows)
+  #
+  # https://github.com/rstudio/rstudio/issues/13840
   if (!interactive())
     return("agg")
 
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -492,6 +492,31 @@ eng_python_initialize_hooks <- function(options, envir) {
 
 }
 
+eng_python_matplotlib_backend <- function() {
+
+  # allow override, just in case
+  envvars <- c("RETICULATE_MPLBACKEND", "MPLBACKEND")
+  for (envvar in envvars) {
+    override <- Sys.getenv(envvar, unset = NA)
+    if (!is.na(override))
+      return(override)
+  }
+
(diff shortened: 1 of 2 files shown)


R/knitr-engine.R
```diff
@@ -492,19 +492,44 @@ eng_python_initialize_hooks <- function(options, envir) {
...
+  for (envvar in envvars) {
+    override <- Sys.getenv(envvar, unset = NA)
+    if (!is.na(override))
+      return(override)
+  }
+
+  # in RStudio Desktop, enforce a non-Qt matplotlib backend
+  # this is especially important with RStudio Desktop as attempting to use a Qt
+  # backend will cause issues due to mismatched Qt versions between RStudio and
+  # Anaconda environments, and will cause crashes when generating plots
```
t-kalinowski, line 508: Is this a conda specific workaround? Would a venv python installation find the same Qt version as the IDE? If not, should the comment be updated?

R/knitr-engine.R
```diff
@@ -492,19 +492,44 @@ eng_python_initialize_hooks <- function(options, envir) {
...
   # early exit if we already initialized
   # (this onload hook is registered for multiple matplotlib submodules)
   if (identical(.globals$matplotlib_initialized, TRUE))
     return(TRUE)
 
-  # attempt to enforce a non-Qt matplotlib backend. this is especially important
-  # with RStudio Desktop as attempting to use a Qt backend will cause issues due
-  # to mismatched Qt versions between RStudio and Anaconda environments, and
-  # will cause crashes when attempting to generate plots
-  testthat <- Sys.getenv("TESTTHAT", unset = NA)
```
t-kalinowski, line 506: Should we keep the check for if we're in a testthat context? I think we can run tests interactively too.

t-kalinowski (approved):
LGTM overall!

R/knitr-engine.R
kevinushey: Thanks for catching that -- I've brought that back.

R/knitr-engine.R
kevinushey: IIRC, this branch was specifically required for older versions of RStudio (which were built against Qt); I've updated the comment to indicate that and also note that we could probably remove this in a future `reticulate` release. Does that seem sensible?

cderv:
Sorry I missed this on Monday. Looks good thanks !

PR #1585, opened 2024-04-12
Allow users to pass extra command-line arguments for conda create, install.

### LINKED ISSUE #1583: Add optional `--override-channels` to all `conda_*` functions that accept `channel=`
When setting up environments via [**basilisk**](https://github.com/Bioconductor/basilisk), I would like to force `conda_create()` and `conda_install()` to only use the channels specified in `channel=`, without any interference from the user's `.condarc`. Some digging suggests that this can be done by setting the `--override-channels` flag in the `conda install/create` calls:

```
  --override-channels   Do not search default or .condarc channels. Requires --channel.
```

Can we have an option in `conda_create()` and `conda_install()` to add this flag in the corresponding system calls to `conda`? This is important to ensure compliance with Anaconda's ToS in a commercial environment, otherwise a user might inadvertently try to pull down packages from the non-free default channels if they have an old `.condarc` file lying around.

I can put together a PR if this is of interest. Seems simple enough, e.g., `restrict_channel = TRUE` would enable the flag.

LTLA:
Fixes #1583 with additional optional arguments. The same general approach could possibly could be extended to all `conda` calls (e.g., `conda clone`, `conda export`), but I've kept the scope limited to `create` and `install` for now.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# reticulate 1.36.0.9000
+
+- Added an option for extra command-line arguments in `conda_create()`,
+  `conda_install()` (#1585).
+
 # reticulate 1.36.0
 
 - Internal refactoring and optimization now give a faster experience,
--- a/R/conda.R
+++ b/R/conda.R
@@ -61,6 +61,12 @@
 #'
 #' @param all Boolean; report all instances of Python found?
 #'
+#' @param additional_create_arguments An optional character vector of
+#'   additional arguments to use in the call to `conda create`. 
+#'
+#' @param additional_install_arguments An optional character vector of
+#'   additional arguments to use in the call to `conda install`. 
+#'
 #' @param ... Optional arguments, reserved for future expansion.
 #'
 #'
@@ -195,7 +201,8 @@ conda_create <- function(envname = NULL,
                          channel = character(),
                          environment = NULL,
                          conda = "auto",
-                         python_version = miniconda_python_version())
+                         python_version = miniconda_python_version(),
+                         additional_create_args = character())
 {
   check_forbidden_install("Conda Environments")
 
@@ -205,7 +212,7 @@ conda_create <- function(envname = NULL,
 
   # if environment is provided, use it directly
   if (!is.null(environment))
-    return(conda_create_env(envname, environment, conda))
+    return(conda_create_env(envname, environment, conda, additional_create_args))
 
   # resolve environment name
   envname <- condaenv_resolve(envname)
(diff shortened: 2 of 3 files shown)


LTLA:
FYI this section of code in `conda_install()` doesn't respect the `channel=`, or even `forge=` arguments:

https://github.com/rstudio/reticulate/blob/2e73fb9757e1a207dd257be871413b810077a584/R/conda.R#L458-L465

So when testing, even though my `--override-channels` is passed in correctly, the `conda install` command fails with:

```
...
+ /home/luna/.cache/R/basilisk/1.15.5/0/bin/conda install --yes --prefix /tmp/RtmpAN86aL/my_package_A 'python=3.10.14' --override-channels

ArgumentError: At least one -c / --channel flag must be supplied when using --override-channels.

Error: installation of 'python=3.10.14' into environment '/tmp/RtmpAN86aL/my_package_A' failed [error code 2]
```

Is this intentional?

t-kalinowski:
> Is this intentional?

I don't think so. Or perhaps, it was at the time it was first written, but we can revise it. 

Can you give an example of optional channels that provide the `python` binary? I've typically only seen `python` pulled from the default channel, and then packages pulled from optional channels.

LTLA:
> Can you give an example of optional channels that provide the `python` binary? I've typically only seen `python` pulled from the default channel, and then packages pulled from optional channels.

`conda-forge` seems to do so. Adding `-c conda-forge` to the arguments above allows `--override-channels` to work.

t-kalinowski:
@LTLA Can you please add a NEWS entry?

t-kalinowski:
Taking a second look at this, should the arg simply be called `additional_args`, with `install|create` already implied by the name of the invoking function?

LTLA:
> Taking a second look at this, should the arg simply be called `additional_args`, with `install|create` already implied by the name of the invoking function?

I thought about that, but the problem is with `conda_install()`, which calls `conda_create()` behind the scenes. In that case, I figured it would be clearest if the `additional_*_args` were explicitly disambiguated in `conda_install()`'s signature. Then I just used the same format for the argument name in `conda_create()`, just for consistency.

I suppose it also provides some future-proofing in case, e.g., `pip` needs some additional arguments.

Anyway, I'm not particularly tied to this name, so just (tell me to) change it if you have something you prefer.

t-kalinowski:
Thank you for the PR!

t-kalinowski:
Yes please! FYI, I'm planning to submit to CRAN on Monday.

PR #1586, opened 2024-04-15
Improve `python_dependencies` vignette

salim-b:
- Fix `DESCRIPTION` field name, i.e. remove the `/autoconfigure` from `Config/reticulate/autoconfigure` as the code doesn't know that suffix, cf. https://github.com/rstudio/reticulate/blob/5774c590c8658da133f53dcc1b2474e276b49748/R/python-packages.R#L231

- Improve language and formatting and fix typos.

Note that another round of editing wouldn't hurt (but I don't have the motivation for this right now):
- The text is still very redundant, i.e. it repeats itself unnecessarily often.
- I think the information about default usage of Miniconda is not accurate anymore.

### DIFF
--- a/vignettes/python_dependencies.Rmd
+++ b/vignettes/python_dependencies.Rmd
@@ -12,12 +12,12 @@ editor_options:
 
 If you're writing an R package that uses `reticulate` as an interface to
 a Python session, you likely also need one or more Python packages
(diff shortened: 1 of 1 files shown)


t-kalinowski:
Thank you! Edits are much appreciated.

PR #1594, opened 2024-04-19
Respect user-specified channels when installing Python in conda_install().

LTLA:
As discussed in #1585, this ensures that Python installation inside `conda_install()` respects the user-defined channels.


t-kalinowski:
Looks great, thank you! Can you please add a NEWS entry?

LTLA:
Done.

t-kalinowski:
Thank you!

PR #1600, opened 2024-04-24
Do not add outputs when results = 'hide' is set

### LINKED ISSUE #1599: Support `results = 'hide'` directly in python engine
Reported in 
* https://github.com/quarto-dev/quarto-cli/issues/9389 

and full example at 
* https://github.com/posit-dev/great-tables/issues/291

## Problem 

Currently, `results = 'hide'` does not seem to be handled by **reticulate**'s knitr engine. Example: 

````markdown
---
title: "Test report"
output: html_document
---

```{r, results = 'hide'}
class MyClass:
    def _repr_html_(self):
        return "<p>uh-oh</p>"

MyClass()
```
````

The results will still be shown. This is a minimal example - see https://github.com/posit-dev/great-tables/issues/291 for one with **great_tables**

## Context

This is because `_repr_html_` is catched to be returned as a `knitr::raw_html()`
https://github.com/rstudio/reticulate/blob/d4e752bf4e6f3970465f327581de1a7185624a1e/R/knitr-engine.R#L744-L750
and provided in the `out` component on which `knitr:::sew()` will apply
https://github.com/rstudio/reticulate/blob/d4e752bf4e6f3970465f327581de1a7185624a1e/R/knitr-engine.R#L626-L628

By returning `engine_output(out = )`, knitr's assumes the output is already filtered with what should be filtered.
Otherwise, it will be handled directly `engine_output()` default usage
https://github.com/yihui/knitr/blob/ba8d9fb0ecd60f3f9ebec58fdde1e689bdc40ad3/R/engine.R#L96-L98

So I do think **reticulate** needs to support `results = 'hide'` and not provide outputs in the list when the option is provided. 

I can make a PR - just need to think about the best place to do this.

### PR DESCRIPTION
**Pull request by cderv**
closes #1599 

As mentioned, `include` is handled by **knitr** directly since a long time 
https://github.com/yihui/knitr/blob/44a7bee566afff3eb4c8b13fd6fa64487e12981f/R/block.R#L350

so I replaced by `results = 'hide'` information. 

It also seems that there are two places where output is handled (because of specific handling for matplotlib, so I added the if check there too 

I think this is all that is missing for results to be supported. 

If you prefer to keep `is_include` for safety, then we can add it back in each `if`

### DIFF
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -349,10 +349,11 @@ eng_python <- function(options) {
       plt$show()
   }
 
-  for (plot in .engine_context$pending_plots$data())
-    outputs_target$push(plot)
-  .engine_context$pending_plots$clear()
-
+  if (!is_hidden && is_include) {
+    for (plot in .engine_context$pending_plots$data())
+      outputs_target$push(plot)
+    .engine_context$pending_plots$clear()
+  }
 
   # if we were using held outputs, we just inject the source in now
   if (is_hold) {
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -197,6 +197,7 @@ eng_python <- function(options) {
 
   # Stash some options.
   is_hold <- identical(options$results, "hold")
+  is_hidden <- identical(options$results, "hide")
   is_include <- isTRUE(options$include)
   jupyter_compat <- isTRUE(options$jupyter_compat)
 
@@ -306,7 +307,7 @@ eng_python <- function(options) {
       }
 
       # append captured outputs (respecting 'include' option)
-      if (is_include) {
+      if (!is_hidden && is_include) {
         # append captured output
         if (!identical(captured, ""))
           outputs_target$push(captured)
--- a/R/knitr-engine.R
+++ b/R/knitr-engine.R
@@ -198,7 +198,6 @@ eng_python <- function(options) {
   # Stash some options.
   is_hold <- identical(options$results, "hold")
   is_hidden <- identical(options$results, "hide")
-  is_include <- isTRUE(options$include)
   jupyter_compat <- isTRUE(options$jupyter_compat)
 
   # line index from which source should be emitted
@@ -306,8 +305,8 @@ eng_python <- function(options) {
         outputs$push(output)
       }
 
-      # append captured outputs (respecting 'include' option)
-      if (!is_hidden && is_include) {
+      # append captured outputs (respecting 'results = "hide"' option)
+      if (!is_hidden) {
         # append captured output
         if (!identical(captured, ""))
           outputs_target$push(captured)
@@ -349,7 +348,7 @@ eng_python <- function(options) {
       plt$show()
   }
 
-  if (!is_hidden && is_include) {
+  if (!is_hidden) {
     for (plot in .engine_context$pending_plots$data())
       outputs_target$push(plot)
     .engine_context$pending_plots$clear()
--- a/tests/testthat/_snaps/python-knitr-engine/knitr-results-hide.md
+++ b/tests/testthat/_snaps/python-knitr-engine/knitr-results-hide.md
@@ -1,0 +1,5 @@
+    class MyClass:
+        def _repr_html_(self):
+            return "<p>uh-oh</p>"
+
+    MyClass()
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
Let's make it a draft for now, because I want to check something regarding `results = 'hide'` and plot output in **knitr** with R

**Comment by cderv** [comment]
Ok so for R, there is a difference between plots and text outputs
https://github.com/yihui/knitr/blob/44a7bee566afff3eb4c8b13fd6fa64487e12981f/R/block.R#L292

`results = 'hide'` only really applies to evaluated output in R where `is.character()` is TRUE. So it does not applies to plot. 
`````
```{r, results="hide"}
library(ggplot2)

ggplot(mtcars, aes(cyl, mpg)) + geom_point()
```
````` 

So it may requires some similar logic for filtering output in the reticulate engine 🤔 

but honestly, I think this could be ok to ignore this (at least for now). I don't see how `results = "hide"` set on chunk would have an expectation of having plots still show up.

t-kalinowski:
Thanks for the PR!

> I don't see how results = "hide" set on chunk would have an expectation of having plots still show up.

Looking at the docs here: https://yihui.org/knitr/options/

> `results`: ('markup'; character) Controls how to display the text results. Note that this option only applies to normal text output (not warnings, messages, or errors). 
> ...
> `hide` (or `FALSE`): Hide text output.

It is pretty clear that `results = 'hide'` is meant to only suppresses text output.

I can imagine that some users use that chunk option to suppress unwanted text output, while still displaying plots.

I find it interesting that evaluating a `ggplot()` object draws to the graphics device still---I think this means that `print()` is called, since merely calling `ggplot()` will not draw to the graphics device. Does this mean that the chunk is evaluated "normally", and then the text output is discarded later, after the chunk is fully evaluated? 

Should we take a similar approach here?

(Should this options be handled in knitr, like `include = FALSE` is?)

cderv:
> I can imagine that some users use that chunk option to suppress unwanted text output, while still displaying plots.

That would be the scenario - I agree. 

> Does this mean that the chunk is evaluated "normally", and then the text output is discarded later, after the chunk is fully evaluated?

`knit_print()` is called as part of the evaluation, and the filtering for text output happens after the chunk evaluation. 
For ggplot, `knit_print` will be a `print`. 

> (Should these options be handled in knitr, like include = FALSE is?)

We do it globally through `sew.character()` method. This means that if an engine like reticulate returns an object of class `character`, we do apply `results = "hide"`. The initial problem here is that in some case, **reticulate** is wrapping some output in `output_asis()` function, making it  a `knit_asis` object for which `sew.knit_asis` will apply. 

Writing this, I am now seeing this is different behavior than in R engine where we filter output any output for which  `is.character` is TRUE (which include a string of class `knit_asis`). 

Maybe this difference is ok. My understanding is that anything that comes from an external non-R engine will be considered by **knitr** as text output. There is no non-text object handled in this case. So `results = 'hide'` would hide everything if we do the same. That is why currently it is handled by the class with `sew()` method like the rest. 
If this option is handled at the engine level, what is considered as text output and will be hidden is the choice of the engine. 

> Should we take a similar approach here?

What approach do you have in mind related to graphic device ?

t-kalinowski:
> What approach do you have in mind related to graphic device ?

I'm not sure. Python plots from `reticulate::eng_python()` don't go through the R graphics device, but it would be nice if we could match the behavior of R chunks that do. We currently output Python plots using `knitr::include_graphics()`.

The question I suppose we should resolve: is html or image chunk output considered "text", when determining if the `results = 'hide'` chunk option applies. 

Consider: what if a user changes a global option in their document, which changes the output emitted by a chunk. Output can be either raw html, an svg, an image, or a plot on the R graphics device. But regardless of how it is emitted, the representation in the final rendered document is roughly the same. Does the `results = 'hide'` option causes the output to be hidden for some of the cases but not others?

I am open to what you think is best. 

Based on the knitr chunk option documentation, my understanding would be that `results = 'hide'` suppresses "text", in the sense that it only suppresses output that would be otherwise presented in the final rendered document as plain, monospaced, chunk output. Any chunk output that would ultimately be seen by the user's eyeballs in the final report as a "richer" representation is not suppressed. E.g, plots and fancy tables are not suppressed, regardless of if they are included via the R graphics device, or an included image file, or an svg, or as raw html. But then, I would expect such a chunk option to be called `output` ... 🤔.

I think the best we can aim for is consistency with the behavior of the R engine. If I'm understanding correctly, the knitr R engine with chunk option `results = 'hide'` does not hide plots or non-ascii tables. We should do the same here, no?

PR #1608, opened 2024-05-13
Support non-conda binaries

AlexandreGuinaudeau:
Allow using a `mamba` binary instead of `conda`.
Currently, the `mamba` binary gets overriden with `conda`, even if using the version of `mamba` in the `bin` directory. We preserve the behavior of overriding `condabin` with `bin` if an executable exists in that folder

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 - `virtualenv_starter()` no longer warns when encountering broken symlinks (#1598).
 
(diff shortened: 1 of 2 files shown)


t-kalinowski:
Hi, thanks for the PR. This seems fine. Can you please add a NEWS entry?

AlexandreGuinaudeau:
Sure, just added one

t-kalinowski:
Thank you!

PR #1609, opened 2024-05-16
Include signals.h on Windows ARM64

andrjohns:
Attempting to install `reticulate` on Windows ARM64 fails with the error:

```
* installing *source* package 'reticulate' ...
** using staged installation
** libs
using C++ compiler: 'clang version 17.0.6'
clang++ -std=gnu++17  -I"C:/PROGRA~1/R-AARC~1/R-devel/include" -DNDEBUG  -I'C:/Users/andrew/AppData/Local/R/aarch64-library/4.5/Rcpp/include' -w  -I"C:/rtools44-aarch64/aarch64-w64-mingw32.static.posix/include"     -O2 -Wall    -c RcppExports.cpp -o RcppExports.o
clang++ -std=gnu++17  -I"C:/PROGRA~1/R-AARC~1/R-devel/include" -DNDEBUG  -I'C:/Users/andrew/AppData/Local/R/aarch64-library/4.5/Rcpp/include' -w  -I"C:/rtools44-aarch64/aarch64-w64-mingw32.static.posix/include"     -O2 -Wall    -c event_loop.cpp -o event_loop.o
event_loop.cpp:40:10: error: unknown type name 'sig_atomic_t'
   40 | volatile sig_atomic_t s_pollingRequested;
```

It appears that the `signals.h` header is additionally required for Windows ARM64

### DIFF
--- a/src/tinythread.h
+++ b/src/tinythread.h
@@ -84,6 +84,9 @@ extern "C" void Rf_error(const char* fmt, ...);
     #undef WIN32_LEAN_AND_MEAN
     #undef __UNDEF_LEAN_AND_MEAN
   #endif
+  #ifdef __aarch64__
+    #include <csignal>
+  #endif
 #else
   #include <pthread.h>
   #include <signal.h>
--- a/src/tinythread.h
+++ b/src/tinythread.h
@@ -85,7 +85,7 @@ extern "C" void Rf_error(const char* fmt, ...);
     #undef __UNDEF_LEAN_AND_MEAN
   #endif
   #ifdef __aarch64__
-    #include <csignal>
+    #include <signal.h>
   #endif
 #else
   #include <pthread.h>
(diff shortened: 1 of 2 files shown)


andrjohns:
The current check failures appear unrelated to this PR, and I think are caused by [this `printf` call](https://github.com/rstudio/reticulate/blob/7ed6ab0a8a0607e8e1f8cb88aa54805d55be9f9c/src/libpython.cpp#L165)

src/tinythread.h
```diff
@@ -84,6 +84,9 @@ extern "C" void Rf_error(const char* fmt, ...);
     #undef WIN32_LEAN_AND_MEAN
     #undef __UNDEF_LEAN_AND_MEAN
   #endif
+  #ifdef __aarch64__
+    #include <csignal>
```
kevinushey, line 88:
```suggestion
    #include <signal.h>
```

PR #1641, opened 2024-08-04
Allow Python backgrounds threads to run in parallel with R

t-kalinowski:
This PR allows spawned Python background threads to run while the main thread is executing R code.  

For example, if you start a Python background thread like this: 
```r
library(reticulate)

py_config()

py_run_string(r"---(
import _thread
from time import sleep, localtime, strftime

def be_chatty():
  for i in range(100000):
    print('Hi', strftime('%H:%M:%S', localtime()))
    sleep(1)

_thread.start_new_thread(be_chatty, ())

)---")
```

With the current reticulate release, we don't see any output from the Python thread while at the R repl. The Python thread only is active while we are in a Python function that gives an opportunity for the GIL to be released (e.g, an explicit sleep, or evaluating Python bytecode). E.g., we only see output from the above background thread if we call in R:
```r
py_sleep <- reticulate::import("time")$sleep
py_sleep(5)
```

However, with this PR, we now see output from the Python background thread even when not calling into Python from the main thread.

The primary motivation is to allow a Positron Python console to be backed by the same Python session reticulate in the Positron R console is using, without needing to block the R console, or spin-up+shutdown the ipykernel event loop on each call to `repl_python()`.


tests/testthat/test-python-threads.R
```diff
@@ -0,0 +1,40 @@
...
+
+  file <- tempfile()
+  on.exit(unlink(file), add = TRUE)
+
+  write_to_file_from_thread <- py_run_string("
+def write_to_file_from_thread(path, lines):
+    from time import sleep, localtime, strftime
+
+    def write_to_file(path, lines):
+        sleep(.1) # don't try to run until we've had a chance to return to the R main thread
```
kevinushey, line 13: Just to confirm, this is just a way to make sure that this thread pauses execution until we later try to invoke something in Python? (In general, users shouldn't need to use `sleep()` to ensure proper sequencing here?)

kevinushey (approved):
LGTM!

tests/testthat/test-python-threads.R
t-kalinowski: This sleep is there to make sure that we've give a chance for the main thread to exit back into R from the python call. Otherwise, it's possible that the background thread can steal the GIL while the main thread is still evaluating Python code, which would lead to a failing test.

Issue #1642, opened 2024-08-05
Default Altair chart dimensions are not respected


joelostblom:
Originally reported for quarto in https://github.com/quarto-dev/quarto-cli/issues/10459, but was told this is probably an issue in reticulate or knitr instread so reposting here

### Bug description

When I create a Altair charts, the default dimensions of the charts are respected when rendered form quarto documents containing only Python code:

![image](https://github.com/user-attachments/assets/996fe736-7e5c-432e-9d84-a9fb13631980)

However, when I mix and match R and Python, the dimensions of the chart are no longer respected and the chart area is stretched so that the chart becomes much too large:

![image](https://github.com/user-attachments/assets/38e0d58d-9d45-494e-93ab-41d046e0035d)

Setting the dimensions explicitly works as expected:

![image](https://github.com/user-attachments/assets/eb096540-0407-45dc-8b20-c4f5be94baf1)

Is this an issue with reticulate? Would it be possible to respect the default dimensions of altair charts just as the dimensions are respected when the width and height are set explicitly?

### Steps to reproduce

Run `quarto preview name.qmd` with the following doc:

````md
---
format: html
---

Python code:

```{python}
import altair as alt

alt.Chart().mark_point()
```

R code:

```{r}
4 + 4
```
````

joelostblom:
I think these are the offending lines https://github.com/rstudio/reticulate/blob/d683b3bc6311c5ebee3504901f8201edb8319341/R/knitr-engine.R#L770-L786

Instead of explicitly setting the chart dimensions in reticulate, it would be great if Altair's default conventions for determining the figure width and size would be used. By default Altair uses a width and height of 300 px for continuous axes and for discrete axes 20 px per category along the axis is used. Currently these conventions are overwritten by reticulate and it leads to the chart appearance not being what is expected. I believe that the current approach would also not respect if a global theme was set in altair that specifies a different width/height (at least for continuous axis since this is how the current altair defaults are set and these are not detected).

[...]

PR #1646, opened 2024-08-13
fix: Do not change the default dimensions of altair charts unless they are set with the altair specific options

joelostblom:
close https://github.com/rstudio/reticulate/issues/1642

This will avoid using the knitr default dimensions from other figure formats for altair charts. I think this makes sense since altair already has sensible and easily overidable defaults. The altair-specific kntir options are left as before and can still be used (I think it would also be an option to scrap these because as noted in the source comments, they don't work for compound charts whereas the settings in altair does).

The behavior this PR introduces is consistent with how altair works elsewhere, such as in web document and IDEs (I could not test in RStudio due to altair charts not being supported there yet as per https://github.com/rstudio/rstudio/issues/8762 and https://github.com/rstudio/rstudio/issues/4259). I believe this would be more intuitive for users than having the default altair dimensions overwritten by knitr (e.g. altair uses adaptive dimensions for categorical data depending on the number of categories and a consistent mark size rather than scaling the marks based on the number of categories, see example with the bar charts below).

Using the following quarto/rmarkdown source doc for testing purposes:

````{md}
---
format: html
engine: knitr
---

```{python}
import altair as alt

alt.Chart().mark_point()
```

```{python}
import pandas as pd
source = pd.DataFrame({
    'a': ['A', 'B', 'C'],
    'b': [28, 55, 43]
})

alt.Chart(source).mark_point().encode(
    x='b',
)
```

```{python}

source = pd.DataFrame({
    'a': ['A', 'B', 'C'],
    'b': [28, 55, 43]
})

alt.Chart(source).mark_bar().encode(
    x='b',
    y='a'
)
```

```{python, altair.fig.height=200, altair.fig.width=200}

source = pd.DataFrame({
    'a': ['A', 'B', 'C'],
    'b': [28, 55, 43]
})

alt.Chart(source).mark_bar().encode(
    x='b',
    y='a'
)
```
````

## Quarto output BEFORE this PR

image
 
## Quarto output AFTER this PR

image

## RMarkdown output AFTER this PR

The behavior before and after is the same as for quarto so just showing one screenshot here indicating that things are working as expected.

image

### CODE REVIEW

File: R/knitr-engine.R [hunk 1]
```diff
@@ -767,22 +767,18 @@ eng_python_autoprint <- function(captured, options) {
 
   } else if (eng_python_is_altair_chart(value)) {
 
-    # set width if it's not already set
+    # set width and height if it's not already set
     # This only applies to Chart objects, compound charts like HConcatChart
     # don't have a 'width' or 'height' property attribute.
     # TODO: add support for propagating width/height options from knitr to
     # altair compound charts
     width <- py_get_attr(value, "width", TRUE)
```
joelostblom, line 775: I am not sure if this line (and the corresponding one for `height`) are still needed, but I left them in for now.

t-kalinowski:
Thank you for the PR! This seems good to me, but I'm a bit removed from altair charts and don't know how impactful or breaking this change will be for users who are currently expecting the default knitr chunk option to be respected. 

@cderv can you weight in? I'm inclined to merge.

t-kalinowski:
I'll go ahead and merge. @joelostblom can you please add a NEWs entry?

joelostblom:
Added, thank you @t-kalinowski !

NEWS.md
```diff
@@ -11,11 +11,18 @@
...
+  Instead, use `altair.fig.height`, `altair.fig.width`,
+  or Altair's `width` and `height` parameters
+  to adjust chart dimensions.
+
 - New `as.raw()` method for `python.builtin.bytes` (#1649, #1652)
 
 - `as.character()` method for `python.builtin.bytes` gains a `nul` argument,
   allowing for convenient handling of embedded NULs in the string. (#1652)
 
+
```
joelostblom, line 25:
```suggestion
```

t-kalinowski:
Thank you!

PR #1648, opened 2024-08-16
Update `repl_python()` to use independent Console session in Positron

t-kalinowski:
This PR works towards making `repl_python()` able to use Positron Python features and making `repl_python()` a peer experience to using the Python language pack in Positron.

The updates primarily involve under-the-hood changes to support this goal. These include:

- A "Pending Python Calls" notifier integrated with R's main thread event loop.
- Automatic finalization of the Python session when the R session exits.
- Support for calling R functions from background Python threads, making `py_main_thread_func()` obsolete and deprecated.

This work continues from #1641, which allowed background Python threads to run concurrently with R on the main thread.


src/pending_py_calls_notifier.h
```diff
@@ -0,0 +1,19 @@
...
+namespace pending_py_calls_notifier {
+
+    // Initialize the notifier with a function that runs pending calls.
+    void initialize(std::function<void()> run_pending_calls);
+
+    // Notify the main thread to run pending calls.
+    void notify();
+
+    // Undo initialize
+    void deinitialize();
```
kevinushey, line 15: Where does this get called?

src/python.cpp
```diff
@@ -2291,9 +2300,26 @@ class AllowPyThreadsScope
 
 // custom module used for calling R functions from python wrappers
 
-extern "C" PyObject* call_r_function(PyObject *self, PyObject* args, PyObject* keywords)
-{
-  // the first argument is always the capsule containing the R function to call
+
+void safe_print_value(SEXP value, const char* warn_print_failed) {
+    // Use R_ToplevelExec to safely execute Rf_PrintValue with an inlined lambda
+    if (!R_ToplevelExec([](void* data) -> void {
```
kevinushey, line 2306: 🧙

kevinushey (commented):
Overall LGTM!

src/pending_py_calls_notifier.cpp
```diff
@@ -0,0 +1,131 @@
...
+    Rf_error("Failed to create pipe for pending Python calls notifier");
+
+  input_handler = addInputHandler(R_InputHandlers, pipe_fds[0], input_handler_function, 88);
+}
+
+void notify() {
+  if (!notification_pending.exchange(true)) {
+    if (write(pipe_fds[1], "x", 1) == -1) {
+      // Called from background threads, can't throw R error.
+      REprintf("Failed to write to pipe for pending Python calls notifier\n");
```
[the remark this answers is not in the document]
t-kalinowski: Ideally we could use `printf()`, which already uses a lock for `stdout`, but CRAN checks fail if I use `printf()` 

```
N  checking compiled code ...
   File ‘reticulate/libs/reticulate.so’:
     Found ‘_printf’, possibly from ‘printf’ (C)
       Object: ‘pending_py_calls_notifier.o’
   
   Compiled code should not call entry points which might terminate R nor
   write to stdout/stderr instead of to the console, nor use Fortran I/O
   nor system RNGs nor [v]sprintf.
   
   See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual.
```

This is going to be such a rare error I'd be surprised if users ever even see it... It seems not worth the complexity to add yet another channel just to communicate if the first channel failed. 

Any other ideas besides `REprintf()`?

src/pending_py_calls_notifier.h
t-kalinowski: nowhere yet, still need to plumb it through to package unload.

t-kalinowski: Is there a counterpart to `// [[Rcpp::init]]` but for finalize/unload?

src/pending_py_calls_notifier.cpp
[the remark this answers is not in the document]
t-kalinowski: I'm starting to think that using `REprintf()` here and risking a segfault is perhaps the least worst option. If writing to a pipe failed, then something else is deeply wrong and this is the least of the processes problems 😓.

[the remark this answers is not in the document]
t-kalinowski: Actually, WRE has:

> Another circumstance when it may be important to use these [`R*printf`] functions is when using parallel computation on a cluster of computational nodes, as their output will be re-directed/logged appropriately.

This suggests that they _are_ threadsafe. https://rstudio.github.io/r-manuals/r-exts/The-R-API.html#printing

src/pending_py_calls_notifier.h
t-kalinowski: It is now called from `py_finalize()`, along with a bunch of other new finalizers.
Also, we now register `py_finalize()` on an object in the namespace, so the Python session is finalized on R session exit.

Issue #1654, opened 2024-08-24
Invalid conda-meta/history path prioritized over configured conda binary


tl-hbk:
When calling `use_python` or `use_condaenv` even if you provide `conda=<path to binary>` the following code path still gets hit

```
use_python -> python_config -> python_munge_path -> get_python_conda_info -> python_info_condaenv_find
```

Some environments I have happened to be created using `python -m conda` which leads to the following being in the history file

```
# cmd: /tmp/conda/lib/python3.8/site-packages/conda/__main__.py create --yes --prefix /tmp/envs/foo --file /tmp/spec-file.txt
```

later calls to `conda_binary` use this bad value instead of whatever was provided to the original `conda` arg.  Could `get_python_conda_info` use `find_conda` first instead of looking through the history file?  That way the option `reticulate.conda_binary` or the environment variable `RETICULATE_CONDA` can be used even when the history file has a bad value.

```
    conda <- find_conda()
    if (is.null(conda)) {
        # not base env, parse conda-meta history to find the conda binary
        # that created it
        conda <- python_info_condaenv_find(root)
    }
```

t-kalinowski:
This sounds reasonable. Would you like to submit a PR?

tl-hbk:
@t-kalinowski I've opened the PR

PR #1659, opened 2024-09-04
Handle invalid conda-meta/history value

tl-hbk:
Resolves https://github.com/rstudio/reticulate/issues/1654

`python_info_condaenv_find` was only being used by `get_python_conda_info` and `python_info_condaenv` so I combined it with `get_python_conda_info` and just repointed `python_info_condaenv` at `get_python_conda_info` so it also gets the fix.

### DIFF
--- a/R/conda.R
+++ b/R/conda.R
@@ -1106,9 +1106,16 @@ get_python_conda_info <- function(python) {
     else
       file.path(root, "bin/conda")
   } else {
-    # not base env, parse conda-meta history to find the conda binary
-    # that created it
-    conda <- python_info_condaenv_find(root)
+    # not base env
+    # in ArcGIS env, conda.exe lives in a parent directory
+    exe <- file.path(root, "../..", "Scripts", "conda.exe")
+    exe <- normalizePath(exe, winslash = "/", mustWork = FALSE)
+    if (is_windows() && file.exists(exe)) {
+      conda <- exe
+    } else {
+      # parse conda-meta history to find the conda binary that created it
+      conda <- conda_history_find(root)
+    }
   }
 
   conda <- normalizePath(conda, winslash = "/", mustWork = FALSE)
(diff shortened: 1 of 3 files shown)


R/conda.R
```diff
@@ -1106,9 +1106,26 @@ get_python_conda_info <- function(python) {
...
-    # that created it
-    conda <- python_info_condaenv_find(root)
+    # not base env
+    conda <- find_conda()
+    if (is.null(conda)) {
+      if (is_windows()) {
+        # in Anaconda base env, conda.exe lives under Scripts
+        exe <- file.path(root, "Scripts", "conda.exe")
+        if (file.exists(exe))
+          conda <- exe
```
tl-hbk, line 1116: Not sure if this is still necessary with the above handling for base conda envs

```
  if (dir.exists(file.path(root, "condabin"))) {
    # base conda env
    conda <- if (is_windows())
      file.path(root, "condabin/conda.bat")
    else
      file.path(root, "bin/conda")
  } 
```

tl-hbk: https://github.com/rstudio/reticulate/blob/14d732716c006d72500be1db57388645c3800183/R/conda.R#L553-L560

`conda.bat` replaces `conda.exe` whenever `conda_binary(".../conda.exe")` is used too

R/conda.R
```diff
@@ -1106,9 +1106,26 @@ get_python_conda_info <- function(python) {
     else
       file.path(root, "bin/conda")
   } else {
-    # not base env, parse conda-meta history to find the conda binary
-    # that created it
-    conda <- python_info_condaenv_find(root)
+    # not base env
+    conda <- find_conda()
```
t-kalinowski, line 1110: Thank you for the PR! I noticed something in the implementation that might need clarification. 

The purpose of `get_python_conda_info()` is to locate the specific `conda` used to create a given `python` conda environment. Since it's common for users to have multiple versions of `conda` installed, this function focuses on identifying the correct one for the environment.

On the other hand, `find_conda()` is meant to find the user's preferred `conda` executable based on expressed preferences in `options()` or other heuristics.

It seems like using `find_conda()` within `get_python_conda_info()` may not be the right approach here. What do you think?

tl-hbk: I think my specific issue is when it hits 

```
  if(!file.exists(conda))
    conda <- NA
```

so I guess in other locations if `NA` is passed I can change it to use `find_conda` then

tl-hbk: Oh I guess `conda_binary` already handles that specific case here

https://github.com/rstudio/reticulate/blob/4277e2dfc4da7403028a05dc3db2d9591d22a212/R/conda.R#L531-L539

The only reason my error was occuring was because `get_python_conda_info` used `python_info_condaenv_find` directly which didn't handle invalid values from the history

tl-hbk:
Hi @t-kalinowski could you take a look at this again?  

I've changed it so that `get_python_conda_info` still only locates the conda specific to an environment but because of the following check

https://github.com/rstudio/reticulate/blob/4201011eb59387cfb1b0134dd8fc05b2e1da0075/R/conda.R#L1121-L1123

`python_info_condaenv` now gets `NA` instead of an invalid path in my edge case.

t-kalinowski:
Thank you for the PR!

PR #1670, opened 2024-09-16
Fix error when running under `radian`

### LINKED ISSUE #1668: reticulate 1.39 breaks embedding R in python environment.
I am the author of [radian](https://github.com/randy3k/radian), I was reported that radian doesn't work with  new version of reticulate, see https://github.com/randy3k/radian/issues/489.

After poking around, it seems that it is caused by this [line](https://github.com/rstudio/reticulate/blob/da89b640524df194b19a158a6cf8d6debfe8b099/src/python.cpp#L2737C3-L2737C18)
```
GILScope scope;
```
in `main_process_python_info_unix`. AFAICT, It is trying to initialize python by running `reticulate:::initialize_python()` 
but `initialize_python` calls `py_discover_config` which calls `main_process_python_info` and `main_process_python_info_unix`.

It creates a dead loop.

I tried commented out this line `// GILScope scope;`,  and it seems that it is now back in working, but not sure if GIL is needed in this function.

t-kalinowski:
closes #1668

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # reticulate (development version)
 
+- Fixed error when using reticulate with radian (#1668, #1670).
+
 - Fixed segfault encountered when running Python finalizer (#1663, #1664)
 
 - Fixed segfault encountered in RStudio when rapidly switching 
--- a/src/python.cpp
+++ b/src/python.cpp
@@ -99,18 +99,6 @@ bool is_interactive() {
   return s_isInteractive;
 }
 
-// a simplified version of loadSymbol adopted from libpython.cpp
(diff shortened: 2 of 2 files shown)


randy3k:
Tested on macOS, works perfectly. Thank you for the quick fix.
Though I wonder why you want to skip windows. radian was working perfectly fine on windows too.

t-kalinowski:
Thank you, @randy3k!

I believe I understand what was happening and have pushed another fix. With this branch, `radian` should now work on all platforms again, including Windows. Would you be able to run Radian CI tests against this branch to confirm?

randy3k:
Perfect, all tests pass https://github.com/randy3k/radian/pull/490.
Thank you for the quick response.

PR #1681, opened 2024-10-09
Positron REPL: restore `__main__` members after ipykernel launch

t-kalinowski:
Launching IPyKernel for the reticulate-backed Positron Python REPL currently clears Python's `__main__` namespace, including the `r` symbol injected for communication from Python to R. This pull request fixes that issue.

### DIFF
--- a/R/thread.R
+++ b/R/thread.R
@@ -53,7 +53,30 @@ py_allow_threads <- function(allow = TRUE) {
 py_run_file_on_thread <- function(file, ..., args = NULL) {
   if (!is.null(args))
     args <- as.list(as.character(args))
-  import("rpytools.run")$`_launch_lsp_server_on_thread`(file, args)
+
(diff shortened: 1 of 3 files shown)


R/thread.R
```diff
@@ -53,7 +53,30 @@ py_allow_threads <- function(allow = TRUE) {
...
+    py_get_attr(main_dict, "pop")("__annotations__")
+  }
+
+  import("rpytools.run")$run_file_on_thread(file, args)
+
+  if (launching_lsp) {
+
+    PositronIPKernelApp <- import("positron_ipykernel.positron_ipkernel")$PositronIPKernelApp
+    while(!PositronIPKernelApp$initialized())
+      Sys.sleep(.5)
```
dfalbel, line 74: Should we have a timeout here, so we can unblock the R session if something goes wrong and the kernel never starts?

PR #1692, opened 2024-11-11
Add Positron Variables Pane Methods

t-kalinowski:
This PR adds support for reticulate objects in the Positron Variables Pane. 

With an example session:

```r
# needed because positron leaks an internal VIRTUAL_ENV env var into the R process
Sys.unsetenv("VIRTUAL_ENV")

library(reticulate)
library(keras3)

anint <- py_eval("1", FALSE)
adict <- py_eval("{'a':1, 'b':2}", FALSE)
adict['foo'] <- adict # recursive

alist <- py_eval("['a', 1]", FALSE)
atuple <- py_eval("('a', 1)", FALSE)

input <- keras_input(c(24, 24))

output <- input |>
  layer_dense(32, "relu") |>
  layer_dense(10, "softmax")

model <- keras_model(input, output)
```
We now see:

https://github.com/user-attachments/assets/30e1d667-63ed-4435-a568-dc4bfff1b04e

Previously, all python objects presented as:
image

This uses the ark interface added in https://github.com/posit-dev/ark/pull/560 and https://github.com/posit-dev/ark/pull/561

### DIFF
--- a/R/ark-variables-methods.R
+++ b/R/ark-variables-methods.R
@@ -17,7 +17,7 @@ ark_positron_variable_display_type.python.builtin.object <- function(x, ..., inc
   i <- .globals$get_positron_variable_inspector(x)
   out <-  i$get_display_type()
 
-  if(startsWith(class(x)[1], "python.builtin."))
+  if (startsWith(class(x)[1], "python.builtin.")) # display convert value?
     out <- paste("python", out)
 
   out
@@ -44,17 +44,15 @@ ark_positron_variable_get_children.python.builtin.object <- function(x, ...) {
 
   get_keys_and_children <- .globals$ark_variable_get_keys_and_children
   if (is.null(get_keys_and_children)) {
-    get_keys_and_children <- .globals$ark_variable_get_keys_and_children <- py_eval(
-      "lambda i: zip(*((str(key), i.get_child(key)) for key in i.get_children()))",
-      convert = FALSE
-    )
+    get_keys_and_children <- .globals$ark_variable_get_keys_and_children <-
+      import("rpytools.ark_variables", convert = FALSE)$get_keys_and_children
   }
 
   keys_and_children <- iterate(get_keys_and_children(i), simplify = FALSE)
-  out <- iterate(keys_and_children[[2L]], simplify = FALSE)
-  names(out) <- as.character(py_to_r(keys_and_children[[1L]]))
+  children <- iterate(keys_and_children[[2L]], simplify = FALSE)
+  names(children) <- as.character(py_to_r(keys_and_children[[1L]]))
 
-  out
+  children
 }
 
 #' @param index An integer > 1, representing the index position of the child in the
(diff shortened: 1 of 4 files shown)


R/ark-variables-methods.R
```diff
@@ -0,0 +1,85 @@
...
+
+ark_positron_variable_get_children.python.builtin.object <- function(x, ...) {
+  # Return an R list of children. The order of children should be
+  # stable between repeated calls on the same object. For example:
+  i <- .globals$get_positron_variable_inspector(x)
+
+  get_keys_and_children <- .globals$ark_variable_get_keys_and_children
+  if (is.null(get_keys_and_children)) {
+    get_keys_and_children <- .globals$ark_variable_get_keys_and_children <- py_eval(
+      "lambda i: zip(*((str(key), i.get_child(key)) for key in i.get_children()))",
```
dfalbel, line 48: I think we should limit the the length of `i.get_children()`, either by providing an argument from ark to `ark_positron_variable_get_children` or hardcoded here.

dfalbel (approved):
LGTM!

R/zzz.R
```diff
@@ -61,9 +61,61 @@
...
 # .onUnload <- function(libpath) {
 #   py_finalize() # called from reg.finalizer(.globals) instead.
 # }
+
+
+import_positron_ipykernel_inspectors <- function() {
+  if(!is_positron())
+    return (NULL)
+
+  x <- Sys.getenv("_")
```
t-kalinowski, line 106: @dfalbel (@lionel-?) Is there a better way to get the path to the Positron installation from R? I'm not sure if this is the most reliable way to resolve the path to the `inspectors.py` file that ships with Positron.

R/ark-variables-methods.R
t-kalinowski: What do you think the limit should be? Would it be dynamic based on something? Maybe a range of which children are visible in the pane?

If we do something here, we should also include a companion PR to make sure that the Python `inspectors.get_children()` is in sync.

t-kalinowski: Do we have a standard way to indicate that not all elements are being shown? E.g., `...truncated` or similar?

dfalbel: The python inspector does something like:

image

t-kalinowski: It looks like this is handled in the frontend: https://github.com/posit-dev/positron/blob/83d516172b17d2ae02da8b01411de646cb1117c9/src/vs/workbench/contrib/positronVariables/browser/components/variableOverflow.tsx#L87-L91

R/zzz.R
t-kalinowski: @lionel- We'll open a PR to Ark, exporting a function like `.ps.positron_ipykernel_path`, to return the path to the `positron_ipykernel` python module that ships with Positron.

lionel-: yes that would be reasonable. `.ps` stands for positron so `.ps_ipykernel_path`?

This function could query the positron-r extension for that path via UIComm RPC.

PR #1718, opened 2025-01-25
Updates to `py_require()`

t-kalinowski:
WIP. Current changes:

- Update order of discovery:
  - continue to discover r-reticualte venv before bootstrapping
  - only bootstrap a venv if initializing python
  - give a way to opt out

- Update `uv_get_or_create_env()`:
  - pass through stderr from uv
  - simplify pre+post processing

- Update `uv_binary()`
  - allow resolving without bootstraping install

- Update uv cache resolution:
  - Fix issue where reticulate resolved the incorrect cache dir if a user-managed uv installation is used. Right now this is done by forcing to always use the reticulate cache, but it might make sense instead to let uv use it's default `uv cache dir` if the uv installation is not managed by reticulate.

- Changes to tests:
  - declare requirements for tests with `py_require()` 
  - capture stderr from uv in tests
  - use color in snapshots
  - skip tests around cache resolution that fail if using user-installed uv
  - skip tests of py_require() that fail if python is already initialized

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 - Fixed an issue with using Python 3.12 on Linux (#1712, #1714).
 
-- Fixed an issue where `virtualenv_starter()` would not discover a 
+- Fixed an issue where `virtualenv_starter()` would not discover a
   custom built Python (#1704).
 
 # reticulate 1.40.0
@@ -26,7 +26,7 @@
 - Fixed error when attempting to use a python venv created with `uv` (#1678)
 
 - Resolved an issue where `py_discover_config()` attempted to detect
-  Windows App Store Python installations. These are now excluded from 
+  Windows App Store Python installations. These are now excluded from
   discovery by both `py_discover_config()` and `virtualenv_starter()` (#1656, #1673).
 
 - Fixed an error when converting an empty NumPy char array to R (#1662).
@@ -35,10 +35,10 @@
 
 - Fixed a segfault encountered when running the Python session finalizer (#1663, #1664).
 
-- Resolved a segfault in RStudio when rapidly switching between 
+- Resolved a segfault in RStudio when rapidly switching between
   R and Python chunks in a Quarto document (#1665).
 
-- Improved behavior when the conda binary used to create an environment 
+- Improved behavior when the conda binary used to create an environment
   cannot be resolved (contributed by @tl-hbk, #1654, #1659).
 
 - Added Positron support for the Variables Pane and `repl_python()`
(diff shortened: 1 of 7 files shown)


t-kalinowski:
```r
Restarting R session...

> reticulate::repl_python()
Python 3.13.1 (/Users/tomasz/Library/Caches/r-reticulate/uv-cache/archive-v0/Kc-xLPZ1uy6xA8d7Kourf/bin/python3)
Reticulate 1.40.0.9000 REPL -- A Python interpreter in R.
Enter 'exit' or 'quit' to exit the REPL and return to R.
>>> 
```

- We might want to change how we present the path to the python installation here, seeing the actual path in the cache dir isn't helpful. Maybe something like:
   ```python
   Python 3.13.1 (Reticulate managed installation)
   Installed Python Packages: numpy, pandas, ...<other requirements from py_require()>
   Reticulate 1.40.0.9000 REPL -- A Python interpreter in R.
   Enter 'exit' or 'quit' to exit the REPL and return to R.
   ```

- We should probably default to Python 3.11 here, given this is a fresh R session.

edgararuiz:
> We should probably default to Python 3.11 here, given this is a fresh R session.

You're talking about only for `repl_python()`, right?

edgararuiz:
> We might want to change how we present the path to the python installation here, seeing the actual path in the cache dir isn't helpful. 

I would suggest to go even simpler: 

```r
Python version: 3.13.1 
Installed Python packages: numpy, pandas, ...<other requirements from py_require()>
Reticulate 1.40.0.9000 REPL -- A Python interpreter in R.
Enter 'exit' or 'quit' to exit the REPL and return to R.
```

t-kalinowski:
> You're talking about only for repl_python(), right? 

Good question. Maybe we should modify the print methods for py_config and py_last_error too, to avoid showing the path to the cached env.

edgararuiz:
Actually, using UV_PYTHON seems to be a good mechanism to set a "soft" default:

```r
> x <- uv_get_or_create_env()
> system2(x, c("--version"))     
Python 3.13.1
> 
> withr::with_envvar(
+   list("UV_PYTHON" = "3.11"), {
+     x <- uv_get_or_create_env()
+     system2(x, c("--version"))
+   }
+ )
Python 3.11.11                   
> 
> withr::with_envvar(
+   list("UV_PYTHON" = "3.11"), {
+     x <- uv_get_or_create_env(python_version = c("3.10"))  
+     system2(x, c("--version"))
+   }
+ )
Python 3.10.16                   
> 
> withr::with_envvar(
+   list("UV_PYTHON" = "3.11"), {
+     x <- uv_get_or_create_env(python_version = c(">=3.11", "<=3.12"))  
+     system2(x, c("--version"))
+   }
+ )
Python 3.12.0  
```

t-kalinowski:
Thanks Edgar! I merged to unblock Max; We'll address remaining issues in followup PR's.

PR #1719, opened 2025-01-26
`py_require()` checks and fails on 'equal' / 'not equal' Python version requests are used

edgararuiz:
- `uv run` does not accept Python version `==3.11`, and `==3.09`. In this change, `py_require()` removes any `==` from Python version requirements
- `uv run` does not accepts any combination of `==` requirements with non-equal requirements (`>=`, `<=`). In this change, `py_require()` checks and returns an informative error
- `uv run` does not accept more than one, different, equal-to Python version. In this change, `py_require()` will check and return an error this occurs 
- Adds tests for the changes mentioned above, and updates tests that were sending equal and non-equal Python requirements
- Unrelated, but just found out that newer pre-release versions of packages will modify the error output (https://pypi.org/project/tensorflow/#history), so switched the tests that requests a `tensorflow` 2.18 against an older snapshot from expect error test.
```
── Failure (test-py_require.R:12:3): Error requesting newer package version against an older snapshot ──
Snapshot of code has changed:
old[9:21] vs new[9:23]
    > py_require(exclude_newer = "2024-10-20")
    > uv_get_or_create_env()
      × No solution found when resolving `--with` dependencies:
-     ╰─▶ Because only tensorflow<2.18.dev0 is available and you require
-         tensorflow>=2.18.dev0, we can conclude that your requirements are
-         unsatisfiable.
+     ╰─▶ Because only the following versions of tensorflow are available:
+             tensorflow<2.18.dev0
+             tensorflow>2.19.dev0
+         and you require tensorflow>=2.18.dev0,<2.19.dev0, we can conclude that
+         your requirements are unsatisfiable.
```

### DIFF
--- a/R/py_require.R
+++ b/R/py_require.R
@@ -52,11 +52,16 @@ py_require <- function(packages = NULL,
   uv_initialized <- is_python_initialized() &&
     is_uv_reticulate_managed_env(py_exe())
 
-  if (uv_initialized && !is.null(python_version)) {
-    stop(
-      "Python version requirements cannot be ",
-      "changed after Python has been initialized"
-    )
+  if (!is.null(python_version)) {
+    if(uv_initialized) {
+      stop(
+        "Python version requirements cannot be ",
+        "changed after Python has been initialized"
+      )
+    }
+    if(substr(python_version, 1, 2) == "==") {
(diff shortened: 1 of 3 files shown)


t-kalinowski:
Thanks!

PR #1720, opened 2025-01-28
Py require default python

edgararuiz:
- Adds `reticulate_default_python()`
- `uv_get_or_create()` uses new function to get the default if no Python version is specified
- Updates `print.` and error `py_require()` output to clarify that it will use, or used, the default version
- Updates test snapshot


t-kalinowski (approved):
Thanks Edgar, LGTM!

PR #1730, opened 2025-02-11
Updates to `py_require()`  

t-kalinowski:
Changes in behavior:

- Don't signal error if called with an already satisfied requirement after Python has been initialized.
- Don't signal error if called with an unsatisfied requirement from a package after Python has been initialized.
-  Don't signal error if called with an unsatisfiable request before Python has been initialized. (Error is deferred until resolving the env, to give an opportunity in user code to proactively fix the requirements.)
-  In `uv_get_or_create_env()`, if python has been initialized, ensure we resolve the initialized python version.
- If an ephemeral uv env is initialized, only `action = "add"` for packages is supported.
- Add `...` to `py_require()` signature, require that `exclude_newer` and `action` are provided as named args. 
- Do not warn or show an error if `py_require()` is called after Python has initialized, but the call produces no actual changes to the requirements (e.g., removing packages that aren't in requirements, or adding packages that are, or setting with an identical set of packages). This allows a script with a `py_require()` call to be rerun in the same R session.
- Revise and expand `?py_require()`

### DIFF
--- a/R/py_require.R
+++ b/R/py_require.R
@@ -44,6 +44,8 @@
 #' @param python_version A character vector of Python version constraints \cr
 #'   (e.g., `"3.10"` or `">=3.9,<3.13,!=3.11"`).
 #'
+#' @param ... Dots are for future extensions, must be empty.
+#'
 #' @param action Defines how `py_require()` handles the provided requirements.
 #'   Options are:
 #'   - `add`: Add the entries to the current set of requirements.
@@ -62,8 +64,12 @@
 #' @export
 py_require <- function(packages = NULL,
                        python_version = NULL,
+                       ...,
                        exclude_newer = NULL,
                        action = c("add", "remove", "set")) {
+
+  if (length(list(...)))
+    stop("... must be empty")
 
   pr <- py_reqs_get()
 
--- a/R/py_require.R
+++ b/R/py_require.R
@@ -16,18 +16,18 @@
 #' with Python for the first time in the R session. Typically, this occurs when
 #' `import()` is first called.
 #'
-#' If `py_require()` is called with new requirements after Reticulate has
+#' If `py_require()` is called with new requirements after reticulate has
 #' already initialized an ephemeral Python environment, a new ephemeral
 #' environment is activated on top of the existing one. Once Python is
-#' initialized, only adding packages is supported--removing packages, changing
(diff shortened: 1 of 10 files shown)


R/py_require.R
```diff
@@ -1,118 +1,183 @@
...
-#' interacts with Python for the first time during the session. Typically,
-#' that would be the first time `import()` is called.
+#' The ephemeral virtual environment is not created until the user interacts
+#' with Python for the first time in the R session. Typically, this occurs when
+#' `import()` is first called.
 #'
-#' If `uv` is not installed, `reticulate` will attempt to download and install
-#' a version of it in an isolated folder. This will allow you to get the
-#' advantages of `uv`, without modifying your computer's environment.
+#' If `py_require()` is called with new requirements after Reticulate has
```
edgararuiz, line 19: Maybe add `py_install()`: "if `py_require()`, or `py_install()`, is called..."?

R/py_require.R
```diff
@@ -1,118 +1,183 @@
...
+#' already initialized an ephemeral Python environment, a new ephemeral
+#' environment is activated on top of the existing one. Once Python is
+#' initialized, only adding packages is supported--removing packages, changing
```
edgararuiz, line 22: Maybe a period would be clearer? "... is supported. Removing packages,..."

R/py_require.R
```diff
@@ -1,118 +1,183 @@
...
+#' the Python version, or modifying `exclude_newer` is not possible.
 #'
+#' Calling `py_require()` without arguments returns a list of the currently
+#' declared requirements.
 #'
-#' @param packages A vector of Python packages to make available during the
-#' working session.
+#' `py_require()` can also be used in R packages (e.g., in `.onLoad()` or
+#' elsewhere) to declare Python dependencies. The print method for
+#' `py_require()` shows the Python dependencies declared by R packages in the
```
edgararuiz, line 30: Maybe "enumerates" or "displays", instead of show?

edgararuiz (approved):
LGTM

R/py_require.R
t-kalinowski: We currently emit a warning if `py_install()` is called on an ephemeral environment, I don't think we want to encourage this workflow.

PR #1737, opened 2025-02-19
Update vignettes for `py_require()`

t-kalinowski:


vignettes/python_packages.Rmd
```diff
@@ -1,13 +1,157 @@
...
+-- Current requirements -------------------------------------------------
+ Python:   3.11.11 (reticulate default)
+ Packages: numpy, numpy>2, numpy<2
+-------------------------------------------------------------------------
+Error in uv_get_or_create_env() :
+  Call `py_require()` to remove or replace conflicting requirements.
+```
+
+You can call `py_require()` again with `action = "remove"` or `action = "set"`
+to update the requirements to make them satisfiable. Options for resolving are:
```
edgararuiz, line 88: There is a colon at the end here, but No 1 doesn't start until after the next paragraph, not sure if that was on purpose

vignettes/python_packages.Rmd
```diff
@@ -1,13 +1,157 @@
...
+    ```
+
+    See `py_require()` for more examples.
+
+## Manually managing Python installations
+
+::: {.alert .alert-info}
+> ## **⚠ Updated Guidance**
+>
+> The remaineder of this vignette describes how to manually manage a Python
```
edgararuiz, line 146: There's an extra 'e' in remainder

edgararuiz (approved):
LGTM

vignettes/python_packages.Rmd
t-kalinowski: Thanks!

vignettes/package.Rmd
```diff
@@ -2,119 +2,250 @@
...
   %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
 ---
 
+## Declaring Python Requirements
 
-## Delay Loading Python Modules
+Beginning with Reticulate version 1.41, R packages can declare Python
+requirements with `py_require()`. Reticulate can then resolve these requirements
+at runtime using an ephemeral Python environment.
```
kevinushey, line 18: What exactly is `py_require()` doing? When do the requested Python packages get loaded?

vignettes/package.Rmd
```diff
@@ -2,119 +2,250 @@
...
 
-If you write an R package that wraps one or more Python packages, it's likely that you'll be importing Python modules within the `.onLoad` method of your package so that you can have convenient access to them within the rest of the package source code.
+Calling `py_require()` is always recommended because it has no effect if the
+user is using a manually managed Python environment, making it generally safe to
+use.
```
kevinushey, line 22: The phrase "Always recommended" feels a bit awkward. Always recommended compared to what?

vignettes/package.Rmd
```diff
@@ -2,119 +2,250 @@
...
-  # use superassignment to update global reference to scipy
-  scipy <<- reticulate::import("scipy", delay_load = TRUE)
+  reticulate::py_require("scipy")
 }
 ```
 
-Using the `delay_load` flag has two important benefits:
+It can also be called from other package functions. For example, the `keras3` R
+package defines a `use_backend()` function that calls `py_require()` with
+different values based on the user-requested backend:
```
kevinushey, line 36: Can we emphasize that the intent here that:

- The `tensorflow` package provides a default backend, which `keras` makes available by default,
- Other packages (e.g. `jax`) provide alternate backends, which can be selected if desired by the user.

kevinushey: I think this is also pointed out in the comments but would be good to emphasize here.

vignettes/package.Rmd
```diff
@@ -2,119 +2,250 @@
...
+
+``` r
+library(pysparklyr) # declares requirements for PySpark
+library(keras3)     # declares requirements for default 'tensorflow' backend
+use_backend("jax")  # removes 'tensorflow' requirements, adds 'jax' requirements
+
+library(reticulate)
+py_require(c("scipy", "polars"))         # user-declared requirements
+py_require(python_version = ">=3.12")
+py_require(exclude_newer = "2025-01-19")
```
kevinushey, line 83: What does this do? `exclude_newer` hasn't been discussed in this vignette; is it needed?

vignettes/package.Rmd
```diff
@@ -2,119 +2,250 @@
...
+2.  Users to specify their Python installation before using your package.
 
-Without the `delay_load`, Python would be loaded immediately and the user's call to `use_virtualenv` would have no effect.
+Example:
+
+``` r
+scipy <- NULL
+
+.onLoad <- function(libname, pkgname) {
+  reticulate::py_require("scipy")
```
kevinushey, line 122: Any reason why we can't do something like `scipy <<- py_require("scipy")`, and have that be identical to delay_load below?

vignettes/python_packages.Rmd
```diff
@@ -1,13 +1,175 @@
...
+
+`py_require()` is the recommended way to declare Python dependencies. A
+`py_require()` call is similar to a `library()` call for Python packages.
+
+`py_require()`:
+
+-   Should appear at the start of an R script.
+-   Declares the Python dependencies needed by the script.
+-   Is limited in scope to the current R session.
+-   Has no effect if called multiple times.
```
kevinushey, line 24: (if called with the same arguments?)

vignettes/package.Rmd
t-kalinowski: `py_require()` calls build up a manifest of requirements internally. The manifest is used when reticulate is initializing Python. The manifest is onl used to resolve an ephemeral virtual environment if no self-managed python installation is higher in the order of discovery.

vignettes/package.Rmd
t-kalinowski: Compared to not calling it? 

I can mention in this paragraph this supersedes previous guidance, like adding a section to DESCRIPTION or calling `use_virtualenv(required = FALSE)` in `.onLoad()`, if that would make it clearer.

vignettes/package.Rmd
t-kalinowski: `exclude_newer` is discussed in the other vignette. I included it here as a reminder to package authors that their requirements will be combined with a wider variety of user requirements. If you think it's too confusing here I can remove it.

vignettes/package.Rmd
t-kalinowski: Unfortunately, there is not a 1:1 correspondence between PyPi Package names and Python module names. Any given PyPi package might install any number of modules, named anything, not necessarily sharing a name with the package. 

Consider the `tensorflow` module, which might be provided by multiple different packages published on PyPi: `tensorflow`, `tensorflow-cpu`, `tensorflow-macos`, `tensorflow-metal`, not to mention version constraints or modifiers that are often included with package names like `tensorflow[and-cuda]>=2.18.*`

t-kalinowski:
Thanks @kevinushey! 

I've updated based on your comments in https://github.com/rstudio/reticulate/pull/1740

PR #1753, opened 2025-02-27
potential fix for windows interrupts

kevinushey:


t-kalinowski:
Many thanks @kevinushey! I was really stumped on the issue with interrupts on Windows in Rstudio IDE, and this helps tremendously.

Do you think it's ready to merge?

src/libpython.cpp
```diff
@@ -453,39 +457,57 @@ bool import_numpy_api(bool python3, std::string* pError) {
...
+  if (buffer == NULL || buffer == Py_None)
+    return true;
 
-  PyObject* sys_stdout(PySys_GetObject("stdout"));  // returns borrowed reference
-  if (sys_stdout == NULL)
-    status = -1;
-  else
-    tmp = PyObject_CallMethod(sys_stdout, "flush", NULL);
+  // try to invoke flush method
+  if (!PyObject_HasAttrString(buffer, "flush"))
```
t-kalinowski, line 470: This probably doesn't matter, but I noticed in the python sources that `PyObject_HasAttr()` just calls `PyObject_GetAttr()` and then checks if an exception was raised. It's trivial, but we're on a hot code path here and the approach seems unnecessarily expensive if we're going to immedeatly fetch the attr again anyway in the common case.

More relevant however, is that `flush` should always be present.
`sys.stdout` and `sys.stderr` are supposed to be a file object, and the abstract io base class defines this mixin method always: https://docs.python.org/3/library/io.html#io.IOBase (the method does nothing for output streams where
flushing doesn't apply)

If `flush` missing, I think we should emit some error output, because it indicates that something is "wrong".

kevinushey: FWIW we have this code that apparently expects stdout to be null in some scenarios.

https://github.com/rstudio/reticulate/blob/bugfix/windows-interrupts/R/output.R#L9-L13

t-kalinowski:
Thank you very much!

PR #1765, opened 2025-03-14
Make sure `py_run_on_thread` knows its initializing PositronIPyKernel

dfalbel:
This fixes the disappearing `r` object in the reticulate python console in Positron.
The problem is that with some folder renaming in Positron we were silently running the script without the workarounds to make `r` correcly appear in the console.
We should definitely update positron to call a dedicated entrypoint so we never silently fail when other file structure refactoring happens.

Opened https://github.com/posit-dev/positron/issues/6788 to track this issue.


t-kalinowski:
Thank you!

Issue #1420, opened 2023-07-20
Micromamba env created from yaml: /bin/micromamba env: No such file or directory


ktmeaton:
## Error

Very exciting to see `micromamba` supported, thank you so much! I'm getting this error when I try to use `reticulate=1.30` in an environment created from a YAML file with `micromamba=1.49`. It appears as if `reticulate` can't find the python binary, and is trying to use the 'conda' binary instead.

```bash
Error in normalizePath(conda, winslash = "/", mustWork = TRUE) :
  path[1]="/home/username/.local/bin/micromamba env": No such file or directory
Calls: <Anonymous> ... python_munge_path -> get_python_conda_info -> normalizePath
Execution halted
```

## Environment from CLI (Works)

Creating the environment plainly from the CLI works for me with no errors.

```bash
micromamba create -y -n reticulate-from-cli r-base=4.2.1 r-reticulate=1.30
micromamba run -n reticulate-from-cli Rscript -e 'reticulate::use_condaenv("reticulate-from-cli")'
```

## Environment from YAML (Error)

However creating  from an  `environment.yaml` file causes errors.

```yaml
name: reticulate-from-yaml
channels:
  - conda-forge
  - defaults
dependencies:
  - conda-forge::r-base=4.2.1
  - conda-forge::r-reticulate=1.30
```

```bash
micromamba env create  -y -f environment.yaml -n reticulate-from-yaml
micromamba run -n reticulate-from-yaml Rscript -e 'reticulate::use_condaenv("reticulate-from-yaml")'

Error in normalizePath(conda, winslash = "/", mustWork = TRUE) :
  path[1]="/home/user5028384/.local/bin/micromamba env": No such file or directory
Calls: <Anonymous> ... python_munge_path -> get_python_conda_info -> normalizePath
Execution halted
```

These two environments are otherwise identical with no differences, and both have python3.11 installed.

```bash
micromamba -n reticulate-from-cli list > reticulate-from-cli.txt
micromamba -n reticulate-from-yaml list > reticulate-from-yaml.txt

diff reticulate-from-cli.txt reticulate-from-yaml.txt
```

## `conda` vs `mamba` vs `micromamba`

This seems to be restricted to `micromamba=1.49` as `mamba=1.49` has no errors.

```bash
mamba env create -f environment.yaml -n reticulate-from-yaml-mamba
[...]

PR #1771, opened 2025-03-20
Add direct check for micromamba and mamba executables, adjust preferred ordering

gdevenyi:
Fixes https://github.com/rstudio/rstudio/issues/15763#issuecomment-2740354896

This is untested as I don't have a proper r development environment to test, but the modification is very straightforward.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,18 +1,21 @@
 # reticulate (development version)
 
+- Fixed usage of micromamba and mamba, next-generation conda environment management tools.
+  reticulate now prefers to use micromamba, mamba, then conda when managing environments
+
 # reticulate 1.42.0
 
 - Fixed an issue in RStudio on Windows where interrupts were
   ignored while Python code was executing (#1753).
-  
+
 - Updates for Positron to fix issues with `repl_python()` and Variables Pane (#1755).
 
 - Fixed an issue where `[` received Python objects as slice arguments.
   e.g., `x[start:end]` when `start` or `end` were Python objects (#1731).
-  
-- The `[` method will now translate symbol `..` to a 
+
+- The `[` method will now translate symbol `..` to a
   Python Ellipsis `...`. (#1763)
-  
+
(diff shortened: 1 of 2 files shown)


t-kalinowski:
Would this potentially be a breaking change for users who have both `conda` and `micromamba` installed?

Would it be safer to add `micromamba` later in the order of discovery, and avoid disrupting existing workflows?

Does setting the option `options("reticulate.conda_binary" = "/path/to/micromamba")` work currently?

gdevenyi:
> Would this potentially be a breaking change for users who have both conda and micromamba installed?

According to the existing code/news, micromamba, mamba and conda are all supported as executables for managing conda environments. 

https://github.com/rstudio/reticulate/blob/main/NEWS.md#misc

> Does setting the option options("reticulate.conda_binary" = "/path/to/micromamba") work currently?

According to this
https://github.com/rstudio/reticulate/issues/1420

Currently they need to explicitly set micromamba binary to get it to work. This would resolve that as well.

t-kalinowski:
> According to the existing code/news, micromamba, mamba and conda are all supported as executables for managing conda environments.

Yes, reticulate allows users to opt in to using micromamba or mamba instead of conda. However, the patch you're proposing would make mamba usage opt-out. I’m not sure how conda's feature set compares to mamba and micromamba. Is mamba truly a drop-in replacement for conda?  

To avoid disrupting existing reticulate+conda workflows, I think it would be better to prefer conda when both conda and mamba are installed, falling back to mamba and micromamba only if `conda` is not found.

gdevenyi:
> Is mamba truly a drop-in replacement for conda?

Yes.

t-kalinowski:
Can you please add a NEWS entry?

gdevenyi:
> Can you please add a NEWS entry?

Sorry for the delay. Done.

t-kalinowski:
Thanks!

Issue #1780, opened 2025-04-08
Error using `py_require()` with standard libraries


lazappi:
I get an error if `py_require()` is called with a standard library before importing it:

``` r
Sys.setenv("RETICULATE_USE_MANAGED_VENV" = "yes")
reticulate::py_require("os")
reticulate::import("os")
#> Error in uv_get_or_create_env() : 
#>   Call `py_require()` to remove or replace conflicting requirements.
#> Error: Installation of Python not found, Python bindings not loaded.
#> See the Python "Order of Discovery" here: https://rstudio.github.io/reticulate/articles/versions.html#order-of-discovery.
```

<sup>Created on 2025-04-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This probably makes sense as standard libraries are not available on PyPI (`pip install os` also fails) but I think it's easy to do naively if you always require a package before importing it. Would it be possible to avoid adding standard libraries to the requirements or at least give a more specific message?

lazappi:
Something like this might be a partial workaround but it requires a relatively recent version of Python is available and initalized:

```r
if (reticulate::py_available() && reticulate::py_version() >= "3.10") {
  sys <- reticulate::import("sys")
  py_builtins <- reticulate::import_builtins()
  standard_modules <- py_builtins$list(sys$stdlib_module_names)
  
  if (!(module %in% standard_modules)) {
    reticulate::py_require(module)
  }
}
```

t-kalinowski:
Thanks, I agree we can have a nicer error message here, or better, accept the already-satisfied requirement silently. 
However, we can't use anything from a Python runtime, because `py_require()` can't initialize Python. 

We can have a character vector of common builtin modules and then filter against them. 

E.g.,
```r
py_stdlib_module_names <- c("os", "sys", ...)

py_require <- function(packages, ...) {
  packages <- setdiff(packages, py_stdlib_module_names)
  ...
}
```
Would you like to submit a PR?

lazappi:
Yeah, I could probably but together something like that. Using a character vector was my initial was my initial thought as well but I wasn't sure how often they change (and there are quite a lot of them). Using the list from the most recent Python should cover most things though.

t-kalinowski:
I wonder if it's a good idea to take on maintaining such a list. There is also the related common stumbling block of discovering that python package names != python module names.

It might better to adjust the error message to point out that python package names != python module names, and python comes with many modules already available.

PR #1788, opened 2025-04-11
Adjust error message in `uv_get_or_create_env()`

lazappi:
Add extra context to the error message in `uv_get_or_create_env()` explain that standard modules should not be passed to `py_require()`.

Related to #1780


lazappi:
Is there something I should do to update the snapshots? I tried running the checks/commands locally but it says there's nothing to update.

Issue #1650, opened 2024-08-22
Support for pixi environments?


royfrancis:
Any plans to support [pixi](https://prefix.dev/) environments?

Robinlovelace:
Was about to request the same feature, big 👍 to that, could save lots of time compared with `reticulate::install_miniconda()`

n8layman:
+1

t-kalinowski:
> Any plans to support [pixi](https://prefix.dev/) environments?

What features would be needed to support pixi environments? 
Does pointing reticulate with env var `RETICULATE_PYTHON` at a pixi-managed python installation work?

n8layman:
It would generally follow the roadmap of #1669. Specifically to get it to work currently I found that I needed to:
1. Address the conda-meta/history file bug (currently I've resolved it via an activation script as suggested by John in #1184)
2. Point `RETICULATE_PYTHON` to the pixi python binary. Adding this `RETICULATE_PYTHON="/Users/nathanlayman/homedir/repo/.pixi/envs/default/bin/python"`to my `.Renviron` file worked for me.
3. I'm also having trouble getting things working with radian, though I suspect that's a separate issue, as in https://github.com/randy3k/radian/issues/282

t-kalinowski:
You linked to the `uv` issue... does `uv` address your needs?

Note we have working `uv` support in the dev version of reticulate right now, see `?py_require()`. https://rstudio.github.io/reticulate/dev/reference/py_require.html

n8layman:
I think pixie actually uses uv under the hood! I did test on the dev version but still ended up needing the hacks above to get it to work.

n8layman:
Another. Potential issue is that pixi, and uv I think, can manage multiple versions of Python within different environments. I've only tested the one here so far in the `.pixi/envs/default` environment. But it is possible to have multiple environments in a single project.

pormr:
> It would generally follow the roadmap of [#1669](https://github.com/rstudio/reticulate/issues/1669). Specifically to get it to work currently I found that I needed to:
> 
> 1. Address the conda-meta/history file bug (currently I've resolved it via an activation script as suggested by John in [`invalid 'path' argument` error when Reticulate looks for Python #1184](https://github.com/rstudio/reticulate/issues/1184))
[...]

Issue #1184, opened 2022-03-30
`invalid 'path' argument` error when Reticulate looks for Python


jnolis:
Hi!

Previously we had a working image of Reticulate that connected to Python (reticulate version 1.22). When we upgraded to 1.24, we now get the following error when starting reticulate:

```
> options(error = function(e) print(rlang::trace_back()))
> reticulate::py_config()
Error in path.expand(path) : invalid 'path' argument
     █
  1. ├─reticulate::py_config()
  2. │ └─reticulate:::ensure_python_initialized()
  3. │   └─reticulate:::initialize_python(required_module, use_environment)
  4. │     └─(function() {...
  5. │       └─reticulate::py_discover_config(required_module, use_environment)
  6. │         └─reticulate:::python_config(python_version, required_module, forced = "RETICULATE_PYTHON")
  7. │           └─reticulate:::python_munge_path(python)
  8. │             └─reticulate:::get_python_conda_info(python)
  9. │               └─base::normalizePath(conda, winslash = "/", mustWork = TRUE)
 10. │                 └─base::path.expand(path)
 11. └─(function (e) ...
```

Note that I am manually setting the location for Python with the env var `RETICULATE_PYTHON=/opt/conda/envs/saturn/bin/python`

I traced it down and it looks like the error is extremely similar to #1149, however using the latest release from github does not resolve it.

It appears that in the `reticulate:::get_python_conda_info(python)` command, the code is in this case calling `reticulate:::python_info_condaenv_find("/opts/conda/env/saturn")`. In this case that function determines:

1. that there is no `condabin` folder
2. that there is a `conda-meta/history` file
3. the `history` file is empty, and thus returns NULL, causing an error on `path.expand`

I'm not sure how this logic was chosen to determine whether or not this is a valid conda installation, but since this conda installation worked in previous versions of reticulate it seems like it should work in this one too?

Thanks!

jnolis:
Just as a note, as a kind of silly work around I ran a conda command and noticed the history had `# cmd: /opt/conda/bin/conda install papermill` at the top, and so I am going to have our images just manually insert that line during the build process which fixes it.

kevinushey:
Thanks for the bug report!

Are you able to share any other details re: how these Anaconda environments are created? I'm curious why the `conda-meta/history` file exists but is empty.

[...]

PR #1803, opened 2025-06-02
Fix normalizePath null call with missing conda

tl-hbk:
I was trying to use reticulate with a pixi env which only populates `conda-meta/history` with

```
// not relevant for pixi but for `conda run -p`
```

I have `options(reticulate.conda_binary = conda)` set to a copy of `micromamba.exe` to handle the conda related operations but when `use_python` tries to get the conda root/binary info it passes `NULL` to `normalizePath` if it can't find conda automatically, which causes the following error `Error in path.expand(path) : invalid 'path' argument` which is mentioned in https://github.com/rstudio/reticulate/issues/1184

```
 1: source("src/r/conda.R")
 2: withVisible(eval(ei, envir))
 3: eval(ei, envir)
 4: eval(ei, envir)
 5: conda.R#9: use_python(pixi_env, required = TRUE)
 6: python_config(python)
 7: python_munge_path(python)
 8: get_python_conda_info(python)
 9: normalizePath(conda, winslash = "/", mustWork = FALSE)
10: path.expand(path)
```

Related issue:
https://github.com/rstudio/reticulate/issues/1650

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,9 @@
 - Fixed error in `install_python()` under R 4.5 when the requested Python
   version has a `":latest"` suffix, as it does by default. (#1792, #1797)
 
+- Fixed error in `get_python_conda_info()` when conda not found through `conda-meta/history` 
+  and `NULL` is passed to `normalizePath` (#1184)
+
 # reticulate 1.42.0
 
 - Fixed an issue in RStudio on Windows where interrupts were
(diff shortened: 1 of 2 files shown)


t-kalinowski:
Thanks for the PR. Can you please add a NEWS item?

t-kalinowski:
Thank you!

PR #1822, opened 2025-07-31
Enable reticulate integration when running within Positron

dfalbel:
Pairs with https://github.com/posit-dev/positron/pull/8769

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -63,6 +63,12 @@
 
   if (is_positron()) {
 
+    tryCatch({
+      maybe_enable_positron_reticulate_integration()
+    }, error = function(e) {
+      # if we fail to enable, it's likely running in a incompatible Positron version
+    })
+
     setHook("reticulate.onPyInit", function() {
 
       disable <- nzchar(Sys.getenv("_RETICULATE_POSITRON_DISABLE_VARIABLE_INSPECTORS_"))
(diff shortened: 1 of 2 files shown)


R/zzz.R
```diff
@@ -100,6 +100,52 @@
   }
 }
 
+.onAttach <- function(libname, pkgname) {
```
t-kalinowski, line 103: What do you think about doing this in `.onLoad()` instead?

R/zzz.R
```diff
@@ -100,6 +100,52 @@
...
+      TRUE 
+    })
+
+    if (!explicitly_disabled) {
+      tryCatch({
+        .ps.ui.executeCommand("positron.reticulate.toggleEnabled")
+      
+        if (requireNamespace("cli", quietly = TRUE)) {
+          cli::cli_inform(
+            "Reticulate integration has been enabled, to disable set {.url positron://settings/positron.reticulate.enabled} to {.val false}.",
```
t-kalinowski, line 134:
```suggestion
            "Positron's reticulate integration is now enabled. To disable, toggle {.href [positron.reticulate](positron://settings/positron.reticulate.enabled)}", 
```

Clicking the links brings me to a GUI checkbox to disable, rather than the source-mode json. Tweaked wording a little to make clear it's a Positron feature and present the link differently.

R/zzz.R
```diff
@@ -100,6 +100,52 @@
...
+        .ps.ui.executeCommand("positron.reticulate.toggleEnabled")
+      
+        if (requireNamespace("cli", quietly = TRUE)) {
+          cli::cli_inform(
+            "Reticulate integration has been enabled, to disable set {.url positron://settings/positron.reticulate.enabled} to {.val false}.", 
+            class = "packageStartupMessage"
+          )
+        } else {
+          packageStartupMessage(
+            "Reticulate integration has been enabled, to disable set positron://settings/positron.reticulate.enabled to `false`."
```
t-kalinowski, line 139:
```suggestion
            "Positron's reticulate integration has been enabled. To disable, set positron://settings/positron.reticulate.enabled to `false`"
```

t-kalinowski (commented):
Thanks! Looks good :)

R/zzz.R
t-kalinowski: Thanks Daniel. Per slack conversation, let's go with `.onLoad` since there are many users that never explicitly attach reticulate that will benifit from the feature. 

This also avoids us needing to handle the situation where the package is attached after reticulate has already initialized Python.

PR #1825, opened 2025-08-13
Adds support for Python versions with a pattern of `x.x.*` such as `==3.12.*`

edgararuiz:
- Modifies regex to avoid leaving `*` as part of the op
- Adds regex to modify the version to remove a trailing `.*` if its there
- Uses a `try()`  to the `op` operation to capture the error and pass a more informative error message
- Adds `version` to the `.mapply` call to capture the original version string and pass it in the error message
- Adds test to make sure the new pattern is supported and returns the expected boolean flags
- Adds test to ensure new error message is returned for future invalid versions that may fail at the point of op
- Adds test to ensure that the pattern `x.*.x` will not work. It does not fail, but any version passed does not return `TRUE`. This is a bit outside the scope of this small effort, but figured it would be worth adding a small test for it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # reticulate (development version)
 
-- Adds support for Python versions with a pattern of `x.x.*` such as `==3.12.*`.
+- Adds support for requesting Python versions with a wildcard pattern `x.x.*` such as `==3.12.*` in `virtualenv_starter()`, `py_require()`, and related functions (#1825)
 
 - Restored compatability with `uv` versions >= 0.8.0 (#1818).
 


R/virtualenv.R
```diff
@@ -683,23 +683,29 @@ as_version_constraint_checkers <- function(version) {
...
     op <- as.symbol(op)
     force(ver)
     force(ver_string)
-
+    force(version)
     # return a "checker" function that takes a vector of versions and returns
     # a logical vector of if the version satisfies the constraint.
     rlang::zap_srcref(eval(bquote(function(x) {
-      op <- .(op)
+      op <- try(.(op), silent = TRUE)
```
t-kalinowski, line 705: I'm having difficulty understanding, when would evaluating `op` error?

R/virtualenv.R
```diff
@@ -683,23 +683,29 @@ as_version_constraint_checkers <- function(version) {
...
+      if (inherits(op, "try-error")) {
+        stop("Version `", version, "` is not valid.")
```
t-kalinowski, line 707: I'm not sure if it's safe to throw an error from the checker here, since this function is used in a few contexts. Also, is it throwing too early? Are we potentially throwing an error on a valid input like `3.13-dev` or `3.14.0rc1` or `3.13.6t`, i.e., strings that don't parse as semantic versions but can be matched exactally?

edgararuiz: I'm only returning a more specific message if the `.(op)` call fails, that's all. With out this code calling, for example: `virtualenv_starter("==3.12.-")` returns this in the current version:

`Error in check(starters$version) : object '==-' not found`

This change simply makes it clear that the version itself is what's actually problematic: 

`Error in check(starters$version) : Version `==3.12.-` is not valid.`

In other words, I'm not adding an extra check, just a more helpful error message

edgararuiz: Btw, it looks like `3.13-dev` fails in the current version, is that to be addressed too?

```
> reticulate::virtualenv_starter("3.13-dev")
Error in check(starters$version) : object '-dev' not found
```

It also fails on this new version, but with a better message:

```
> virtualenv_starter("3.13-dev")
Error in check(starters$version) : Version `3.13-dev` is not valid.
```

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # reticulate (development version)
 
+- Adds support for Python versions with a pattern of `x.x.*` such as `==3.12.*`.
```
t-kalinowski, line 3:
```suggestion
- Adds support for requesting Python versions with a wildcard pattern `x.x.*` such as `==3.12.*` in `virtualenv_starter()`, `py_require()`, and related functions (#1825)
```

t-kalinowski:
Thanks!

Issue #1820, opened 2025-07-29
conda channel Terms of Service error


xgirouxb:
Hello, I've tried running `reticulate::install_miniconda()` and end up with the same error as below. I've also tried installing miniconda as well as installing manually, specifying the path to conda in `.Rprofile` b ut I end up where I started when I try and create an environment. Could this somehow be related to https://github.com/scikit-learn/scikit-learn/issues/31773 ?

``` r
reticulate::conda_create()
#> + "C:/Users/xgirouxb/AppData/Local/r-miniconda/condabin/conda.bat" create --yes --name r-reticulate "python=3.10" --quiet -c conda-forge

#> CondaToSNonInteractiveError: Terms of Service have not been accepted for the following channels. Please accept or remove them before proceeding:
#>     • https://repo.anaconda.com/pkgs/main
#>     • https://repo.anaconda.com/pkgs/r
#>     • https://repo.anaconda.com/pkgs/msys2
#> 
#> To accept a channel's Terms of Service, run the following and replace `CHANNEL` with the channel name/URL:
#>     ‣ conda tos accept --override-channels --channel CHANNEL
#> 
#> To remove channels with rejected Terms of Service, run the following and replace `CHANNEL` with the channel name/URL:
#>     ‣ conda config --remove channels CHANNEL

#> Error: Error creating conda environment 'r-reticulate' [exit code 1]
```

<sup>Created on 2025-07-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">

<summary>

Session info
</summary>

``` r
sessionInfo()
#> R version 4.5.0 (2025-04-11 ucrt)
#> Platform: x86_64-w64-mingw32/x64
#> Running under: Windows 11 x64 (build 22631)
#> 
#> Matrix products: default
#>   LAPACK version 3.12.1
#> 
#> locale:
#> [1] LC_COLLATE=English_Canada.utf8  LC_CTYPE=English_Canada.utf8   
#> [3] LC_MONETARY=English_Canada.utf8 LC_NUMERIC=C                   
#> [5] LC_TIME=English_Canada.utf8    
#> 
#> time zone: America/Toronto
#> tzcode source: internal
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
[...]

Issue #1800, opened 2025-05-22
Switch to Miniforge by default


benz0li:
[Anaconda](https://www.anaconda.com) seems to actively contact organisations with two hundred (200) or more employees regarding the requirement of a paid license (Anaconda Business or Anaconda Enterprise) in order to legally access their [default channel](https://www.anaconda.com/docs/tools/working-with-conda/reference/default-channels) ([`repo.anaconda.com`](https://repo.anaconda.com)).

> ### 2. USERS & LICENSING
> 
> **2.1 Organizational Use.** Your registration, download, use, installation, access, or enjoyment of all Anaconda Offerings on behalf of an organization that has two hundred (200) or more employees or contractors (“Organizational Use”) requires a paid license of Anaconda Business or Anaconda Enterprise. For sake of clarity, use by government entities and nonprofit entities with over 200 employees or contractors is considered Organizational Use.

– https://legal.anaconda.com/policies/en/#terms-of-service

R is open source software. Most users are not aware that [Miniconda](https://www.anaconda.com/docs/getting-started/miniconda/main) (the default for reticulate) uses a [freemium](https://en.wikipedia.org/wiki/Freemium) business model that could cause their employer legal trouble.

Therefore, [Miniforge](https://conda-forge.org/download/) – which uses [conda-forge](https://conda.anaconda.org/conda-forge) as the default channel – should be considered as the new default.

benz0li:
Cross reference:

* https://github.com/jupyter/docker-stacks/pull/1189

t-kalinowski:
Thanks for raising the issue. I agree this is something we should do for the next release. 

Would you like to submit a PR?

benz0li:
> Would you like to submit a PR?

Unfortunately not. I am swamped with work.

benz0li:
Resolved by

* https://github.com/rstudio/reticulate/pull/1829

@t-kalinowski I will close this as soon as a new version of reticulate is released that includes this change.

PR #1827, opened 2025-08-28
Add miniforge alternative for miniconda

raphaelbetschart:
Hi, I added miniforge as an alternative for miniconda. This should fix the error of the `Conda Terms of Service` error on Windows [https://github.com/rstudio/reticulate/issues/1820](https://github.com/rstudio/reticulate/issues/1820), and adds the pull request from [https://github.com/rstudio/reticulate/issues/1800](https://github.com/rstudio/reticulate/issues/1800). As a disclaimer, I am not very familiar with pull requests and creating R packages.


raphaelbetschart:
Hi Tomasz, so basically switching from miniconda to miniforge?
This seems even easier, as long as it doesn't break anything.

t-kalinowski:
Yes. I think we already install miniforge on arm macs, and now we can switch over to doing the same on windows and linux.

PR #1844, opened 2025-10-02
Fix expansion of `!!` in a string literal

### LINKED ISSUE #1843: In Python-REPL, `!!` is converted to `%system ` even when it's inside a string
Inside a REPL initialised with `reticulate::repl_python()`, `!!` is explicitly replaced by `%system ` to allow it to be used as a magic command. However, `!!` is also replaced when it occurs inside of a string, which should not happen.

Here are a few examples of what happens inside a Python-REPL:

```
>>> "!!"
'%system '
>>> "ab!!cd!!ef"
'ab%system cd%system ef'
>>> "!!" == "%system "
True
>>> len("!!")
8
```

t-kalinowski:
Closes #1843

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -57,7 +57,7 @@ Suggests:
     pillar,
     testthat
 LinkingTo: Rcpp
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.3
 Roxygen: list(markdown = TRUE)
 VignetteBuilder: knitr
 Config/build/compilation-database: true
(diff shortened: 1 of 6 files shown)


stibu81:
Thanks, I checked that this works with my examples as expected. I might be wrong, but shouldn't the regex in `repl_expand_bangbang()` also allow for an asterisk in the lhs to catch something like this:

```
a, *b = !!ls
```

t-kalinowski:
Thanks for the report! The fix is on main now.

PR #1859, opened 2025-10-30
Fixes display values in positron

dfalbel:
In https://github.com/posit-dev/positron/pull/8545 the `get_display_value()` API changed and no longer allows a `width` argument.

The approach here should also work on previous positron versions, as width shouldn't be required and we still truncate on this side.

### DIFF
--- a/R/ark-variables-methods.R
+++ b/R/ark-variables-methods.R
@@ -5,7 +5,8 @@
 
 
 ark_positron_variable_display_value.python.builtin.object <- function(x, ..., width = getOption("width")) {
-  .globals$get_positron_variable_inspector(x)$get_display_value(width)[[1L]]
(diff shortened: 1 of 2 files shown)


t-kalinowski (approved):
Thank you!

This works great again now after the python session has initialized.

I noticed one small issue: modules imported with `lazy_load=TRUE` before python has initialized still do not display correctly in the variables inspector.

Image

I don't think this is worth, since it would introduce some overhead and complexity, and this is an edge case that most people will never encounter, since most modules in scripts are not imported with `lazy_load`, and most package exports are not visible in the inspector typically.

One approach to fixing might be to initialize the `get_positron_variables_inspector` with something that mimics the variables api simply (i.e., dispatches to `str()`). This would then get overwritten when we initialize python.

PR #1878, opened 2026-02-03
Fix #1872

### LINKED ISSUE #1872: Bug: can't add requirements after Python is initialized
I noticed it is difficult or even a coincidence if adding packages to requirements works after Python is initialized: 

```r
reticulate::py_require("pigar")
pigar <- reticulate::import("pigar")
reqs <- c("numpy", "seaborn")
reticulate::py_require(reqs, action = "add")
#> Error in `reticulate::py_require()`:
#> ! After Python has initialized, only `action = 'add'` with new packages is supported.
```
The issue is that `numpy` in `reqs` is already a requirement by default, which triggers this:

https://github.com/rstudio/reticulate/blob/b33bfbf8bc0f935747cde49c9c5de43dbf407daf/R/py_require.R#L265-L269

There is an easy workaround, namely ignoring these packages:

```r
reqs <- setdiff(reqs, reticulate:::py_reqs_get()$packages)
reticulate::py_require(reqs, action = "add")
```

But `py_reqs_get` is not normally available to the user and is not documented. This seems like a bug to me and I'm happy to fix it in a PR. Unless there is a reason why this was implemented.

JBGruber:
Fix #1872

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,9 @@
 - `virtualenv_starter()` now excludes free-threaded CPython builds,
   which are not supported by reticulate (#1883).
 
+- Fixed `py_require()` to allow `action = "add"` with already-required packages
+  after Python initialization, while still erroring on true version conflicts (#1878, #1872; @JBGruber).
+
 - Fix spurious `Error: ignoring SIGPIPE signal` on Unix when embedded Python writes to a closed pipe (#1868).
 
 - Fixed Positron UI interactions, such as restart/create new session with reticulate (#1871, #1869).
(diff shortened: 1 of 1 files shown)


t-kalinowski:
Thank you for the PR! This looks like the right fix to my eyes.

Would you be able to add a snapshot/regression test that tests the happy path (no error), and also, captures the new error? Something like:

https://github.com/rstudio/reticulate/blob/0ce0b7430350776d18a24729a5a448a29e6aacab/tests/testthat/test-py_require.R#L19

```r
test_that("....", {
  local_edition(3)

  expect_snapshot2(r_session({
    library(reticulate)
    py_require(....)
    import("sys") # force python initialize
    py_require(....) # happy path
    py_require(....) # new error 
  })
 )
})
```

